PR #6, opened 2019-09-13
Name change to orix

pc494:
This is going to be more work than I had anticipated, but once the name change is completed we should be able to get something to pypi.


pc494:
I think the plan here is to leave the majority of adjustments that bring this inline with the rest of code base until `0.2.0` so work towards something like:

In time for `0.1.0`
---------------------
- [ ] Have CI for linux 
- [ ] Have github and pypi releases

Beyond `0.1.0`
------------------
Move over to GNU licensing (and pyxem based documentation)
Have coveralls setup
Have a directory structure consistent with pyxem
Have a CI for windows 
Website

PR #8, opened 2019-09-23
orix 0.1.0

dnjohnstone:


dnjohnstone:
@pc494 any ideas why "no data is collected" about coverage when travis runs here?

pc494:
> @pc494 any ideas why "no data is collected" about coverage when travis runs here?

Not at first pass, I'll have another look when I get in.

dnjohnstone:
Ok, I'm going to merge this since it's now got quite a lot of changes in it as a result of packaging tests and running pepstorm - we can make a new PR to explore why coverage is odd.

Issue #21, opened 2019-12-02
Misorientation.equivalent() has a broken codepath


pc494:
The property `equivalent` (of `Misorientation`, but inhereted by `Orientation`) is broken for the non-grain exchange case. Inspecting the copy+paste snippet below it's clear that for `Gl._tuples != Gr._tuples` we will always get an UnboundLocal error (on "`orientations`")

```
    def equivalent(self):
        """Equivalent misorientations
        Returns
        -------
        Misorientation
        """
        Gl, Gr = self._symmetry
        if Gl._tuples == Gr._tuples:  # Grain exchange case
            orientations = Orientation.stack([self, ~self]).flatten()
        equivalent = Gr.outer(orientations.outer(Gl))
        return self.__class__(equivalent).flatten()
```

dnjohnstone:
@pc494 do you know how to fix this? It would be good to have tied up for 0.2.0 release....

pc494:
Short answer, no not properly. Correct answer, yes I could find out.

PR #24, opened 2019-12-19
Fix for #21

pc494:
The missing case was actually the simpler one.


dnjohnstone:
Thanks very much for this!

PR #25, opened 2019-12-19
First PR addressing coverage.

pc494:
Hopefully this will take out some of the big uncovered chunks.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,5 @@
 [run]
 source = orix
-include = */orix/*
+include = /orix/*
 omit =
-     */orix/__init__.py
+     /orix/__init__.py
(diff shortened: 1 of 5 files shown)


pc494:
Because coveralls doesn't automatically play nicely with multiple version builds [1] while this PR is under work I'm going to drop the `3.6` support for now. I'll raise an issue in the pyxem repo about fixing this properly, and before this gets merged I'll put `3.6` back in. This has gone under the radar recently as when you have 100% coverage it doesn't matter.

[1] - https://docs.coveralls.io/parallel-build-webhook

pc494:
@dnjohnstone assuming you are happy with the `tqdm` workaround (installed in the test environment but not as a dependency as it's well isolated) I think this PR is ready to merge.

PR #39, opened 2020-02-10
Rewrite misorientation distance function

### LINKED ISSUE #37: Profile (and hopefully improve) the speed of our distance functions
At present the two (poorly named) distance functions (`distance_1` and `distance_2`) are somewhat limited by RAM/speed. Enhancing the speed (of the slow function) or refactoring (for memory) would be a very productive use of developers time.

shogas:
name: Rewrite misorientation distance function

**Release Notes**

    performance / developer change
    Summary: Improve memory usage and speed of misorientation distance function

**What does this PR do? Please describe and/or link to an open issue.**
Rewrite the _distance_2 function to use less memory by calculating the distance in blocks at the time. The _distance_1 function is marked as deprecated, since it is much slower and now only uses less memory for very small misorientation lists.
This resolves #37.

I have run some performance tests with the new implementation, but will run some more for larger lists of orientation and write up the results here.

It would also be great if someone with a real dataset could test that this does not change the output. In addition to running the tests, I have run equivalence tests with randomly generated orientations, but that is very likely to not hit corner cases.


orix/quaternion/orientation.py
```diff
@@ -219,23 +260,30 @@ def _distance_1(misorientation, verbose):
...
     return distance
 
-
-def _distance_2(misorientation, verbose):
-    if misorientation.size > 1e4:  # pragma no cover
-        confirm = input('Large datasets may crash your RAM.\nAre you sure? (y/n) ')
-        if confirm != 'y':
-            raise InterruptedError('Aborted')
-    from itertools import product as iproduct
+def _distance_2(misorientation, verbose, split_size=100):
```
shogas: The split size works well on my machine, but most likely not on other machines. Larger is not necessarily better because of the overhead of allocating the arrays.

orix/quaternion/orientation.py
```diff
@@ -219,23 +260,30 @@ def _distance_1(misorientation, verbose):
...
+    num_orientations = misorientation.shape[0]
     S_1, S_2 = misorientation._symmetry
-    mis2orientation = (~misorientation).outer(S_1).outer(S_1).outer(misorientation)
     distance = np.full(misorientation.shape + misorientation.shape, np.infty)
-    symmetry_pairs = iproduct(S_2, S_2)
+    split_size = split_size // S_1.shape[0]
```
shogas: Reduce the size based on the number of symmetries in the first symmetry object. Better suggestions welcome.

pc494:
This looks promising to me, I'll try and run it on more realistic data sometime towards the end of this week.

shogas:
I posted some speed/memory results in the related issue #37

pc494:
After the discussion on #37, I'm fairly happy that this works. @shogas my plan is now:

1) Make a variety of boring edits to the docstring/comments etc
2) Run this branch against the demos
3) Ask you to confirm you're happy with it
4) merge

Does that sound okay?

shogas:
Perfect.

pc494:
The conflict is caused by the formatting done in #40, I'll fix this in https://github.com/shogas/orix/pull/1

pc494:
Bump: https://github.com/shogas/orix/pull/1 and @shogas.

pc494:
This ran smoothly and so I'm going to merge it. Thanks @shogas.

Issue #36, opened 2020-01-07
Fix coveralls integration such that the line coverage is correct


pc494:
This issue arises because we make use of multiple Python versions, `pyxem` should have solution soon enough.

Issue #35, opened 2020-01-07
Add Windows CI


pc494:
This can follow the same style as was used for pyxem and diffsims.

pc494:
#44 closed this

PR #41, opened 2020-02-18
Ci improvements for travis

pc494:
This PR just does some simple things to clarify the testing base. It doesn't address #35 nor #36 - will save these for another time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,4 @@ install:
 script:
   - export MPLBACKEND=Agg
   - pytest --cov=orix
-
-after_success:
   - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 env:
   global:
     - DEPS="ipython tqdm"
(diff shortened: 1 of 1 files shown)


pc494:
https://chocolatey.org/packages/miniconda3

Looks like it might give us access to windows testing within travis which would be nice.

hakonanes:
I use it in KikuchiPy. I have adopted HyperSpy's setup for Travis where platform specific snippets in bash scripts are executed from `travis.yaml` (https://github.com/kikuchipy/kikuchipy/blob/master/.travis.yml#L56). When Travis builds Windows, I execute https://github.com/kikuchipy/kikuchipy/blob/master/ci/travis_setup_conda_windows.sh. It has worked fine for me so far.

pc494:
Thanks @hakonanes! It looks like a very clean testing system you've got there for `kikuchipy`. I'll probably do this in a separate PR, but I am very happy to be able keep all my CI in one provider.

Issue #50, opened 2020-03-16
Working with multiple axis conventions in indexation


pc494:
Specifically, currently the code works with "an orientation as a passive rotation from the crystal reference frame to the specimen reference" (see preprint) while many other adopt the opposite convention (ie - 
"orientation matrix transforms the specimen reference frame into the crystal reference frame") - this should be a fairly simple change to implement.

dnjohnstone:
Discussion from #49 

        Yeah at some point we're going to have to ask users to specify the axis convention in which their data was indexed. I think ideally that would be at load time. Unfortunately I don't think that's very well stored in the data files? So it's going to have to be a user defined input at some point I think.

    It is actually stored in many h5ebsd (not KikuchiPy's, though...), which is discussed in the h5ebsd paper by Jackson et al (https://doi.org/10.1186/2193-9772-3-4). But at least not in the .ang files from TSL, which I'm most familiar with. I think we could do what MTEX does, which is to display a warning if it is not set upon loading, and tell the user how to set it.

pc494:
From an implementation perspective, having an argument/kwarg with default set to `None` called 'convention' on `.from_euler()` seems like a sensible approach. Still checking some sources so that I can be sure the things we implement are what people expect.

pc494:
It is my belief that these two conventions correspond to the direct inverse of one another. Thus I suggest simply inverting the final quaternions produced as a conversion step. @dnjohnstone if you could check that you agree that would be great (I used Equation (1) from the preprint & another read through of the Rowenhorst paper to arrive here).

pc494:
#55 closed this.

PR #55, opened 2020-04-20
Improving .from_euler() method

pc494:
Supercedes #54 

1) Allows the user to specify whether a rotation (euler angles) is `lab2crystal` or `crystal2lab` (see #50)
2) Moves the inter-conversion of euler/quat onto a specific set of conventions, defined in the literature - the paper can be found at: https://iopscience.iop.org/article/10.1088/0965-0393/23/8/083501/meta


pc494:
Will check the demos and then ping you for a merge.

pc494:
Yep, demos look good @dnjohnstone

Issue #58, opened 2020-05-14
0.3.0 release


pc494:
As mentioned https://github.com/pyxem/pyxem/pull/596 we have 'enough' for a new minor version of `orix`, this issue serves to collate developer request for things that it would be good to have in by the release point - @hakonanes 

From my perspective

- [x] Clean up the deprecations from the `0.2` series 
- [x] Improve the documentation for `OrientationRegion`
- [x] Address #57 
- [x] Address the review comment on #60
- [x] Add reader/writer for CrystalMap #59
- [x] check the "grain-exchange" (GE) case of the `.equivalent` method, add path for `S1==S2` no GE

I'll add more things as they occur to me.

dnjohnstone:
@pc494 and @hakonanes - could we set a target date for 0.3.0 release, please?

I'm thinking end of June might be realistic?

Also - do we think we want to put #62 in this one?

hakonanes:
Yes, #62 please. I would like to couple it with stereopgrahic projections of EBSD master patterns in kikuchipy's `EBSDMasterPattern` class. I want to use orix a lot more in general in kikuchipy v0.3 (currently just released v0.2), so doing #59 is high on my list. End of June sounds good to me.

dnjohnstone:
Not necessarily for 0.3.0 - but for the future - I think we should think about which packages we currently know of wanting to use orix upstream and what for.

In particular we have - pyxem, kikuchipy and diffsims , which would all benefit from different parts of orix functionality.

Not sure I'm actually saying anything here - just trying to map things out in my head an thought it might be good to make some kind of a plan.

hakonanes:
Desired uses for kikuchipy, from the top of my head:
* Import orientation data along an EBSD scan into a `CrystalMap` object available as an attribute `s.crystal_map` or just `s.cm`. Then, we can make all maps available as Signal2D navigators e.g., to inform analysis. Also, to create masks, e.g. get one averaged pattern per grain ID etc.
* Store crystal data in `Phase` objects via metadata.
* Save quality maps or other maps produced from diffraction patterns via a `CrystalMap` HDF5 writer, for general saving and use in e.g. MTEX. (When the HDF5 reader/writer is done, I plan to make a PR to MTEX with an orix reader. An orix writer would also be nice, but we'll see.)
* Orientation sampling (have to move stuff over from pyxem to orix for this? Quite a lot of work?), to get an orientation library to create a diffraction pattern library from an EBSD master pattern.
[...]

PR #60, opened 2020-05-19
0.2.3 commits merging up into master

pc494:
These late addition to the 0.2.x series resolve a couple of issues. Specifically:

1) Provide support for the import convention used in the demos
2) Correct a sign error in the `.get_plot_data() `method of `OrientationRegion`
3) Avoid a round-off error by comparing to 1e-9 precision in the overriden `__gt__ `method or `OrientationRegion`

The 0.2.3 release will be built off a branch, these can then be merged (pending some small packaging changes).

### DIFF
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -264,7 +264,8 @@ class Rotation(Quaternion):
         Parameters
         ----------
         convention : 'bunge'
-            The Euler angle convention used. Only 'bunge' is supported as present
+            The Euler angle convention used. Only 'bunge'
+            is supported as present
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


pc494:
@dnjohnstone 

Can we hold off merging this into master until I've packaged 0.2.3 please? If you could check you can use this code to run the demos that would be ideal. You should only need to change the syntax on the import to have the convention kwarg set to "Krakow_Hielscher" I think

dnjohnstone:
> @dnjohnstone
> 
> Can we hold off merging this into master until I've packaged 0.2.3 please? If you could check you can use this code to run the demos that would be ideal. You should only need to change the syntax on the import to have the convention kwarg set to "Krakow_Hielscher" I think

Sure - do you want to flag the things you want merging or else just merge yourself in this instance?

pc494:
If you can confirm the demo works on your machine I'll sort the rest (and merge my own PRs in this case)

orix/quaternion/rotation.py
```diff
@@ -325,10 +326,33 @@ def from_euler(cls, euler, convention="bunge", direction="crystal2lab"):
...
+            )
+            qgamma = Quaternion(
+                np.stack((np.cos(gamma / 2), zero, zero, np.sin(gamma / 2)), axis=-1)
+            )
+            data = qalpha * qbeta * qgamma
+            rot = cls(data.data)
+            rot.improper = zero
+            return rot
+
         if convention != "bunge":
```
dnjohnstone, line 356: Not sure this is the right condition now that we have Krakow_Hielscher also?

pc494: Pushed to #58

PR #62, opened 2020-05-21
Adding pole figure functionality

dnjohnstone:


hakonanes:
This is awesome. I just wrote a reader for EMsoft's EBSD master patterns and an `EBSDMasterPattern` class (Signal2D). I think this will couple really nicely with the stereographic projection of the master pattern,.

dnjohnstone:
@EirikOpheim I'd be curious to know if this helps with your IPF plotting plans?

orix/plot/stereographic_projection.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def can_zoom(self):
+        return False
+
+    def can_pan(self):
+        return False
+
+    def transform(self, xs):
+        from texpy.quaternion.rotation import Rotation
+        from texpy.vector import Vector3d
```
dnjohnstone, line 193: make this orix

EirikOpheim:
> @EirikOpheim I'd be curious to know if this helps with your IPF plotting plans?

This looks interesting! I have a routine now where I export pyxem results, load them in MatLab and use MTEX with a script from Håkon to plot an IPF. It works fine for my purposes, but it would be easier if everything could be done in pyxem.  

What I do in the MTEX part of the workflow is use a list of orientations with a corresponding score for every orientation, for instance the reliability index, and a crystal symmetry to generate a heat map of the score on the IPF. It looks like I can use the code in this PR to transform a Rotation object into angles that can be plotted on the IPF, and also make a scatter plot, am I right? I will be happy to try this out when it is merged! :)

hakonanes:
I'm looking into implementing the gnomonic projection in kikuchipy (https://github.com/kikuchipy/kikuchipy/pull/186). However, I think that the projection (not the pattern centre or detector classes) better fit into orix. We could create a `orix.projections` subpackage, where the stereographic and gnomonic projections live. What do you think?

dnjohnstone:
> I'm looking into implementing the gnomonic projection in kikuchipy ([kikuchipy/kikuchipy#186](https://github.com/kikuchipy/kikuchipy/pull/186)). However, I think that the projection (not the pattern centre or detector classes) better fit into orix. We could create a `orix.projections` subpackage, where the stereographic and gnomonic projections live. What do you think?

I think I'm good with that. What do you intend to use the gnomonic projection for?

hakonanes:
The natural coordinate system for an EBSD detector, and hence an EBSD pattern, is the gnomonic projection with the pattern centre's x and y as the projection's centre. I want to calibrate my `EBSD` object's signal axes with a pattern centre, so that all detector pixels, and thus pattern features, can be described in this projection, and move on from there.

pc494:
I'm not yet sold on having projection code that interacts with experimental patterns placed within `orix`. But I may have misunderstood your meaning @hakonanes

hakonanes:
I understand. The gnomonic projection is a spherical projection, which could benefit from inhereting some functionality of a more general spherical projection class. I don't know if there are any plans to generalize implementations of projections?

I think I can keep it in kikuchipy for now, and see if it fits in the future.

dnjohnstone:
I'm pushing this to `future` because I don't think we need it immediately and doing it properly is more tied up with handling crystal geometry properly, which is a more general topic being discussed elsewhere in various issues. So unless anyone really wants this I think it's best if we just leave it for now and get 0.4.0 out with the various other developments that have been made. Is that ok with you @pc494 and @hakonanes ?

hakonanes:
Yep, that's okay with me. This will be really useful, so should be a priority for 0.5 I think.

PR #75, opened 2020-07-17
Housekeeping on grain exchange case and docstrings

### LINKED ISSUE #57: Misorientation/Orientation set_symmetry() docstring examples and code produce different results
After #50 (c5daae0bf1b3225b14a7aeac903babf1e0ad34ce), but perhaps also before, I do not get the results from the docstrings examples for `Misorientation.set_symmetry()` and `Orientation.set_symmetry()`:

Docstring:

```python
>>> from orix.quaternion.orientation import Misorientation, Orientation
>>> from orix.quaternion.symmetry import C4, C2
>>> data = np.array([[0.5, 0.5, 0.5, 0.5], [0, 1, 0, 0]])
>>> Misorientation(data).set_symmetry(C4, C2)
Misorientation (2,) 4, 2
[[-0.7071  0.     -0.7071  0.    ]
 [ 0.      0.7071 -0.7071  0.    ]]
>>> Orientation(data).set_symmetry(C4)
Orientation (2,) 4
[[-0.7071  0.     -0.7071  0.    ]
 [ 0.     -0.7071 -0.7071  0.    ]]
```

Actual:

```python
>>> from orix.quaternion.orientation import Misorientation, Orientation
>>> from orix.quaternion.symmetry import C4, C2
>>> data = np.array([[0.5, 0.5, 0.5, 0.5], [0, 1, 0, 0]])
>>> Misorientation(data).set_symmetry(C4, C2)
Misorientation (2,) 4, 2
[[-0.7071  0.7071  0.      0.    ]
 [ 0.      1.      0.      0.    ]]
>>> Orientation(data).set_symmetry(C4)
Orientation (2,) 1, 4
[[-0.7071  0.     -0.7071  0.    ]
 [ 0.      1.      0.      0.    ]]
```

What is wrong, the docstring or the code?

We should also make a `Orientation.__repr__()`, since it is with reference to `C1` and only has one associated symmetry.

pc494:
Will close #57 and #58 

Purpose
-----------

Corrects the docstring error touched on in #57 
Sets up the grain exchange case for the `.equivalent` method.

Comments
--------------

To make this work, `equivalent` was changed from a `@property` to a method. This allows the user to specify grain exchange at runtime, but isn't the style in which most of `orix` is written. @hakonanes, does `CrystalMap` have a natural preference here?

### DIFF
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -127,12 +127,11 @@ class Misorientation(Rotation):
         """tuple of Symmetry"""
         return self._symmetry
 
-    #@property
     def equivalent(self,grain_exchange=False):
         """Equivalent misorientations
 
         grain_exchange : bool
-            If true the rotation g and g^{-1} are considered the identical
+            If true the rotation g and g^{-1} are considered to be identical
 
         Returns
         -------
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -128,8 +128,11 @@ class Misorientation(Rotation):
         return self._symmetry
 
     @property
-    def equivalent(self):
+    def equivalent(self,grain_exchange=False):
         """Equivalent misorientations
+
+        grain_exchange : bool
+            If true the rotation g and g^{-1} are considered the identical
 
         Returns
         -------
@@ -137,10 +140,12 @@ class Misorientation(Rotation):
 
         """
         Gl, Gr = self._symmetry
-        if Gl._tuples == Gr._tuples:  # Grain exchange case
-            orientations = Orientation.stack([self, ~self]).flatten()
+
+        if grain_exchange and (Gl._tuples == Gr._tuples):
+             orientations = Orientation.stack([self, ~self]).flatten()
         else:
             orientations = Orientation(self)
+
         equivalent = Gr.outer(orientations.outer(Gl))
         return self.__class__(equivalent).flatten()
 
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -166,8 +166,8 @@ class Misorientation(Rotation):
         >>> m = Misorientation(data).set_symmetry(C4, C2)
         >>> m
         Misorientation (2,) 4, 2
-        [[-0.7071  0.     -0.7071  0.    ]
-         [ 0.      0.7071 -0.7071  0.    ]]
+        [[-0.7071  0.7071  0.      0.    ]
+        [ 0.      1.      0.      0.    ]]
 
         """
         symmetry_pairs = iproduct(Gl, Gr)
@@ -270,7 +270,7 @@ class Orientation(Misorientation):
         >>> o
         Orientation (2,) 4
         [[-0.7071  0.     -0.7071  0.    ]
-         [ 0.     -0.7071 -0.7071  0.    ]]
+        [ 0.      1.      0.      0.    ]]
 
         """
         return super(Orientation, self).set_symmetry(C1, symmetry)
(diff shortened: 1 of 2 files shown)


orix/quaternion/orientation.py
```diff
@@ -127,20 +127,25 @@ def symmetry(self):
         """tuple of Symmetry"""
         return self._symmetry
 
-    @property
-    def equivalent(self):
+    #@property
```
hakonanes, line 130: Is it anticipated that this will become a property again in the future? If not, we could remove this line.

orix/quaternion/orientation.py
```diff
@@ -127,20 +127,25 @@ def symmetry(self):
...
+    def equivalent(self,grain_exchange=False):
```
hakonanes, line 131: Spaces are nice. Running black will take care of all this.

orix/quaternion/orientation.py
```diff
@@ -127,20 +127,25 @@ def symmetry(self):
...
         """Equivalent misorientations
 
+        grain_exchange : bool
+            If true the rotation g and g^{-1} are considered the identical
```
hakonanes, line 135: ... to be identical?

orix/quaternion/orientation.py
pc494: I commented it out because I considered it an open question with the PR. If we keep it as a property, we would also make `grain_exchange` a property, which would get overridden for `Orientations` (to `False`). If not I would drop the "`#@property`"

orix/quaternion/orientation.py
pc494: yep, was going to do that after review to keep the diffs small (although it sounds like your PR's are all blacked anyway so maybe that was needless on my part)

orix/quaternion/orientation.py
pc494: yeah, this is badly worded, will fix.

orix/quaternion/orientation.py
hakonanes: We could people force running black before commiting via pre-commit: https://pre-commit.com/. We use it for kikuchipy, I can set it up here if it is desirable.

orix/quaternion/orientation.py
hakonanes: Okay. I haven't directly used this functionality, so whatever you find best.

orix/quaternion/orientation.py
dnjohnstone: seems like a useful idea to me!

orix/quaternion/orientation.py
dnjohnstone: I think the method looks alright

pc494:
Cool, will loop back round to this at the end of the day and finish it up and request a review on it.

dnjohnstone:
Assuming you're on it with travis, then I think good to go.

pc494:
The travis is a separate thing #76, which we can address and then put this in.

orix/tests/test_orientation.py
```diff
@@ -91,17 +91,17 @@ def test_getitem(orientation, symmetry):
...
-    Gl == C4 ~ "grain exchange"
-    Gl == C2 ~ "no grain exchange"
+    Gl == C4 is grain exchange
+    Gl == C2 is no grain exchange
     """
     m = Misorientation([1, 1, 1, 1])  # any will do
     m_new = m.set_symmetry(Gl, C4, verbose=True)
     m_new.symmetry
-    _m = m_new.equivalent
+    _m = m_new.equivalent(grain_exchange=True)
```
hakonanes, line 104: Sorry I didn't mention this the first time. How about adding a check like 

```python
assert np.allclose(_m.size, 32)
```

for C4 and 8 for C2?

pc494: also good, will add

PR #82, opened 2020-07-20
Release 0.3.0

dnjohnstone:


hakonanes:
> I am fairly certain we don't need to pin the matplotlib version. Other than that LGTM.

If we don't I think, I have to make a check for Matplotlib versions in my test where I look for where whether a scalebar was plotted or not. I.e. the test might fail now.

hakonanes:
The above mentioned test fails with Matplotlib v3.2 now. The part of the test which fails checks whether the scalebar has a frame or not depending on input to `CrystalMapPlot.add_scalebar()`... I think this isn't that important a check, so I'll remove it. Should I make a new PR to master or commit to this branch?

pc494:
I think testing for the frame of a scalebar is probably a safe test to drop so that (at least for now) we can support a few version of matplotlib

dnjohnstone:
Yeah I agree - so we merge this one as is and then @hakonanes will make a new PR for the test?

Issue #84, opened 2020-07-21
orix-0.4.0


dnjohnstone:
Since orix-0.3.0 is now released it seems a good time to write a list of functionality on the road map for 0.4.0.

The following will be part of orix-0.4.0:

- [x] Porting and improving the creation of orientation space grids from diffsims to orix #71 
~Adding stereographic projection #62~
- [x] Implementation of to_matrix() etc. methods in Rotation #79 
- [x] Developer improvements (#88)
- [x] Add space group to `Phase` class (#104)

Does anyone want anything else on the list? @hakonanes @pc494

pc494:
That looks good to me.

hakonanes:
Have you guys considered creating a `Miller` class, inheriting from `Vector3d`, for handling directions with respect to crystallographic axes, or do you think this is unnecessary? I'm fidling with creating plane/direction families for a given unit cell, and it would be nice to have: a `Miller.equivalent()` method to compute the equivalent indices of planes/directions, a `Miller.length()` for computing the length of a direction in direct or reciprocal space, etc. These computations should be made with the direct/reciprocal metric tensors available in `diffpy.structure.Structure.lattice.metrics`/`diffpy.structure.Structure.lattice.reciprocal().metrics`.

I cannot think of anything more at the moment.

hakonanes:
I see there are some stuff in `diffsims.utils.sim_utils` which could be relevant.

dnjohnstone:
I think that's something that would be useful - I think we've not gone there so far primarily because a part of me thinks it shouldn't really be in orix though and should rather be in diffpy.structure or whatever crystal structure package we use.

However, we essentially need this for some of the orientation space grid options that are currently in diffsims so maybe the weight balance of effort to reward is tipping.

I do think we want to think quite hard about how we want to approach crystal geometry in general though i.e. do we want to reach out to the diffpy.structure people to try and work with them to extend that. I feel like this the kind of situation that has led to the already bad fragmentation of crystallographic python libraries.

pymatgen had some of this stuff in and I'm sure it's in cctbx somewhere but we steered away from both of those because they're behemoths.

dnjohnstone:
Is there anything else we want for 0.4.0? If not I'll set the "due date" to be the end of August?

hakonanes:
[...]

PR #89, opened 2020-07-28
[feature] gridding utils

### LINKED ISSUE #85: Uniformity of grids in SO3
Our grid generations are not uniform on SO3, this can be corrected by using the Harr measure, specifically applying a cos() term within the `grid_rzxz` functionality. Although this code is going to `orix` it seems correct to fix this particular bit now.

pc494:
As detailed elsewhere #84 we're porting (some) of the gridding functionality from `diffsims` to `orix`. Because this change will allow a significant clean up of diffsims I'm PRing to `0.3.x`, as the new files won't touch any `0.3` code. I will also bump the matplotlib version in `setup.py`. Then a "quick" `0.3.1` release here will solve that problem.

Release Notes:
orix can now produce uniform grids of rotations around a fixed rotations, or within a stated fundamental zone.

Notes:
This will close #85

### DIFF
--- a/orix/__init__.py
+++ b/orix/__init__.py
@@ -1,5 +1,5 @@
 __name__ = "orix"
-__version__ = "0.3.1dev"
+__version__ = "0.4.0dev"
 __author__ = "Ben Martineau, Phillip Crout, Håkon Wiik Ånes, Duncan Johnstone"
 __author_email__ = "pyxem.team@gmail.com"
 __description__ = (
--- a/orix/gridding/grid_generators.py
+++ b/orix/gridding/grid_generators.py
@@ -15,3 +15,5 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with orix.  If not, see <http://www.gnu.org/licenses/>.
+
+""" This user facing code generates 'grids' in orientation spaces """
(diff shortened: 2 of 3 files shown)


dnjohnstone:
Think this should be a PR to master not 0.3.x?

pc494:
I was planning to build a release off it, but if nobody is writing new code I can merge it to master and do it that may. I just noticed we had made a 0.3.x branch on this occasion so it seemed a shame to waste the chance.

dnjohnstone:
Mmm I'm not really that bothered, just never sure if we "should" include new stuff in "patch" release numbers. I.e. would a small 0.4.0 release be "more correct"?

pc494:
> Mmm I'm not really that bothered, just never sure if we "should" include new stuff in "patch" release numbers. I.e. would a small 0.4.0 release be "more correct"?

You are right. I had been trying to run `orix` on a "if the minor version doesn't change all the code will run" ploicy, which is why, for example, I maintained the `0.2.x` a bit, but if we want that we should agree to make a `1.0.0`, and we aren't really there yet.

Also our experience with pyxem suggests the version numbers don't rocket that quickly, even if some versions are "not that new". Will repull to master.

Issue #86, opened 2020-07-21
get_grid_stereographic() for rotation_list_generators seems to have a bug for input ('cubic'


EirikOpheim:
**Describe the bug**
The function get_grid_stereographic() for rotation_list_generators seems to have a bug for input ('cubic'). By exporting every orientation in the orientation list and plotting in MTEX I get the following picture: 

![cubic2083randofAll_Z](https://user-images.githubusercontent.com/34092703/85163951-e7484100-b263-11ea-962c-3c87ceb1e1dc.png)
There is an area that is not covered close to the [-1 1 1] corner. 
![cubic2083randofAll](https://user-images.githubusercontent.com/34092703/85163967-ed3e2200-b263-11ea-9edd-8f8d59462be8.png)
The x-projection looks fine.

**To Reproduce**
1. Run rotation_list_generators.get_grid_stereographic()
2. Export list of rotations to MTEX
3. Plot orientations in MTEX using plotIPDF

**Expected behavior**
I believe the full z-projection should be covered.

**Screenshots**
For option 'orthorhombic' there is no issue:
![orthorhombic12500ofAll](https://user-images.githubusercontent.com/34092703/85164352-88cf9280-b264-11ea-8f09-803d62794ff7.png)
![orthorhombic12500ofAll_X](https://user-images.githubusercontent.com/34092703/85164364-8c631980-b264-11ea-8734-7abbd80aa6e6.png)

Furthermore, (and I am not sure if this is an issue) if I project the rotation list for symmetry (-43m) I get the following result:
![Cubic_m43mz](https://user-images.githubusercontent.com/34092703/85165537-460eba00-b266-11ea-90e8-7d4d17c64716.png)
Good apart from the issue near the corner.
![Cubic_m43mx](https://user-images.githubusercontent.com/34092703/85165546-4909aa80-b266-11ea-93da-6156ccd35ff0.png)
Is this an oversampling?
Should only half the IPF be covered?

PR #90, opened 2020-07-28
[feature] gridding utils

### LINKED ISSUE #85: Uniformity of grids in SO3

pc494:
#89 but to the correct branch.

As detailed elsewhere #84 we're porting (some) of the gridding functionality from diffsims to orix. I will also bump the matplotlib version in setup.py.

Release Notes:
orix can now produce uniform grids of rotations around a fixed rotations, or within a stated fundamental zone.

Notes:
This will close #85


dnjohnstone:
Will this address #86 also?

pc494:
#86 is going back to diffsims I think, as I'm not doing stereographic projection stuff in orix (but need a few of the `utils` that I had made - in diffsims-  for the ZAP map generator).

dnjohnstone:
> #86 is going back to diffsims I think, as I'm not doing stereographic projection stuff in orix (but need a few of the `utils` that I had made - in diffsims- for the ZAP map generator).

Ok - when this one is done, I think maybe we should remove the stereographic_projection stuff from diffsims since we can use this fundamental zone stuff instead and just leave the issue in orix as "for the future" functionality?

pc494:
Yep, just wanted to do it carefully so as not to break the zap map code. Normal template matching should be done with fundamental zone grids.

pc494:
Just in case github doesn't auto-ping on this, @hakonanes I'm ready for this to go to review :)

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+# orix is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with orix.  If not, see <http://www.gnu.org/licenses/>.
+
+""" This user facing code generates 'grids' in orientation spaces """
+
```
dnjohnstone, line 20: Module for generating `grids` in orientation spaces.

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+from orix.gridding.gridding_utils import (
+    create_equispaced_grid,
+    _get_proper_point_group,
+)
+from orix.quaternion.orientation_region import OrientationRegion
+
+def get_grid_fundamental(resolution, point_group=None, space_group=None):
+    """
+    Generates a grid of rotations that lie within a fundamental zone
```
dnjohnstone, line 31: Generates an equispaced grid of rotations within a fundamental zone.

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+        The smallest distance between a rotation and its neighbour (degrees)
+    point_group : orix.symmettry
+        One of the 11 proper point groups
+    space_group: int
+        Between 1 and 231
+
+    Returns
+    -------
+    q : orix.Rotation
+        grid of rotations the lie within the specified fundamental zone
```
dnjohnstone, line 45: Grid of rotations lying within the specified fundamental zone.

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+def get_grid_local(resolution, center, grid_width):
+    """
+    Generates a grid of rotations about a given rotation
+
+    Parameters
+    ----------
+    resolution : float
+        The smallest distance between a rotation and its neighbour (degrees)
+    center : orix.Rotation
+        The rotation to act as the center of the grid
```
dnjohnstone, line 73: The rotation at which the grid is centered.

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+    grid_width :
+        The largest angle of rotation away from center that is acceptable (degrees)
+
+    Returns
+    -------
+    q : orix.Rotation
+        grid of rotations the lie within grid_width of center
```
dnjohnstone, line 80: Grid of rotations lying within `grid_width` of `center`

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018-2020 The pyXem developers
```
hakonanes, line 2: Should it be changed from pyxem to orix developers in the future if orix developers contribute to this?

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+
+def get_grid_fundamental(resolution, point_group=None, space_group=None):
+    """
+    Generates a grid of rotations that lie within a fundamental zone
+
+    Parameters
+    ----------
+    resolution : float
+        The smallest distance between a rotation and its neighbour (degrees)
+    point_group : orix.symmettry
```
hakonanes, line 37: orix.quaternion.symmetry.Symmetry? If this is supposed to be picked up by an autogenerated API via sphinx-doc we need the "full path".

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+        One of the 11 proper point groups
+    space_group: int
+        Between 1 and 231
+
+    Returns
+    -------
+    q : orix.Rotation
```
hakonanes, line 44: orix.quaternion.rotation.Rotation?

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+        grid of rotations the lie within the specified fundamental zone
+
+    See Also
+    --------
+    orix.gridding.utils.create_equispaced_grid
+
+    Examples
+    --------
+    >>> from orix.quaternion.symmetry import C2,C4
+    >>> grid = get_grid_fundamental(1,point_group=C2)
```
hakonanes, line 54: A whitespace.

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+
+import numpy as np
+
+from orix.gridding.gridding_utils import (
+    create_equispaced_grid,
+    _get_proper_point_group,
+)
+from orix.quaternion.orientation_region import OrientationRegion
+
+def get_grid_fundamental(resolution, point_group=None, space_group=None):
```
hakonanes, line 29: Should we provide some defaults for `resolution`? EMsoft has a default of about 1,5 degrees for their homochoric orientation sampling of orientation space for space group 225.

orix/gridding/grid_generators.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    if point_group is None:
+        point_group = _get_proper_point_group(space_group)
+
+    q = create_equispaced_grid(resolution)
+    fundamental_region = OrientationRegion.from_symmetry(point_group)
+    return q[q < fundamental_region]
+
+
+def get_grid_local(resolution, center, grid_width):
```
hakonanes, line 64: Same here, defaults?

setup.py
```diff
@@ -28,7 +28,7 @@
     install_requires=[
         "diffpy.structure >= 3",
         "h5py",
-        "matplotlib",
+        "matplotlib > 3.2",
```
hakonanes, line 31: Why not >= 3.3?

pc494:
I'll fix all the small ones now, and then we can discuss EMSoft checks when @hakonanes is back from holiday.

orix/gridding/grid_generators.py
pc494: Went with 2 here.

orix/gridding/grid_generators.py
pc494: and here, with `2`, `None` and `10`, with a slightly code change to map None ---> identity rotation

setup.py
pc494: carelessness, fixed

orix/gridding/grid_generators.py
pc494: Good question, to be addressed somewhere else.

hakonanes:
Many things:
1. What is the reference for the `get_fundamental_zone()` implementation? A piece of code, or literature?

2. It's called the Haar measure, right? Should be corrected in `create_equispaced_grid()` docstring.

3. How about calling the module `sampling`, yielding `orix.sampling.get_fundamental_zone()`?

4. Sampling orientations for the point group 32 with a resolution of 2.2 degrees in EMsoft (`EMSampleRFZ`), MTEX (`equispacedSO3Grid`) and orix (`get_grid_fundamental`) and analysing the resulting ODF in MTEX yields (texture index, max, entropy*):
* EMsoft, 72 227 orientations (1.0000, 1.0223, -0.0000239)
* MTEX, 54 514 orientations (1.0000, 1.0125, -0.00000381)
* orix, 161 914 orientations (1.0028, 1.0406, -0.0016)

*entropy: "Another measure of the overall texture strength is the entropy S, which actually measures the deviation from the uniform texture or lack of ‘organization’ of orientations" (https://dx.doi.org/10.1107/S0021889806055476)

This leads me to conclude that the uniformity of the orientation sampling with `get_fundamental_zone` is pretty good but not as good as EMsoft/MTEX, and that the function oversamples. What do you think?

5. I've made a naive Python implementation (for loops) of EMsoft's `SampleRFZ` as explained in Singh and De Graef (2016) (https://dx.doi.org/10.1088/0965-0393/24/8/085013) by adapting the Fortran code from here (https://github.com/EMsoft-org/EMsoft/blob/develop/Source/Utilities/EMsampleRFZ.f90#L77). Only for cubic fundamental zone so far, but should be OK to get the remaining zone shapes. I hope to include it as part of `orix.sampling/gridding` after this PR is done. It includes several of the rotation conversions (https://github.com/EMsoft-org/EMsoft/blob/develop/Source/EMsoftLib/rotations.f90), namely cu2ro, cu2ho (as explained in Appendix A in the paper and this code https://github.com/EMsoft-org/EMsoft/blob/develop/Source/EMsoftLib/Lambert.f90#L952), ho2ro, ho2ax, ax2ro, ro2eu, ro2om, om2eu, ro2ax, and ax2om. Many of these are already in orix, I will adapt my implementation to existing functionality. Question: The cubochoric representation is not neo-Eulerian, where should we put it?

dnjohnstone:
>     4\. Sampling orientations for the point group 32 with a resolution of 2.2 degrees in EMsoft (`EMSampleRFZ`), MTEX (`equispacedSO3Grid`) and orix (`get_grid_fundamental`) and analysing the resulting ODF in MTEX yields (texture index, max, entropy*):
> 
> 
>     * EMsoft, 72 227 orientations (1.0000, 1.0223, -0.0000239)
> 
>     * MTEX, 54 514 orientations (1.0000, 1.0125, -0.00000381)
> 
>     * orix, 161 914 orientations (1.0028, 1.0406, -0.0016)

Does this suggest that the MTEX implementation is the "best"? Is there a reason we haven't followed the implementation there, given how similar to MTEX our code base is?

hakonanes:
> Does this suggest that the MTEX implementation is the "best"? Is there a reason we haven't followed the implementation there, given how similar to MTEX our code base is?

I'm not qualified to answer that... And I haven't looked at the MTEX sampling (https://github.com/mtex-toolbox/mtex/blob/develop/geometry/geometry_tools/equispacedSO3Grid.m). In figure 4 in their paper, Singh and De Graef shows that a sampling of about 1,5 degrees is necessary to obtain a significant dot product value, i.e., the chance that there are some simulated patterns in the dictionary that stand out, so that ordering the matches from best to worst has meening. I haven't indexed experimental patterns with a dictionary populated by patterns of orientations from MTEX, only EMsoft.

pc494:
I think, assuming people agree what the word "resolution" means (by no means certain), only 1 of the answers can be correct. I've had a look at the MTEX implementation, and I'm satisfied that my implementation is "better", where "better" comes from flexibility and code simplicity. Creating a uniform grid and cutting the appropriate bit out is a very modular approach, and it takes almost no lines new lines of code too!

I think the best approach is probably to run the implementations and check against some definitions of a "uniform grid of resolution x"

Uniform ===> integrating any region of a fixed size yields the same result (which I think is already in the tests)
"resolution x" is up for discussion, but I think, in euler space, saying that a "resolution of 1 degree" is such that the dalpha and dgamma terms in the volume element:

* `dV = (1/8) * sin beta dalpha dbeta dgamma ` [1]

are both equal to 1. So within 0.9degrees of a fixed rotation you would find zero rotations. If we write those down we could trial all the routes?

[1] - https://math.stackexchange.com/questions/3316481/

pc494:
@hakonanes, any chance you could drop over the files with the orientations you pulled out of the codes here? Couple of things I'm curious to check, but I don't quite fancy getting working Matlab and Fortran just yet if I don't have to.

Similarly if you've got the raw python code version of the EMSoft code I wouldn't mind scanning that, turn out I really can't read Fortran code quickly.

hakonanes:
> @hakonanes, any chance you could drop over the files with the orientations you pulled out of the codes here? Couple of things I'm curious to check, but I don't quite fancy getting working Matlab and Fortran just yet if I don't have to.

EMsoft:
* name list file for `EMsampleRFZ` command line program [samplerfz.nml.txt](https://github.com/pyxem/orix/files/5029459/samplerfz.nml.txt) (remove `.txt` to run as `EMsampleRFZ samplerfz.nml`)
* orientations [pg32_n60_emsoft.txt](https://github.com/pyxem/orix/files/5029433/pg32_n60_emsoft.txt)

orix:
* to reproduce:
```python
from orix.symmetry import Oh
from orix.gridding import get_fundamental_zone
oris = get_fundamental_zone(2.273, Oh)
```
* orientations [pg32_res2_273_orix.txt](https://github.com/pyxem/orix/files/5029435/pg32_res2_273_orix.txt)

MTEX 5.3.1:
* to reproduce:
```Matlab
cs = crystalSymmetry('Oh')
oris = equispacedSO3Grid(cs, 'resolution', 2.273 * degree)
oris.export('pg32_res2_273_mtex.txt)
```
* orientations [pg32_res2_273_mtex.txt](https://github.com/pyxem/orix/files/5029436/pg32_res2_273_mtex.txt)

> 
> Similarly if you've got the raw python code version of the EMSoft code I wouldn't mind scanning that, turn out I really can't read Fortran code quickly.

Implementation taken more or less directly from EMsoft source code (relevant files linked in the text file) [emsoft_samplerfz.py.txt](https://github.com/pyxem/orix/files/5029516/emsoft_samplerfz.py.txt)

> 
> Addition: (not to sound defensive but) looking in https://iopscience.iop.org/article/10.1088/0965-0393/24/8/085013 and Table 3, reading along to <\theta>^{32} and down to 2.2 degrees we get (more or less) and N value of 60. Then from Table 1 / Table 2 do we not have this as c.1 700 000 points. Do we understand this discrepancy?

1 700 000 cubochoric points are sampled, and after conversion to Homochoric and then Rodrigues-Frank, the points outside the RF fundamental zone are discarded. This actually lead me to wonder if we could only implement cubochoric->Homochoric->Rodrigues-Frank->quaternion and use your `q[q < fundamental_region]`?

pc494:
So I've had a little think about this, and I think in general we are on the same page. I think my next steps will be:

1) rename this to "sampling" 
2) I think they higher point count is due to a bug, which I'll have a hunt for (if we find that it would clear a lot of stuff up)
3) Having done both of those, decide if we want `create_equispaced_grid` to run off the cubochoric rep (cubo-->homo-->quaternion) or the Euler method (as currently implemented) - the modularity of what we have means that we can defer this decision until after (1) and (2)

pc494:
I think I have found the bug, if the samples created post https://github.com/pyxem/orix/pull/90/commits/6495611e76d1d7d4599ba12e8a1ae4b234378eb2 are still uniform (according to MTEX) I think they will have a sensible number of `n` ~ 161 914 / 2

Assuming that has worked I would hope to wrap this (and my review of #92) early next week which puts us in a good place wrt v0.4

hakonanes:
Pulling the recent changes and sampling point group 32 with `get_sample_fundamental(2.273, Oh)` I get 83 693 orientations. The ODF computed with MTEX with default parameters has texture index 1.0084, entropy -0.0046 and max intensity 1.1160. Before https://github.com/pyxem/orix/commit/6495611e76d1d7d4599ba12e8a1ae4b234378eb2, these four numbers were 161 914 1.0028, -0.0016, 1.0406.

The number of orientations is halfed, while the sampling seems to retain most of its uniformity!

hakonanes:
Plotting 10 000 random orientations from the sample with MTEX:
Orix (83 693 orientations) | EMsoft (72 227) | MTEX (54514)
:----:|:----------:|:---------:
sampling_orix | sampling_emsoft | sampling_mtex

hakonanes:
I would just like to note that I hope we (I) can implement the orientation sampling procedure from EMsoft as well... Most likely as part of a later release, as my hack included above requires multiple optimizations and generalising to all crystal systems.

pc494:
Okay, my feelings are:

- I can tidy this up to something that we can merge, with functionality as is
- The different number of orientations is due to ambiguous definitions of "resolution"
- The function `uniform_sample_SO3` could easily have a kwarg added in a future release for other methods (eg EMsoft's method)

I propose the tidy here and merge, and I'll raise a new issue to cover the latter two bullet points.

hakonanes:
> I propose that I perform the tidy up here and get it merged, and I'll raise a new issue to cover the latter two bullet points.

Sounds good.

pc494:
This is now ready to review :)

orix/sampling/sample_generators.py
```diff
@@ -0,0 +1,93 @@
...
+from orix.quaternion.orientation_region import OrientationRegion
+
+
+def get_sample_fundamental(resolution=2, point_group=None, space_group=None):
+    """
+    Generates an equispaced grid of rotations within a fundamental zone.
+
+    Parameters
+    ----------
+    resolution : float
```
hakonanes, line 36: Should state that these are optional? Less important, although common, is to state the default value in the parameter explanation?

orix/sampling/sampling_utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+def uniform_SO3_sample(resolution):
+    """
+    Returns rotations that are evenly spaced according to the Haar measure on
+    SO3
+
+    Parameters
+    ----------
+    resolution : float
+        The characteristic distance between a rotation and its neighbour (degrees)
```
hakonanes, line 81: "characteristic" changed to "smallest" other places, should it be changed here as well?

pc494: actually it should be characteristic almost everywhere, will change.

orix/sampling/sample_generators.py
pc494: Will include relevant optionals, where meaningful will add the default values to descriptions (although probably not in this specific case, as I'm not sure it adds a whole lot)

pc494:
I think this is good to go now

hakonanes:
I took (for the first time) the liberty of merging... If you guys @pc494 @dnjohnstone want that task to continue to be your responsibility, I understand that.

pc494:
Nope that's great, it means I can cross "figure out how to get @hakonanes made an admin on orix" off my list!

dnjohnstone:
> I took (for the first time) the liberty of merging... If you guys @pc494 @dnjohnstone want that task to continue to be your responsibility alone, I understand that.

Absolutely great @hakonanes ! I think you've got more claim to that than me these days!

hakonanes:
Glad to hear it, thanks a lot!

Issue #87, opened 2020-07-22
Get a list of symmetry operations for a space group


hakonanes:
I can get all symmetry operations for space group 229 with diffpy.structure:

```python
>>> from diffpy.structure.spacegroups import GetSpaceGroup
>>> sg229 = GetSpaceGroup(229)
>>> list_sym_ops = sg229.symop_list[:sg229.num_primitive_sym_equiv]
>>> sym_ops = np.stack([op.R for op in list_sym_ops])
>>> print(sym_ops.shape)
(48, 3, 3)
>>> print(sym_ops[0])
[[1. 0. 0.]
 [0. 1. 0.]
 [0. 0. 1.]]
```

Is it possible to get the same list from `orix.quaternion.symmetry.Symmetry`?

hakonanes:
My question is more that is there some canonical way of getting the symmetry operations for a space group from the point group operations?

dnjohnstone:
I *think* we only ever worried about point groups thus far because we weren't using the actual crystal structure anywhere before we added the diffpy dependency.

hakonanes:
Yes, so actually what you should do when you define a phase is that you should state the space group instead of the point group?

The `diffpy.structure.Structure` object doesn't actually store any symmetry information at all, and I don't understand why. Should we add it to the `Phase` object ourselves? If so, we must update this whenever we change the point group or space group...

hakonanes:
The point group is stored along with the spacegroup in diffpy.structure, so we just have to get a map from their notation to ours when we change the space group to update the point group.

The choice of space group is of course ambiguous when going the other way... I think whether a phase has a space group should depend on whether you explicitly set it or not:
* If you only set the point group, the phase does not have a space group.
* If a phase has a space group and you change the phase group, the space group is lost and you have to reset it.

dnjohnstone:
So when we talk about a crystal structure it's definitely natural that the symmetry related with it is the space group. But when we talk about symmetry equivalence of crystal orientations it's only the proper point group that matters - I think that not everyone who might potentially use orix is as familiar as they should be with this so my ideal preference would be to give as much information as possible i.e. a phase should be explicitly related to: space group, point group, and proper point group

Does that make sense?

I think the thing with diffpy.structure.Structure not storing the space group information is that it was really developed to support pair distribution function analysis and in this case you probably want to refine the structure without symmetry by default and optionally impose symmetry operations as constraints if appropriate and/or later.

I think the ideal solution for us would be that there could be:

[...]

PR #92, opened 2020-08-06
Rename Phase.symmetry to .point_group, change cm/IO accordingly

hakonanes:

Renames `Phase.symmetry` to `Phase.point_group`, in anticipation of adding `Phase.space_group`, from `diffpy.structure.spacegroupmod.SpaceGroup`, as discussed in #87. Accompanying changes are made to the `PhaseList` and `CrystalMap` classes, as well as all IO plugins.

The `CrystalMap`, `Phase` and `PhaseList` representations now look like this:

```python
>>> cm
Phase  Orientations       Name  Point group       Color
    1  5657 (48.4%)  austenite          432    tab:blue
    2  6043 (51.6%)    ferrite          432  tab:orange
Properties: iq, dp
Scan unit: um
>>> cm.phases
Id       Name  Point group       Color
 1  austenite          432    tab:blue
 2    ferrite          432  tab:orange
>>> cm.phases["austenite"]
<name: austenite. point group: 432. color: tab:blue>
>>> cm.phases["austenite"].point_group
Symmetry (24,) 432
[...]  # Many rotations
 [ 0.5    -0.5     0.5     0.5   ]]
```

Initializing `CrystalMap`, `Phase` and `PhaseList` is now done by passing a `point_group` argument instead of `symmetry`.

Issues resulting from this PR:
* Might the naming of `point_group` compared to `symmetry` be confusing for the user? I actually think it might be a better naming and help the user's understanding.  
* The orix HDF5 plugin now writes the phase point group to an HDF5 dataset with name `point_group` and not `symmetry`, with similar changes when reading. So, if this PR is merged into master, reading a phase HDF5 group  from a file written with the current writer won't work... unless we check for this in the reader. Doing this is easy. However, it raises the question of how many such small changes we should accomodate.

### DIFF
--- a/orix/crystal_map/crystal_map.py
+++ b/orix/crystal_map/crystal_map.py
@@ -51,11 +51,11 @@ class CrystalMap:
     phase_id : numpy.ndarray
         Phase IDs of points in data.
     phases : orix.crystal_map.PhaseList
-        List of phases with their IDs, names, crystal symmetry objects and
-        colors (possibly more than are in the data).
+        List of phases with their IDs, names, point groups and colors
+        (possibly more than are in the data).
     phases_in_data : orix.crystal_map.PhaseList
-        List of phases in data with their IDs, names, crystal symmetry
-        objects and colors.
+        List of phases in data with their IDs, names, point groups and
+        colors.
     prop : orix.crystal_map.CrystalMapProperties
         Dictionary with properties, like quality metrics, in each data
         point.
@@ -122,7 +122,7 @@ class CrystalMap:
             map is assumed to be 2D or 1D, and it is set to None.
         phase_list : PhaseList, optional
             A list of phases in the data with their with names,
-            symmetries and structures. The order in which the phases
+            point groups and structures. The order in which the phases
             appear in the list is important, as it is this, and not the
             phases' IDs, that is used to link the phases to the input
             `phase_id` if the IDs aren't exactly the same as in
@@ -164,7 +164,7 @@ class CrystalMap:
         ...     x=x,
         ...     y=y,
         ...     phase_name=["austenite", "ferrite"],  # Overwrites Structure.title
-        ...     symmetry=["432", "432"],
+        ...     point_group=["432", "432"],
         ...     structure=structures,
         ...     prop=properties,
         ... )
@@ -219,7 +219,7 @@ class CrystalMap:
                 # default initial values
                 phase_list = PhaseList(
                     names=phase_list.names,
-                    symmetries=phase_list.symmetries,
+                    point_groups=phase_list.point_groups,
                     colors=phase_list.colors,
                     structures=phase_list.structures,
                     ids=unique_phase_ids,
@@ -375,7 +375,7 @@ class CrystalMap:
                 rotations = self.rotations[:, 0]
             else:
                 rotations = self.rotations
-            return Orientation(rotations).set_symmetry(phases[:].symmetry)
+            return Orientation(rotations).set_symmetry(phases[:].point_group)
         else:
             raise ValueError(
                 f"Data has the phases {phases.names}, however, you are executing a "
@@ -460,9 +460,9 @@ class CrystalMap:
         A CrystalMap object can be indexed in multiple ways...
 
         >>> cm
-        Phase   Orientations       Name  Symmetry       Color
-            1   5657 (48.4%)  austenite       432    tab:blue
-            2   6043 (51.6%)    ferrite       432  tab:orange
+        Phase   Orientations       Name  Point group       Color
+            1   5657 (48.4%)  austenite          432    tab:blue
+            2   6043 (51.6%)    ferrite          432  tab:orange
         Properties: iq, dp
         Scan unit: um
         >>> cm.shape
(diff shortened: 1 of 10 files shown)


pc494:
_" The orix HDF5 plugin now writes the phase point group to an HDF5 dataset with name point_group and not symmetry, with similar changes when reading. So, if this PR is merged into master, reading a phase HDF5 group from a file written with the current writer won't work... unless we check for this in the reader. Doing this is easy. However, it raises the question of how many such small changes we should accomodate."_

I think the reader should be stable within minor versions (without any clever additions). If the hdf5 contains the version we could then support a finite number of versions at any given time. Say nine months worth?

Worst case, you've saved something with v0.3 and that version is no longer supported (say we've got to v0.6), your workaround (tedious) is that you have to load with 0.4 save to 0.4 load with 0.6 and save again or similar?

The problem gets smaller if we try to make all the breaking changes one release cycle at a time?

tl;dr ~ happy for this change, and would be inclined to support v0.3 in this instance

hakonanes:
> I think the reader should be stable within minor versions (without any clever additions). If the hdf5 contains the version we could then support a finite number of versions at any given time. Say nine months worth?

The orix version is always written to the top HDF5 group "/" in the `version` dataset. Nine months... two minor versions? Sounds good to me.

> Worst case, you've saved something with v0.3 and that version is no longer supported (say we've got to v0.6), your workaround (tedious) is that you have to load with 0.4 save to 0.4 load with 0.6 and save again or similar?

Should have mentioned this above... but the simplest solution is to open the file in HDFView (or with h5py) and rename the dataset. I do this all the time with EMsoft HDF5 files when some dataset names or values are inconsistent.

> The problem gets smaller if we try to make all the breaking changes one release cycle at a time?

Every time we release a major version? Agreed, we should adhere to semver.

> tl;dr ~ happy for this change, and would be inclined to support v0.3 in this instance

Okidoki, will add a check for the different dataset names.

hakonanes:
Instead of checking which orix version the orix HDF5 file was written with, I just used

```python
try:
    point_group = dictionary["point_group"]
except KeyError:
    point_group = dictionary["symmetry"]
```

Also added a TODO comment to remove the check in orix v0.6.

pc494:
Is this good for review now?

hakonanes:
yup, ready for review

orix/crystal_map/crystal_map.py
```diff
@@ -610,26 +610,28 @@ def __repr__(self):
...
         p_sizes = [np.where(phase_ids == i)[0].size for i in unique_phases]
         id_len = 5
-        ori_len = max(max([len(str(p_size)) for p_size in p_sizes]) + 9, 13)
-        name_len = max(max([len(n) for n in names]), 5)
-        sym_len = max(max([len(sn) for sn in symmetry_names]), 8)
-        col_len = max(max([len(i) for i in phases.colors]), 6)
+        ori_len = max(max([len(str(p_size)) for p_size in p_sizes]) + 8, 12)
+        name_len = max(max([len(n) for n in names]), 4)
+        pg_len = max(max([len(sn) for sn in point_group_names]), 11)
+        col_len = max(max([len(i) for i in phases.colors]), 5)
```
pc494, line 624: is the changing numbers here important?

orix/io/plugins/orix_hdf5.py
```diff
@@ -164,6 +164,8 @@ def dict2phase(dictionary):
     ----------
     dictionary : dict
         Dictionary with phase information.
+    version : list, optional
```
pc494, line 167: this can be deleted now?

orix/io/plugins/orix_hdf5.py
```diff
@@ -172,13 +174,17 @@ def dict2phase(dictionary):
...
     structure.title = dictionary["name"]
-    symmetry = dictionary["symmetry"]
-    if symmetry == "None":
-        symmetry = None
+    # TODO: Remove this check in v0.6.0
+    try:
+        point_group = dictionary["point_group"]
+    except KeyError:  # v0.3.0
+        point_group = dictionary["symmetry"]
+    if point_group == "None":
```
pc494, line 182: PEP 8 prefers `is None`, although it's certainly more of a convention than a necessity

orix/crystal_map/crystal_map.py
hakonanes: This ensures the `repr()` table looks good when changing `Symmetry` to `Point group`.

orix/io/plugins/orix_hdf5.py
hakonanes: If a phase has no point group set, it is written as a string "None" to the HDF5 file, an subsequently read as a string "None". So I'm pretty sure this check is correct.

hakonanes: Of course, I could add a conversion from string "None" to actually `None` in the HDF5 reader. Will look into it.

pc494: Nope, I was reading on auto-pilot, the string "None" is the way to go here. No need for further action.

orix/io/plugins/orix_hdf5.py
hakonanes: Thanks, deleted.

hakonanes:
Should be good now.

PR #96, opened 2020-08-10
copyright the orix developers

dnjohnstone:
I think @hakonanes mentioned this and it's a good point so here it is.


pc494:
yep, on board with this

PR #97, opened 2020-08-10
Add Rotation.to/from_matrix() methods to convert between quaternions and orientation matrices

hakonanes:

Release notes: Add Rotation.to/from_matrix() methods to convert between quaternions and orientation matrices.

To do:
- [x] More robust tests
- [x] Think of any edge cases or different input shapes (help wanted)

### DIFF
--- a/orix/tests/test_rotation.py
+++ b/orix/tests/test_rotation.py
@@ -1,3 +1,4 @@
+# from diffpy.structure.spacegroups import
 from math import cos, sin, tan, pi
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


pc494:
Using randomly generated quaternions and running them in a loop is sufficiently robust for me (in terms of testing) as long as we spot the edge cases.

hakonanes:
Ready for review.

orix/tests/test_rotation.py
```diff
@@ -1,3 +1,4 @@
+# from diffpy.structure.spacegroups import
 from math import cos, sin, tan, pi
```
dnjohnstone, line 2: Don't think we need the commented out diffpy import?

hakonanes: Yup, thanks.

PR #100, opened 2020-08-15
Remove possibility of sqrt of small negative numbers in Rotation.from_matrix()

hakonanes:

Subtracting 1 from diagonal elements in some rotation matrices from diffpy.structure's `SymOp.R` equal to 1 might lead to small negative numbers, causing `Rotation.from_matrix()` to complain. This should now be fixed. Added a test to check for this.

Also updated notation in `Rotation.to_matrix()` to use a, b, c, d instead of q0-3.

### DIFF
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -426,25 +426,25 @@ class Rotation(Quaternion):
         a, b, c, d = self.a.data, self.b.data, self.c.data, self.d.data
         om = np.zeros(self.shape + (3, 3))
 
-        q11 = b ** 2
-        q22 = c ** 2
-        q33 = d ** 2
-        qq = a ** 2 - (q11 + q22 + q33)
-        q12 = b * c
-        q03 = a * d
-        q13 = b * d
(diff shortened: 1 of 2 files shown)


hakonanes:
By the way, I tried using `Rotation._differentiators()` to get the a, b, c, d, aa, bb etc. arrays, but that is not the purpose of that method?

pc494:
> By the way, I tried using `Rotation._differentiators()` to get the a, b, c, d, aa, bb etc. arrays, but that is not the purpose of that method?

It looks like that ought to have worked to me, what went wrong? On the flip side, I think what you have done is probably more readable.

hakonanes:
> It looks like that ought to have worked to me, what went wrong? On the flip side, I think what you have done is probably more readable.

Yeah, it works if I transpose the `abcd` array before extracting a, b, c, d etc., since the output shape is (n rotations, 11), 11 being the a, b, c, d etc. This transposing would lead to a (small? haven't checked) speed penalty, so I didn't do it.

PR #102, opened 2020-08-17
Modify spacegroup2pointgroup conversion to include improper point groups

hakonanes:

* Changed location and name of function `orix.sampling.sampling_utils._get_proper_point_group()` to `orix.quaternion.symmetry._get_point_group()`
* Added a `proper: bool = False` parameter to `_get_point_group()`, returning either the proper or "improper" point group from the old `conversion_dict`. Since the name is now `_get_point_group()`, default is to return the improper point group.
* Added "improper" point groups to the `conversion_dict` space group to point group conversion dictionary. Also moved the dictionary to `orix.quaternion.symmetry`.
* Moved and updated tests accordingly

This PR, or a version of it, is necessary when deriving the point group from the space group in the `Phase` class. The idea is that when you define a `Phase` specifying a space group, you don't necessarily want the proper point group, just the point group.

Feel free to change the setup of the dictionary (proper/improper) or propose a change if you think it's ugly and can be done better.

I'm sure the docstring wording can be improved also.

Example usage:
```python
>>> from orix.quaternion.symmetry import _get_point_group
>>> from diffpy.structure.spacegroups import GetSpaceGroup
>>> sg225 = GetSpaceGroup(225)
>>> sg225.short_name
'Fm-3m'
>>> sg225.point_group_name
'PGm3barm'
>>> _get_point_group(225).name
'm-3m'
>>> _get_point_group(225, proper=False).name
'm-3m'
>>> _get_point_group(225, proper=True).name
'432'
```

### DIFF
--- a/orix/quaternion/symmetry.py
+++ b/orix/quaternion/symmetry.py
@@ -477,6 +477,16 @@ def _get_point_group(space_group_number, proper=False):
     -------
     point_group : orix.quaternion.symmetry.Symmetry
         One of the 11 proper or 32 point groups.
+
+    Examples
+    --------
+    >>> from orix.quaternion.symmetry import _get_point_group
+    >>> pgOh = _get_point_group(225)
+    >>> pgOh.name
+    'm-3m'
+    >>> pgO = _get_point_group(225, proper=True)
+    >>> pgO.name
+    '432'
     """
     spg = GetSpaceGroup(space_group_number)
     pgn = spg.point_group_name
--- a/orix/quaternion/symmetry.py
+++ b/orix/quaternion/symmetry.py
@@ -37,6 +37,7 @@ improper rotations. A mirror symmetry is equivalent to a 2-fold rotation
 combined with inversion.
 
 """
+from diffpy.structure.spacegroups import GetSpaceGroup
 import numpy as np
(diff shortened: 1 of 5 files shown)


hakonanes:
I should note that the update to the space group to point group conversion dictionary is based upon Table 10.2 in De Graef and McHenry's Structure of Materials.

pc494:
If this is ready for a review I'll get on it later today?

hakonanes:
Yes, it's ready for review.

I try to always remember to use the WIP label if the PR is not ready for review.

pc494:
> Yes, it's ready for review.
> 
> I try to always remember to use the WIP label if the PR is not ready for review... A better way would be for me to request a review!

Sorry, in the past we've worked the other way around (requesting reviews rather than using the WIP label) because when there are only 2 devs you know whose going to be reviewing! For `orix` will developers using your method seems the best option

hakonanes:
> Sorry, in the past we've worked the other way around (requesting reviews rather than using the WIP label) because when there are only 2 devs you know whose going to be reviewing! For `orix` developers using your method seems the best option

You might be right. We could add a `ready-for-review` label...

PR #103, opened 2020-08-17
Update Travis CI configuration YAML file

hakonanes:

Travis flagged three infos and one warning in the travis.yml config file (https://travis-ci.org/github/pyxem/orix/builds/718608014/config). Those flags should be removed after this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: shell
+os: linux
+dist: bionic
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,10 +1,12 @@
 language: shell
+os: linux
+dist: bionic
```
pc494, line 3: why this an not the travis default of `xenial`?

hakonanes: Good question. Figured we could use bionic instead of xenial since it is newer?

pc494: Sorry, was been slow, `bionic` is good

PR #104, opened 2020-08-17
Add diffpy.structure space group to Phase class

### LINKED ISSUE #99: Adding a SpaceGroup object from diffpy.structure to the Phase class
See initial discussion and motivation in issue #87. I'll submit a WIP PR tomorrow.

Questions:
1. Should we allow initializing a `Phase` only by a point group (and not providing the space group), and setting `Phase.space_group` to `None` by default? Simplest solution is to require the space group, and let the point group always be a derived property.
2. After initialization, should we allow setting the point group, or should this always be derived from the space group? (Difficult if it is `None`.) If allowed, should we set the `Phase.space_group` to `None` if the current space group is not derived from the new point group the user wants to set?
3. What symmetry info should the `__repr__()` method of `Phase`, `PhaseList` and `CrystalMap` show? As it is now they show the point group name. Should we change it to show "symmetry" with both the space group name, and point group name in parantheses? Or just the space group?
4. Should the properties be `point/space_group` or `point/spacegroup`?

I'll only add the `SpaceGroup` object to the `Phase` class, i.e. not extract any of its properties for easier access than via `phase.spacegroup.this_property`. If we want to use the `SpaceGroup.symop_list` (with `SymOp` objects) internally, we can extract the symmetry elements (rotation matrices and translation vectors) like so

```python
>>> from orix.quaternion.rotation import Rotation
>>> from diffpy.structure.spacegroups import sg35
>>> sg35.short_name
'Cmm2'
>>> Rotation.from_matrix([i.R for i in sg35.symop_list])
Rotation (8,)
[[1.     0.     0.     0.    ]
 [0.     0.     0.     1.    ]
 [0.5774 0.5774 0.     0.5774]
 [0.5774 0.     0.5774 0.5774]
 [1.     0.     0.     0.    ]
 [0.     0.     0.     1.    ]
 [0.5774 0.5774 0.     0.5774]
 [0.5774 0.     0.5774 0.5774]]
>>> Vector3d([i.t for i in sg100.symop_list])
Vector3d (8,)
[[0.  0.  0. ]
 [0.  0.  0. ]
 [0.  0.  0. ]
 [0.  0.  0. ]
 [0.5 0.5 0. ]
 [0.5 0.5 0. ]
 [0.5 0.5 0. ]
 [0.5 0.5 0. ]]
```

hakonanes:
Closes #99 and #87 (see discussion therein).

Changes in this PR:
* Add `Phase.space_group` property, which stores a `diffpy.structure.spacegroups.SpaceGroup` object or `None`.
* If `Phase.space_group` is not `None`, derive `Phase.point_group` from `space_group`, otherwise return whatever is stored in `Phase._point_group`
* `Phase.space_group` can be set by passing a `SpaceGroup` or an integer 1-230
* `Phase` initialization is possible by either passing:
1. space group (point group is derived)
2. space group and point group (they have to "match" or a warning is raised [not error] and the space group is set to `None`)
3. point group (space group is set to `None`)
4. neither (both set to `None`)
* Add `space_groups` parameter to `PhaseList.__init__()`
* Add `PhaseList.space_groups` property
* `Phase`, `PhaseList`, and `CrystalMap` string representation now show
```python
>>> cm
Phase  Orientations       Name  Space group  Point group  Proper point group       Color
    1  5657 (48.4%)  austenite         None          432                 432    tab:blue
    2  6043 (51.6%)    ferrite         None          432                 432  tab:orange
Properties: iq, dp
Scan unit: um
>>> cm.phases
Id       Name  Space group  Point group  Proper point group       Color
 1  austenite         None          432                 432    tab:blue
 2    ferrite         None          432                 432  tab:orange
>>> cm.phases[1]
<name: austenite. space group: None. point group: 432. proper point group: 432. color: tab:blue>
```
* Update relevant docstrings

Todo in this PR:
- [x] #102 should be merged into this before proceding
- [x] Think of the best way to warn the user when trying to set space group and point group that doesn't "match"
- [x] Update docstring examples
- [x] Update tests

Other PRs:
- [ ] Improve adding a `Phase` to the `PhaseList`, via e.g. `PhaseList.append(Phase)`
- [ ] Write/read space group number to/from the crystal map orix HDF5 file, specifically the `phase2dict()` and `dict2phase()` functions

### DIFF
--- a/orix/crystal_map/crystal_map.py
+++ b/orix/crystal_map/crystal_map.py
@@ -459,9 +459,9 @@ class CrystalMap:
         A CrystalMap object can be indexed in multiple ways...
 
         >>> cm
-        Phase   Orientations       Name  Space/point group       Color
-            1   5657 (48.4%)  austenite         Fm-3m/m-3m    tab:blue
-            2   6043 (51.6%)    ferrite         Im-3m/m-3m  tab:orange
+        Phase  Orientations       Name  Space group  Point group  Proper point group       Color
+            1  5657 (48.4%)  austenite         None          432                 432    tab:blue
+            2  6043 (51.6%)    ferrite         None          432                 432  tab:orange
         Properties: iq, dp
         Scan unit: um
         >>> cm.shape
@@ -471,18 +471,18 @@ class CrystalMap:
 
         >>> cm2 = cm[20:40, 50:60]
         >>> cm2
-        Phase  Orientations       Name  Space/point group       Color
-            1   148 (74.0%)  austenite         Fm-3m/m-3m    tab:blue
-            2    52 (26.0%)    ferrite         Im-3m/m-3m  tab:orange
+        Phase  Orientations       Name  Space group  Point group  Proper point group       Color
+            1   148 (74.0%)  austenite         None          432                 432    tab:blue
+            2    52 (26.0%)    ferrite         None          432                 432  tab:orange
         Properties: iq, dp
         Scan unit: um
         >>> cm2.shape
         (20, 10)
         >>> cm2 = cm[20:40, 3]
         >>> cm2
-        Phase  Orientations       Name  Space/point group       Color
-            1    16 (80.0%)  austenite         Fm-3m/m-3m    tab:blue
-            2     4 (20.0%)    ferrite         Im-3m/m-3m  tab:orange
+        Phase  Orientations       Name  Space group  Point group  Proper point group       Color
+            1    16 (80.0%)  austenite         None          432                 432    tab:blue
+            2     4 (20.0%)    ferrite         None          432                 432  tab:orange
         Properties: iq, dp
         Scan unit: um
         >>> cm2.shape
@@ -492,8 +492,8 @@ class CrystalMap:
 
         >>> cm2 = cm[10, 10]
         >>> cm2
-        Phase  Orientations     Name  Space/point group       Color
-            2    1 (100.0%)  ferrite         Im-3m/m-3m  tab:orange
+        Phase  Orientations     Name  Space group  Point group  Proper point group       Color
+            2    1 (100.0%)  ferrite         None          432                 432  tab:orange
         Properties: iq, dp
         Scan unit: um
         >>> cm.shape
@@ -502,26 +502,26 @@ class CrystalMap:
         ... by phase name(s)
 
         >>> cm2 = cm["austenite"]
-        Phase   Orientations       Name  Space/point group     Color
-            1  5657 (100.0%)  austenite         Fm-3m/m-3m  tab:blue
+        Phase  Orientations       Name  Space group  Point group  Proper point group     Color
+            1  5657 (100.0%)  austenite         None          432                 432  tab:blue
         Properties: iq, dp
         Scan unit: um
         >>> cm2.shape
         (100, 117)
         >>> cm["austenite", "ferrite"]
-        Phase  Orientations       Name  Space/point group       Color
-            1  5657 (48.4%)  austenite         Fm-3m/m-3m    tab:blue
-            2  6043 (51.6%)    ferrite         Im-3m/m-3m  tab:orange
-        Properties: dp, iq
+        Phase  Orientations       Name  Space group  Point group  Proper point group       Color
+            1  5657 (48.4%)  austenite         None          432                 432    tab:blue
+            2  6043 (51.6%)    ferrite         None          432                 432  tab:orange
+        Properties: iq, dp
         Scan unit: um
 
         ... by "indexed" and "not_indexed"
 
         >>> cm["indexed"]
-        Phase  Orientations       Name  Space/point group       Color
-            1  5657 (48.4%)  austenite         Fm-3m/m-3m    tab:blue
-            2  6043 (51.6%)    ferrite         Im-3m/m-3m  tab:orange
-        Properties: dp, iq
+        Phase  Orientations       Name  Space group  Point group  Proper point group       Color
+            1  5657 (48.4%)  austenite         None          432                 432    tab:blue
+            2  6043 (51.6%)    ferrite         None          432                 432  tab:orange
+        Properties: iq, dp
         Scan unit: um
         >>> cm["not_indexed"]
         No data.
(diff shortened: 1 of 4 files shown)


hakonanes:
> This looks like a good thing overall. But I think we should still be explicitly giving the proper point group of the phase in the tables and things because it's the proper point group that is considered when you're considering symmetry equivalent (mis)orientations. Is there any reason not to?

Good point, no, no reason... It comes down to how much info we want to show in the `__repr__()` methods. Should we show the proper point group *instead* of the point group, or all three?

dnjohnstone:
> Good point, no, no reason... It comes down to how much info we want to show in the `__repr__()` methods. Should we show the proper point group _instead_ of the point group, or all three?

My inclination is to go for all 3 - but I'm not too bothered if you prefer only the proper point group.

hakonanes:
> My inclination is to go for all 3 - but I'm not too bothered if you prefer only the proper point group. I think showing all 3 has the advantage of making someone who doesn't immediately know why all 3 are there think about it a little. It's also clear that we're not saying the proper point group is somehow equivalent to the point group, which feels like the kind of confusion that could be made.

Agree. How does this look?

```python
<name: austenite. space/point/proper point group: Fm-3m/m-3m/432. color: tab:blue>
```

That's 82 characters.

dnjohnstone:
I think maybe 3 separate entries?

hakonanes:
```python
>>> Phase(name="austenite", space_group=225)
<name: austenite. space group: Fm-3m. point group: m-3m. proper point group: 432. color: tab:blue>
```

98 characters.

dnjohnstone:
We could call it the **rotation group** instead to make things shorter too. I think that's common terminology and maybe actually clearer here. E.g. https://en.wikipedia.org/wiki/Point_groups_in_three_dimensions

dnjohnstone:
In fact - if we use **rotation group** - I'd be ok with having just the space group and the rotation group. It's not so difficult to find the point group from the space group really and there's less potential confusion with that terminology?

hakonanes:
> In fact - if we use **rotation group** - I'd be ok with having just the space group and the rotation group. It's not so difficult to find the point group from the space group really and there's less potential confusion with that terminology?

I haven't seen rotation group used as a term in crystallography text books, is it common in use in crystallography?

If we don't include the point group, only the proper point group, in the repr methods, I have two thoughts:
* we should make `_get_point_group()` part of the public API
* we should reconsider the `Phase.point_group` and `PhaseList.point_groups` properties

dnjohnstone:
I think SO(3) is pretty commonly called the rotation group in quite many areas and its meaning is not ambiguous. I've certainly seen it in molecular symmetry / chemical crystallography books. Here's one example of a textbook menting the rotation group in a crystallography context https://link.springer.com/content/pdf/bbm%3A978-1-4612-3226-1%2F1.pdf

On the topic of crystallographic conventions, I noticed that you changed **symmetry operation** to **symmetry element** in this PR. I think that this is incorrect so please revert that to symmetry operation. See https://dictionary.iucr.org/Symmetry_operation and https://dictionary.iucr.org/Symmetry_element for definitions.

Unfortunately, many textbooks that cover crystallography do not follow IUCr definitions though.

hakonanes:
> I think SO(3) is pretty commonly called the rotation group in quite many areas and its meaning is not ambiguous. I've certainly seen it in molecular symmetry / chemical crystallography books. Here's one example of a textbook menting the rotation group in a crystallography context https://link.springer.com/content/pdf/bbm%3A978-1-4612-3226-1%2F1.pdf

You've convinced me, should be OK using rotation group instead of proper point group. That would of course require some renaming, considering properties like `orix.quaternion.symmetry.Symmetry.proper_subgroup` etc. Although it's long, I think we should stick to

```python
<name: austenite. space group: Fm-3m. point group: m-3m. proper point group: 432. color: tab:blue>
```

Would be good to hear from @pc494 on using rotation group.

> On the topic of crystallographic conventions, I noticed that you changed **symmetry operation** to **symmetry element** in this PR. I think that this is incorrect so please revert that to symmetry operation. See https://dictionary.iucr.org/Symmetry_operation and https://dictionary.iucr.org/Symmetry_element for definitions.

Ah, thanks a lot, I'll fix this.

pc494:
I was mid-comment when the most recent one appeared, so posting now.

Low hanging fruit first:

- `_get_point_group() ` should go into the public-API, it was private when I wrote it because it was unclear and for internal use only, since it's been improved it is easily good enough to be public.
- operation seems to be correct as @dnjohnstone suggests

I am having a think about "rotation group".

hakonanes:
>     * `_get_point_group() ` should go into the public-API, it was private when I wrote it because it was unclear and for internal use only, since it's been improved it is easily good enough to be public.

Great, I'll make a separate PR after this is merged.

dnjohnstone:
Is this just waiting on @pc494 to give an opinion on rotation group?

> You've convinced me, should be OK using rotation group instead of proper point group. That would of course require some renaming, considering properties like `orix.quaternion.symmetry.Symmetry.proper_subgroup` etc. Although it's long, I think we should stick to
> 
> ```python
> <name: austenite. space group: Fm-3m. point group: m-3m. proper point group: 432. color: tab:blue>
> ```
> 
> Would be good to hear from @pc494 on using rotation group.

Given the points you raise here - shall we just go for proper point group and the format you give here, then deal with rotation group renaming in a separate PR where we can deal with the properties that would need updating if/when we decide on that?

pc494:
I'm happy with this, will review after this meeting is over

hakonanes:
> > ```python
> > <name: austenite. space group: Fm-3m. point group: m-3m. proper point group: 432. color: tab:blue>
> > ```
> Given the points you raise here - shall we just go for proper point group and the format you give here, then deal with rotation group renaming in a separate PR where we can deal with the properties that would need updating if/when we decide on that?

Sounds good, I'll push the changes to get this.

hakonanes:
See the new string representations in the top comment https://github.com/pyxem/orix/pull/104#issue-468826132. Although it is complete, it is arguably not pretty, and verbose.

hakonanes:
If you're good with this, I'm ready for this to be merged.

I'll immediately get to the two points in the "Other PRs" in the top comment.

pc494 (approved):
LGTM, merging

Issue #88, opened 2020-07-23
Continuous integration and releases


dnjohnstone:
Two questions:

1. Would we prefer CircleCI + CodeCov to our current Travis + Coveralls - or do we not see much difference?

2. Do we think it's worth automating releases to avoid manual errors, make life easier, and/or enable nightly builds?

hakonanes:
1. I haven't used CircleCI/CodeCov, so I couldn't tell if the benefits are great enough to switch.
2. Two parts:
* Yes to automating releases. What do you mean by automating? With kikuchipy, we tell Travis to deploy to PyPI after it runs with a `deploy` flag set to `True` (https://github.com/kikuchipy/kikuchipy/blob/master/.travis.yml#L81). The `deploy` flag is only set to `True` when a new release tag is made from GitHub. This setup has worked perfecly so far. The conda-forge bot will pick up on the new release on PyPI, and automatically create a PR which we can just merge (after some maintenance changes are merged into that PR), and thus the PyPI release is also available via conda-forge.
* I know nothing about nightly builds: What are their use cases? For people who wants to build orix themselves with the latest changes from `master`? Will these be available from PyPI/conda-forge? If not, is there any difference from installing from GitHub? Do we know of any similar projects doing this?

pc494:
1) Not that interested in changing how we do CI

2) Yes, I had a brief look at the github workflow that would build when we mark a release on github. Although the `kikuchipy` also sounds good.

3) If you mean nightly builds for CI purposes, yes we can do that, `pyxem` currently has one (for example)

hakonanes:
I'm for nightly builds for CI purposes.

hakonanes:
I should note that I nicked the Travis-deploy-on-release-tag from HyperSpy, although it seems that they've switched to deploying from Azure Pipelines (https://github.com/hyperspy/hyperspy/blob/RELEASE_next_minor/azure-pipelines.yml#L74).

pc494:
I will do these for `0.4.0`, also; will delete the stray `0.2.3dev` from pypi?

pc494:
- [ ] Nightly travis build
- [ ] Release triggered pypi uploads

pc494:
After some tinkering this is now up and working - hitting the release button on github will upload a version to PyPi. Be aware, the version (for PyPi) is taken from the code and not from github,

dnjohnstone:
[...]

PR #105, opened 2020-08-18
Automating package publishing with github actions

pc494:
Addressing issue #88 this PR will build a pypi release when a release is marked in github.


pc494:
This should now work, I will make a `0.4.0.x` release (on github) to test that this works as expected, I won't update any of the code to do this, and I'll yank the pypi release (and not merge on conda) so that it leaves a minimal footprint.

dnjohnstone:
@pc494 - do you want to just merge what you need to for this yourself? Consider this a green light if you do.

PR #114, opened 2020-08-24
Restructure API, move Vector3d and Quaternion classes

### LINKED ISSUE #113: Simplify and standardize package structure and API
I think this is a good time to (1) formalize some guides for the import structure and (2) simplify the API. 

I see there was an attempt to simplify the API via `orix.objects` and `orix.symmetry`, is this still the goal?

I think there are generally 2,5 (2 and 3 go together) options for an import structure:
1. Import everything
```python
>>> import orix
>>> orix.vector.Vector3d.xvector()
Vector3d (1,)
[[1 0 0]]
```
2. Import modules/submodules/sub-packages
```python
>>> from orix import vector
>>> vector.Vector3d.xvector()
Vector3d (1,)
[[1 0 0]]
```
3. Import functions and classes
```python
>>> Vector3d.xvector()
Vector3d (1,)
[[1 0 0]]
```

I think we shouldn't make anything available from `orix` (except `__version__` etc.), and force users to either import modules, or functions and classes from modules.

We should also make the API as intuitive as possible. This is an example of how it is now for the `orix.quaternion` module, which I think is unintuitive

```python
>>> from orix.quaternion.orientation import Orientation
>>> from orix.quaternion.orientation_region import OrientationRegion
>>> from orix.quaternion.symmetry import Symmetry
```

It should be like this

```python
>>> from orix.quaternion import Quaternion, Orientation, OrientationRegion, Symmetry
```

In general I'm a fan of these guiding principles:
* To use a module it must be imported explicitly
* Functions and classes can either be called from an imported module, or directly imported from the module
* A function or class in the public API should only be importable directly from a module

`scikit-image` does it like this: https://github.com/scikit-image/scikit-image/blob/master/skimage/__init__.py. I like their import structure a lot, and would like us to use that.

Practically speaking I think what is required is to
* move class definitions Vector3d, Quaternion, Scalar (and Object3d if that's to be part of the public API) away from __init__.py to their own files
* import what's supposed to be available from module/__init__.py into that file, and perhaps specify a __all__ list to really show what is part of the public API and not, like `scikit-image` does

These changes would create this (public) API

```python
>>> from orix.crystal_map import CrystalMap, CrystalMapProperties, Phase, PhaseList
>>> from orix.io import load, save, loadang, loadctf
>>> from orix.plot import CrystalMapPlot, AxAnglePlot, RodriguesPlot, RotationPlot
>>> from orix.quaternion import Quaternion, Rotation, Misorientation, Orientation, OrientationRegion, Symmetry
>>> from orix.sampling import get_sample_fundamental, get_sample_local, uniform_SO3_sample
>>> from orix.scalar import Scalar
>>> from orix.vector import Vector3d, AxAngle, Homochoric, NeoEuler, Rodrigues, SphericalRegion
```

hakonanes:

This PR was motivated by discussions in https://github.com/pyxem/orix/issues/113.

* Removed `orix/objects.py` and `orix/symmetry.py`, as it is not clear how these should be used
* Moved `Quaternion` class from `orix/quaternion/__init__.py` to `orix/quaternion/quaternion.py`, but made available via `orix/quaternion/__init__.py`
* Moved `Vector3d` class from `orix/vector/__init__.py` to `orix/vector/vector3d.py`, but made available via `orix/vector/__init__.py`
* Imported all public classes and functions in respective module `__init__.py` scripts.
* Added a `__all__` list to every module `__init__.py`. This specifies what will be imported when calling `from orix.quaternion import *`

The changes made leads to the following import structure/public API
```python
>>> from orix.base import check, DimensionError, Object3d
>>> from orix.crystal_map import CrystalMap, Phase, PhaseList, CrystalMapProperties
>>> from orix.io import loadang, loadctf, load, save
>>> from orix.plot import convert_unit, CrystalMapPlot, AxAnglePlot, RodriguesPlot, RotationPlot
>>> from orix.quaternion import check_quaternion, Quaternion, Rotation, von_mises, Misorientation, Orientation, get_proper_groups, OrientationRegion, get_distinguished_points, get_point_group, Symmetry
>>> from orix.sampling import get_sample_fundamental, get_sample_loca, uniform_SO3_sample
>>> from orix.scalar import Scalar
>>> from orix.vector import check_vector, Vector3d, AxAngle, Homochoric, NeoEuler, Rodrigues, SphericalRegion
```

Closes #113.


pc494 (approved):
LGTM, will merge

PR #116, opened 2020-08-26
Add spherical coordinate attributes to Vector3d, plus Vector3d.to_polar()

### LINKED ISSUE #115: Vector3d class should have radius and polar coordinates as attributes
MTEX' `Vector3d` has these properties (https://github.com/mtex-toolbox/mtex/blob/develop/geometry/%40vector3d/vector3d.m#L47), we should too.

I just found myself wanting the spherical coordinates of `Vector3d`, i.e. the radius r = sqrt(x**2 + y**2 + z**2), polar coordinate theta = atan2(y, x) and azimuthal coordinate phi/rho =  arccos(z / r). The computations are simple, so I think these should be attributes of `Vector3d` computed on the fly.

Will make a PR shortly if there is agreement.

hakonanes:

See #115 for brief motivation. Closes #115.

* Add derived `Vector3d.theta`, `Vector3d.phi` and `Vector3d.r` attributes.
* Add `Vector3d.to_polar()` method to return `theta`, `phi` and `r`. The method takes no parameters.

Written simple tests, which pass. Coverage remains at 100%.

Todo:
- [x] @pc494, were you thinking of a `polar()` function defined outside of `Vector3d`, so this could be used elsewhere? If so, where do you think this should be placed?
- [x] Decide whether `Vector3d.r` etc. should return `Scalar` objects or numpy arrays. I prefer numpy arrays, since I haven't yet understood the use of `Scalar`.


pc494:
In the short term I prefer `Scalar`, it's architecturally consistent even is Scalar as a class has always seemed a bit shallow in orix. I don't think we need to abstract `polar` just yet, for example if you got axes (from axis-angle rep) they would already be Vector3D. Review coming up.

orix/vector/vector3d.py
```diff
@@ -307,6 +307,20 @@ def xyz(self):
         """tuple of ndarray : This vector's components, useful for plotting."""
         return self.x.data, self.y.data, self.z.data
 
+    @property
+    def r(self):
+        return np.sqrt(self.x.data ** 2 + self.y.data ** 2 + self.z.data ** 2)
+
+    @property
+    def theta(self):
```
pc494, line 315: can we have a one liner describing convention here please?

orix/vector/vector3d.py
```diff
@@ -307,6 +307,20 @@ def xyz(self):
...
+        return np.arccos(self.z.data / self.r)
+
+    @property
+    def phi(self):
```
pc494, line 319: same again, just be clear which angle we are using

orix/vector/vector3d.py
```diff
@@ -409,3 +423,7 @@ def _tuples(self):
     def mean(self):
         axis = tuple(range(self.data_dim))
         return self.__class__(self.data.mean(axis=axis))
+
+    def to_polar(self):
```
pc494, line 427: I know we've not always been consistent, but this should have normal method docstrings, especially as it's returning 3 separate things, and it's not obvious that the order would be what it is.

pc494 (changes requested):
Conventions and docstrings, I think we are using "ISO 31-11" - see https://mathworld.wolfram.com/SphericalCoordinates.html

orix/vector/vector3d.py
hakonanes: Added a bit more than a oneliner... Please say so if you think it is too much.

orix/vector/vector3d.py
hakonanes: Now returns `Scalar`. This lead to e.g. me having to add the `.data` when feeding coordinates back into `Vector3d.to_polar()`, which should be unnecessary:

```python
    def test_polar_loop(self, vector):
        theta, phi, r = vector.to_polar()
        vector2 = Vector3d.from_polar(theta=theta.data, phi=phi.data, r=r.data)
        assert np.allclose(vector.data, vector2.data)
```

Either `Vector3d.from_polar()` should accept `Scalar`, or we should stop using it.

pc494: I actually don't mind this as is, I've got quiet used (from hyperspy) to `.data` whenever I'm expecting numpy arrays and the "test loop case" is an edge case in day to day usage. If you @hakonanes are okay with leaving it as it is, this is good to go

pc494 (approved):

orix/vector/vector3d.py
hakonanes: I'm fine with this now.

I agree that accessing .data is simple and not at all a user barrier. `Scalar` might make some internal tasks/logic simpler, I don't know, it's just that I've yet to encounter a situation where I need `Scalar`.

pc494:
Merging, we scan (and probably should) consider `Scalar`'s place within `orix` at a future point

PR #119, opened 2020-09-06
New Release: v0.5.0

pc494:
If nobody objects I'm going to release the code as is now, so that an upstream bug in diffsims can be addressed. I think there isn't anything outstanding that needs looking at; all that springs to mind is:

(1) @hakonanes had some comments about pinning the website (?!) that I've now forgotten, are these relevant?
(2) A Changelog, which I am going to suggest we defer to v0.6

### DIFF
--- a/orix/__init__.py
+++ b/orix/__init__.py
@@ -1,5 +1,5 @@
 __name__ = "orix"
-__version__ = "0.5.0.dev"
+__version__ = "0.5.0"
 __author__ = "Ben Martineau, Phillip Crout, Håkon Wiik Ånes, Duncan Johnstone"
 __author_email__ = "pyxem.team@gmail.com"
 __description__ = "orix is an open-source python library for handling crystal orientation mapping data."


hakonanes:
> If nobody objects I'm going to release the code as is now, so that an upstream bug in diffsims can be addressed. I think there isn't anything outstanding that needs looking at; all that springs to mind is:

What's the upstream diffsims issue?

> (1) @hakonanes had some comments about pinning the website (?!) that I've now forgotten, are these relevant?

It's just that with a 0.5 tag with Sphinx stuff (doc directory and readthedocs config yaml file) part of the source we can set up a doc build from 0.5 called `stable` in addition to `latest` from the `master` branch source code. I'll do this when we have the tag.

> (2) A Changelog, which I am going to suggest we defer to v0.6

What do people think of "keep a changelog" (https://keepachangelog.com/en/1.1.0/), which we use for kikuchipy (https://kikuchipy.org/en/latest/changelog.html)? I could set this up after 0.5 if desired.

pc494:
So, the diffsims function that calls `get_sample_local` will give nonsense if the user has orix 0.4 installed as the function it calls is broken in that release.

(1) great, that sounds perfect
(2) and yes, I think from 0.5 onwards we should move over to a that way of working. It have no doubt it will save us and users time in the long run

dnjohnstone:
Go for it

PR #121, opened 2020-09-14
Add a changelog

pc494:
Starting the next cycle off with a development change, we're getting a changelog! I've basically copied what we need over from keep-a-changelog. 

For discussion
-------------------

At present I'm happy using the suggested names (Added, Changed, Deprecated, Removed, Fixed and Security) and not specifying exactly which module is effected. If anyone has objections now would be the time to shout. I anticipate that the CHANGELOG will need to be "prettified" a bit for the RELEASES section of github, but that shouldn't be a huge issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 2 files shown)


hakonanes:
Yay! I'm all for how you describe its use, as this is how we do it with kikuchipy.

PR #130, opened 2020-10-26
Ensure correct phase ID from ANG reader, update point group aliases

hakonanes:

* Ensure phase IDs are read from ANG file and phases in the header are returned with correct ID
* Add some more point group aliases found in an ANG file, move this dictionary to the `orix.quaternion.symmetry` module
* Separate IO tests into their own files (general IO and ANG, EMsoft and orix plugins)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,5 +9,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - This project now keeps a Changelog
 
 ### Fixed
-- Phase IDs found in the ANG file header are used when returning av CrystalMap from this
-  file format.
+- ANG file reader now recognises phase IDs defined in the header
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,3 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - This project now keeps a Changelog
+
+### Fixed
+- Phase IDs found in the ANG file header are used when returning av CrystalMap from this
+  file format.
--- a/orix/quaternion/symmetry.py
+++ b/orix/quaternion/symmetry.py
@@ -496,7 +496,9 @@ def get_point_group(space_group_number, proper=False):
         return spacegroup2pointgroup_dict[pgn]["improper"]
 
 
-# Point group alias mapping
+# Point group alias mapping. This is needed because in EDAX TSL OIM
+# Analysis 7.2, e.g. point group 432 is entered as 43.
+# Used when reading a phase's point group from an EDAX ANG file header
 POINT_GROUP_ALIASES = {
     "432": ["43",],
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -7,3 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - This project now keeps a Changelog
+
+### Fixed
+- Phase IDs found in the ANG file header are used when returning av CrystalMap from this
```
pc494, line 12: Can you make this a bit shorter? "ANG file reader now recognises Phase IDs defined in the header" perhaps?

orix/quaternion/symmetry.py
```diff
@@ -494,3 +494,11 @@ def get_point_group(space_group_number, proper=False):
         return spacegroup2pointgroup_dict[pgn]["proper"]
     else:
         return spacegroup2pointgroup_dict[pgn]["improper"]
+
+
+# Point group alias mapping
+POINT_GROUP_ALIASES = {
```
pc494, line 500: is there a reason this is all caps? And could we make the comment a bit clearer?

pc494 (commented):
Couple of small questions, but broadly looks good to me.

orix/tests/io/test_io.py
```diff
@@ -0,0 +1,158 @@
...
+        assert os.path.isfile(temp_file_path) is True
+
+        crystal_map.phases[0].name = "hepp"
+        save(temp_file_path, crystal_map, overwrite=overwrite)
+
+        crystal_map2 = load(temp_file_path)
+        assert crystal_map2.phases[0].name == expected_phase_name
+
+
+def test_loadctf():
```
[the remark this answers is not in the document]
hakonanes: Yep, one you wrote I think?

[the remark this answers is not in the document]
pc494: As it happens it was Ben, it's fairly crude, one to consider deprecating/updating perhaps.... Not action needed for the time being though.

orix/quaternion/symmetry.py
hakonanes: Well, I considered this dictionary a "constant" (https://www.python.org/dev/peps/pep-0008/#constants) at the module level... But I see that other dictionaries in that file are lowercase, will change.

CHANGELOG.md
hakonanes: Done.

hakonanes:
By the way, note that I've asked Readthedocs to build our docs (API reference) on every PR. This enables us to check how any changes to the docs will look after merging.

hakonanes:
I wonder if Travis would build quicker if we used .com instead of .org?

pc494:
The .com migration does need to happen soon. Don't think that's what is going on here though, although hard to be sure.

pc494:
Okay, dealing with the .org .com elsewhere, merging this.

PR #133, opened 2020-10-28
Convert unrefined Euler angles from degrees to radians in EMsoft IO plugin

hakonanes:

Fixes major bug in the EMsoft IO plugin where unrefined Euler angles read from the array of all Euler angles used for simulating patterns, in degrees, were passed to `quaternion.Rotation.from_euler()`... They are now converted to radians:

* Rotations from DictionaryEulerAngles HDF5 data set are converted to radians correctly in the EMsoft h5ebsd plugin.
* Artificial EMsoft h5ebsd file (ment to reflect what's returned by EMsofts EMEBSDDI program) now writes dictionary Euler angles in degrees (not radians, as before) to DictionaryEulerAngles HDF5 data set in `conftest.py`, in the angle range [150, 160].
* A test is added to ensure that the dictionary Euler angles read from the artificial file is in the range [150, 160].

Did not discover this previously because I always performed orientation refinement with EMsoft and read those rotations from file instead of the initial best matching rotations (`orix.io.plugin.emsoft_h5ebsd.file_reader(filename, refined=True)`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,4 @@ ENV/
 
 # Visual Studio Code
 *.code-workspace
+.vscode
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,4 +9,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - This project now keeps a Changelog
 
 ### Fixed
+- EMsoft file reader reads Euler angles correctly
 - ANG file reader now recognises phase IDs defined in the header
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,5 +9,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - This project now keeps a Changelog
 
 ### Fixed
-- EMsoft file reader reads Euler angles correctly
+- EMsoft file reader reads unrefined Euler angles correctly
 - ANG file reader now recognises phase IDs defined in the header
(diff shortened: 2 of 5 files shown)


pc494:
@hakonanes this looks fine, could you spell out in a little more detail the two workflows. Why did saving and loading from a file get around the bug in the code as it currently is?

hakonanes:
An indexing run with EMsoft consists of two steps:
1. Indexing with a finite dictionary (typically ~333000 simulated patterns for point group m-3m) with EMEBSDDI.f90, all results written to an HDF5 file. Best matching rotations are read from this list of ~333000 rotations in the file (stored in degrees). (These were not converted to radians in the orix EMsoft plugin.)
2. n best matching rotations per point are refined with EMFitOrientation.f90 (bound optimization by quadratic approximation, BOBYQA), and the refined Euler angles are written to the same HDF5 file with the unrefined results (stored in radians).

Does that answer your question?

The rotations aren't tested in the EMsoft h5ebsd IO plugin tests, which read from an artificially created EMsoft h5ebsd file generated in `conftest.py`... Which I now see writes radians to the `DictionaryEulerAngles` data set, not corresponding to the degrees in the same data set in real EMsoft h5ebsd files.

Will fix this and update the tests!

pc494:
I think the comment answered most of my questions, it seemed strange that you got different units depending on where you exported from EMSOFT, but that's okay. Will let you add that fix and update and then it'll be good to review

hakonanes:
> I think the comment answered my questions, it seemed strange that you got different units depending on where you exported from EMSOFT, but that's okay.

Don't know the why they use both degrees and radians, but that is the case.

hakonanes:
Updated the top comment with a bullet point list explaining the final changes in this PR.

pc494 (approved):

PR #135, opened 2020-10-29
Adds a documentation page into the rtfd

pc494:
With a patch release coming up it seemed a good time to slowly start writing docs. This PR is more about getting myself used to the format of rtfd. It will eventually tell users about the nice overwriding of methods we have done, to show people that operations like:

`q1 * v * (~q1)`

exist and work nicely.


pc494:
Moment of appreciation for @hakonanes, building the docs on pull requests looks like a godsend for debugging documentation! Thank you!

hakonanes:
We're moving over to writing the user guide in Jupyter notebooks in kikuchipy, replacing the reStructuredText (reST) files. So far only one user guide page, pattern processing, is built from a notebook: https://kikuchipy.org/en/latest/pattern_processing.html

This enables:
1. Checking that notebooks run all the way through upon PR commits
2. Quick feedback in the notebook when writing docs
3. Cell output cheks with nbval (not set up for kikuchipy yet)
4. Potential to set up a Binder from these notebooks, enabling the user to try out orix in their browser

I suggest we use notebooks instead of reST files to build an orix user guide, what do you think @pc494?

This means introducing a dependency on `npsphinx`.

pc494:
Thanks for the reminder @hakonanes, I've been following this conversation. 

I think for now (and here) I'm going to go with the low friction option. In my mind I had been toying with the idea of a "mixed" system. Given that such a port would probably involve getting rid (or at least deprecating) orix-demos it's probably worth waiting till (at least) a new minor release, if not the `1.0.0`. Also you can consider this my commitment to converting the content of this PR into whatever our final form documentation is when the time comes.

hakonanes:
> I think for now (and here) I'm going to go with the low friction option.

Sounds good.

> a "mixed" system.

Agreed, if you mean stuff like API, readme, contributing guide, installation etc. is reST, user guide is notebooks (what we plan for kikuchipy).

> Given that such a port would probably involve getting rid (or at least deprecating) orix-demos

The issue there, which we haven't solved for kikuchipy yet, is the larger test datasets... I hope to set up a system where small datasets (single EBSD patterns/EBSD master patterns) are included in a `kikuchipy.data` module, and larger datasets are downloaded by CI/the user from Zenodo or similar for some notebooks/user guides. We'll see how that goes.

> Also you can consider this my commitment to converting the content of this PR into whatever our final form documentation is when the time comes.

Great!

pc494:
> > a "mixed" system.
> 
> Agreed, if you mean stuff like API, readme, contributing guide, installation etc. is reST, user guide is notebooks (what we plan for kikuchipy).

Yes, I think there is a cross over point where the amount of "text" to "code" means you probably want to use reST. I find the hyperspy user guide pretty readable for example. I'm slightly hoping that our trials of different things over the coming months will make an "easy" answer drop out. Maybe that's optimistic...

===
Unrelated, but reviewing #133 I realised that `from_euler` doesn't state in docstrings that it say it's radians, so I'll add that here so it goes in for `v0.5.1`

pc494:
@hakonanes can you see something obvious I am doing wrong here? I had expected the toctree to spit the two .rst files onto the sidebar, but it doesn't seem to have rendered like that.

pc494:
TODO:

- [ ] inline code
- [ ] Orientation example
- [ ] doctest
- [ ] proofread

pc494:
Closing this as the documentation style has moved on since then.

PR #144, opened 2020-12-06
.ang file format writer for CrystalMap

### LINKED ISSUE #127: Export CrystalMap to the ANG format
So far we only have the orix HDF5 writer to export a CrystalMap... Which no software other than, you guessed it, orix!, has a reader for. So, I'll add a writer to the ANG format plugin (which we have a reader for), supporting at least the format we have in our lab, which, most importantly, MTEX can read.

hakonanes:
* Enable writing a `CrystalMap` to the .ang file format, readable at least by MTEX.
* Improve `CrystalMap.get_map_data()`, so that boolean arrays in returns a boolean array. Also converts array to float if `np.nan` is to be used as the filling value for points not in the data. Plus other minor improvements.
* Add orix to the list of .ang formats in the .ang reader

Close #127.

Todo:
- [x] Write (a working) writer
- [x] If `CrystalMap.rotations_per_point` > 1, should we be able to specify which rotations (best, second best, or third best, etc.) to write to the .ang file? Currently, just the top rotations can be written.
- [x] If we could write other values than image quality, confidence index, SEM signal, and pattern fit to file, would it be used?
- [x] ANG file can be read by (and treats points not in the scan, and non-indexed points [these are distinctly different classes of points]) correctly
    - [x] MTEX
    - [x] EDAX TSL OIM Analysis 7 (phase ID flip issue... see if flipping phase ordering in header fixes this)
- [x] Ensure that a 1D map can be written correctly and read by MTEX and TSL.
- [x] Only write phases in the data set (i.e. `CrystalMap.phases_in_data`) to the file header.
- [x] Test writer
- [x] Add writer to API reference
- [x] Update changelog

After this, the possibility should be added, in a PR to be merged after this feature is released, to the CrystalMap notebook in orix-demos.

**Edit:**

This

```python
>>> import os
>>> from orix import io
>>> data = "/home/hakon/phd/data/jarle_emsoft/sdss/em/sdss_ferrite_austenite.ang"
>>> xmap = io.load(data)
>>> xmap.prop["iq_times_dp"] = xmap.iq * xmap.dp
>>> xmap
Phase  Orientations                 Name  Space group  Point group  Proper point group       Color
    1  5657 (48.4%)  austenite/austenite         None          432                 432    tab:blue
    2  6043 (51.6%)      ferrite/ferrite         None          432                 432  tab:orange
Properties: iq, dp, iq_times_dp
Scan unit: um
>>> datadir, _ = os.path.split(data)
>>> io.save(
...     os.path.join(datadir, "test.ang"),
...     xmap,
...     confidence_index_prop="dp",
...     extra_prop="iq_times_dp",
... )
```

gives a file `test.ang` which looks like this (top lines):

```
# TEM_PIXperUM           1.000000
# x-star                 0.000000
# y-star                 0.000000
# z-star                 0.000000
# WorkingDistance        0.000000
# 
# Phase 2
# MaterialName    austenite/austenite
# Formula    austenite/austenite
# Info
# Symmetry    43
# LatticeConstants    3.595 3.595 3.595 90.000 90.000 90.000
# NumberFamilies    0
# Phase 1
# MaterialName    ferrite/ferrite
# Formula    ferrite/ferrite
# Info
# Symmetry    43
# LatticeConstants    2.867 2.867 2.867 90.000 90.000 90.000
# NumberFamilies    0
# GRID: SqrGrid
# XSTEP: 1.500000
# YSTEP: 1.500000
# NCOLS_ODD: 117
# NCOLS_EVEN: 117
# NROWS: 100
# 
# OPERATOR: orix_v0.5.1
# 
# SAMPLEID:
# 
# SCANID:
# 
# Column names: phi1, Phi, phi2, x, y, confidence_index, image_quality, phase_id, pattern_fit, detector_signal, iq_times dp
# 
-2.73531   0.67696   2.98719     0.00000     0.00000   24.40000   0.79900   1   0.00000   0.00000   19.49560
 2.71855   0.93469  -2.25186     1.50000     0.00000   24.00000   0.79700   2   0.00000   0.00000   19.12800
 2.72159   0.93613  -2.25384     3.00000     0.00000   30.30000   0.82500   2   0.00000   0.00000   24.99750
 2.71813   0.93655  -2.25410     4.50000     0.00000   32.00000   0.83100   2   0.00000   0.00000   26.59200
```

The `orix.io.plugins.ang.file_writer()` signature:

```python
def file_writer(
    filename,
    xmap,
    index=None,
    image_quality_prop=None,
    confidence_index_prop=None,
    detector_signal_prop=None,
    pattern_fit_prop=None,
    extra_prop=None,
):
    """Write a :class:`~orix.crystal_map.crystal_map.CrystalMap` to an
    .ang file readable (at least) by MTEX and EDAX TSL OIM Analysis v7.

    The columns are phi1, Phi, phi2, x, y, image_quality,
    confidence_index, phase_id, detector_signal, and pattern_fit.

    Parameters in masked out or non-indexed points are set to
        - euler angles = 4 * pi
        - image quality = 0
        - confidence index = -1
        - phase ID = 0 if single phase or -1 if multi phase
        - pattern fit = 180
        - detector signal = 0
        - extra properties = 0

    Parameters
    ----------
    filename : str
        File name with an ".ang" file extension to write to.
    xmap : CrystalMap
        Crystal map to write to file.
    index : int, optional
        If the crystal map has more than one rotation/match and phase
        ID per point, this index can be set to write that "layer" of
        data to the file. For properties to be written as well, these
        must also have multiple values per point. To get the best match
        at every point, use None (default).
    image_quality_prop : str, optional
        Which map property to use as the image quality. If None
        (default), "iq" or "imagequality", if present, is used, or just
        zeros. If the property has more than one value per point, only
        the first value is used.
    confidence_index_prop : str, optional
        Which map property to use as the confidence index. If None
        (default), "ci" or "confidenceindex", if present, is used, or
        just zeros. If the property has more than one value per point,
        only the first value is used.
    detector_signal_prop : str, optional
        Which map property to use as the detector signal. If None
        (default), "ds", or "detectorsignal", if present, is used, or
        just zeros. If the property has more than one value per point,
        only the first value is used.
    pattern_fit_prop : str, optional
        Which map property to use as the pattern fit. If None
        (default), "fit" or "patternfit", if present, is used, or just
        zeros. If the property has more than one value per point, only
        the first value is used.
    extra_prop : str or list of str, optional
        One or multiple properties to add as extra columns in the .ang
        file, as a string or a list of strings. If None (default), no
        extra properties are added.
    """
```

Any suggestions of improvements would be much appreciated.


pc494:
This is an important addition, I approve

hakonanes:
> does the failing tests means that github actions understands what a draft PR is?

Nope, I cancelled them. I've read that GitHub Actions doesn't respect `[skip ci]` in the commit message either, so for now we have to live with cancelling them manually (and the subsequent email saying that the build failed).

hakonanes:
I consider the writer done for now, but would appreciate a quick review on the API, specifically parameters and docstring. I'll write tests and do the rest after we've decided on that.

pc494:
I ended up here anyway, so I'll add my thoughts. The API you have now looked fine to me.

> If `CrystalMap.rotations_per_point` > 1, should we be able to specify which rotations (best, second best, or third best, etc.) to write to the ANG file? Currently, just the top rotations can be written.

While it's obvious why we might want to do this, I would prefer to build those kind of workflows into orix in the medium term, and save dev time by not implementing them now.

> If we could write other values than image quality, confidence index, SEM signal, and pattern fit to file, would it be used?

As above, although even more strongly think that that kind of workflow can be conducted in orix.

> Ensure that the written ANG files can be read by NanoMegas ASTAR MapViewer

Yup, probably best practice
> Ensure that the written ANG files can be read by EDAX TSL OIM Analysis 7 (what we have in our SEM lab at NTNU)

Again, probably best practice (although rotation conventions!)

>  Ensure that masked out points (`False` in the `CrystalMap.is_in_data` array) are handled as "notindexed" by MTEX and 
"No solution"-points by TSL

This seems likely to be the biggest technical challenge. 

> Only write phases in the data set (i.e. `CrystalMap.phases_in_data`) to the file header.

If it's easier to write all of them I don't think there is a massive reason to do one way or the other. If you conducted an indexing with phase R,G and B, even if you only found R and G you might want MTEX to know about B. What I'm saying is I don't think it's worth sinking too much dev time on this.

hakonanes:
Thanks for the feedback, Phillip. I ended up implementing all the suggestions I made, will see tomorrow if TSL and ASTAR can read the files with the custom property columns and non-indexed/masked out point values (MTEX can).

hakonanes:
I'm satisfied with the writer now, however I'll not remove the WIP status until I've checked the last two boxes in the top comment.

hakonanes:
This writer should now operate as expected in most uses case I could think of. Note that I've tested against MTEX 5.4 and TSL OIM Data Analysis v7.

pc494:
I will review this now :)

orix/crystal_map/crystal_map.py
```diff
@@ -160,9 +160,15 @@ def __init__(
...
-                    del phase_list[i]
+                # Remove superfluous phases:
+                # First, try to remove the phases whose ID is not in the
+                # ID array, in descending list order
+                for i in phase_ids[::-1]:
+                    if n_different == 0:
+                        break
+                    if i not in unique_phase_ids:
+                        del phase_list[i]
+                        n_different -= 1
```
pc494, line 171: Can we swap the order of these if clauses for readability?

orix/io/plugins/ang.py
```diff
@@ -319,3 +344,375 @@ def _get_phases_from_header(header):
...
+        - extra properties = 0
+
+    Parameters
+    ----------
+    filename : str
+        File name with an ".ang" file extension to write to.
+    xmap : CrystalMap
+        Crystal map to write to file.
+    index : int, optional
+        If the crystal map has more than one rotation and phase ID per
```
pc494, line 381: I think we could add as the final sentence. "To get the best match at every pixel, use None (default)"

pc494 (commented):
Only two very minor thoughts, another easy to read PR, many thanks.

orix/crystal_map/crystal_map.py
hakonanes: Agreed, this is done.

orix/io/plugins/ang.py
hakonanes: Agreed! This is done.

pc494:
Great to have this in, an excellent addition.

PR #146, opened 2020-12-11
Add stereo plot

din14970:
A quick and dirty stereographic projection and plot. I use this in my example notebook https://github.com/din14970/indexing_demo. No unit testing and not as nice as one would like, but for me it did the job.


hakonanes:
Thanks @din14970!

At a first glance, this looks very nice. Will have to test it out and see if I get what I expect. Also, we should have a careful think about the API, which in my opinion should be what's simplest for the user in terms of number of lines and default values, but allows most thinkable modifications if desired.

Is there a rush to get this in? Otherwise, I might wait until middle of next week or so before looking at this, considering the diffsims deadline and other orix and kikuchipy related work that's quite pressing.

din14970:
> Is there a rush to get this in? 

No, no rush at all. Agree that it should be fleshed out properly for including it, now it's just quick and dirty because I really needed this type of functionality for my problem.

hakonanes:
I've just pulled this branch and tried it out and tested it against [MTEX](https://mtex-toolbox.github.io/VectorsImport.html). With some minor modifications (I'll get more into it in a review of the code so far early next week), this 

```python
import numpy as np
from orix import plot, vector

theta, phi = np.loadtxt("/home/hakon/vectors.txt", skiprows=1, unpack=True)  # From MTEX' test data
theta = np.deg2rad(theta)
phi = np.deg2rad(phi)
v = vector.Vector3d.from_polar(theta=theta, phi=phi)

fig, ax = plt.subplots(subplot_kw=dict(projection="stereographic"))
ax.scatter(v, s=2)
```

produced this

orix_vectors

which looks like expected, great!

Immediate thoughts:
* We should decide on a default look, e.g. the one above with a specific scatter point size etc.
* We should also find out how we could do the most obvious customizations, like displaying a grid, controlling the grid size etc. And perhaps expose these in some custom methods attached to `StereoPlot`
* Rename `StereoPlot` to `StereographicPlot`?
* Decide whether we want `StereographicProjection` to have a `iproject` instead of the `InverseStereographicProjection` class. MTEX has the former, which I prefer.

din14970:
> We should also find out how we could do the most obvious customizations, like displaying a grid, controlling the grid size etc. And perhaps expose these in some custom methods attached to StereoPlot

I looked briefly into this when I was making this and found examples like [this](https://matplotlib.org/3.2.0/gallery/misc/custom_projection.html#sphx-glr-gallery-misc-custom-projection-py). It got a little bit involved so I didn't bother with it at the time as it seemed like a lot of work for something I didn't really need. An additional issue is that one might like different grids for different applications when making stereoplots e.g.:
- a Wulff grid that is rotatable in the plane depending on which features are of interest
- a projected polar grid to demonstrate how far vectors are tilted from zone axis
- not really a grid, but when plotting single crystal orientations, stereographic projections of the most important traces of the crystal would be nice e.g. [here](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fwww.researchgate.net%2Fprofile%2FMarek_Niewczas%2Fpublication%2F228679019%2Ffigure%2Fdownload%2Ffig8%2FAS%3A637361738244096%401528970288139%2F100-stereographic-projection-showing-the-initial-541-orientation-of-the-tensile-axis.png&f=1&nofb=1)

Fine with your other points

hakonanes:
ok great, it seems like we agree! yes, im interested in the polar angle grid and the traces as well, i think these things should be added at a later stage when we have the basic functionality in this PR merged and tested.

i need this myself, so i would like to contribute anyway i can. are you ok with me pushing commits to your branch, or do you want to make changes, testing and documentation?

din14970:
I haven't looked at this in a while so feel free to directly modify as you see fit. Without really changing much on implementation details, testing and improved documentation + fix according to your suggestions I could manage relatively quickly I think. The grid I would have to study a bit more how it actually works and don't have that time at the moment. What would also be nice is somehow adding symmetry to the equation, e.g. plot crystallographically identical vectors or plot only fundamental zones. This I have absolutely no idea how to do effectively at this point.

hakonanes:
Before reviewing, I suggest that the file with `StereographicTransform` (from @dnjohnstone) is left for another PR, since I don't think it's ready to be used for plotting as it is written now. It might have been without using the `StereographicProjection` class, I don't know. It might be simplest to just set up a new PR, but that can be up to you @din14970. What do you think?

hakonanes:
Apart from the above suggestion, I think it's more or less ready after tests are written. Let's get this in, and then build upon it.

hakonanes:
For your information, I got Ben's code (which Duncan pushed in #62) working nicely, and will make a new PR soon. This will be more flexible and extendable than overwriting `Axes.scatter()` and `Axes.plot()`.

I suggest two options:
1. @din14970 remove's the work in #62 from this PR, and we merge as is without tests, which I will write in another PR building upon the `projections` module started in this PR.
2. I pull the `projections` module in another PR which replaces this PR, with tests and all.

What do you prefer @din14970?

din14970:
@hakonanes sorry for my non-responsiveness, I've been doing a lot of other things recently and haven't gotten around to it. I'll remove the file from #62 . I would leave the other one in for now (stereo_plot) as my examples depend on it. I'll write a couple unit tests and deal with the comments probably tomorrow, then I'll see how you evolve it in your PR and adapt accordingly. Ok for you?

hakonanes:
No problem! Sounds great, I'll review after that.

din14970:
@hakonanes I think it looks ok now, if you feel like things fit in better elsewhere feel free to edit

orix/projections/stereographic_projection.py
```diff
@@ -0,0 +1,175 @@
...
+"""
+
+import numpy as np
+from orix.vector import Vector3d
+
+
+class StereographicProjection:
+    """Project 3D vectors on the x-y plane stereographically"""
+    @classmethod
+    def project(cls, v, pole=-1):
```
hakonanes, line 30: `pole` is not reachable from `StereographicPlot.scatter()`, is that as intended?

By passing `pole=1`, then we would plot the antipodal vector, say on the southern [00-1] hemisphere? So a natural extension of this functionality would be to provide both hemispheres in the same plot if the user passes `antipodal=True` to `scatter()`? If so, it's great that you've included that possibility via the API.

orix/projections/stereographic_projection.py
```diff
@@ -0,0 +1,175 @@
...
+        """Convert vector3d to [X,Y]"""
```
hakonanes, line 31: As per the language on Wikipedia (https://en.wikipedia.org/wiki/Stereographic_projection#First_formulation), this is the inverse stereographic projection, so to be consistent with that formulation we should flip the `project` and `iproject` naming. What do you think?

orix/projections/stereographic_projection.py
hakonanes: I see that you've already included the `project_split()`, nice!

orix/tests/test_projections.py
```diff
@@ -0,0 +1,132 @@
+from orix.projections.stereographic_projection import StereographicProjection
```
hakonanes, line 1: Should add the copyright (with 2018-2021) to this file, to be consistent for other test files.

orix/tests/test_projections.py
```diff
@@ -0,0 +1,132 @@
...
+def test_verify(vector3d):
+    vector3d = vector3d.unit
+    vector3d_up = vector3d[vector3d.z >= 0]
+    vector3d_down = vector3d[vector3d.z < 0]
+    stereocoords_u = StereographicProjection.project(vector3d_up, -1)
+    stereocoords_d = StereographicProjection.project(vector3d_down, 1)
+    inv_u = StereographicProjection.iproject(stereocoords_u, -1)
+    inv_d = StereographicProjection.iproject(stereocoords_d, 1)
+    np.testing.assert_allclose(inv_u.data, vector3d_up.data, rtol=1e-5, atol=1e-5)
+    print(vector3d_down)
```
hakonanes, line 114: These prints statements should be removed.

hakonanes (changes requested):
Looks good, apart from some minor corrections!

I also have some questions in the comments, related to extending your work, which I would appreciate if you took a quick look at.

orix/plot/stereo_plot.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018-2020 the orix developers
```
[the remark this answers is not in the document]
din14970: is there a way to have these things update automatically? In my projects I never bother putting this in there.

orix/projections/stereographic_projection.py
din14970: from the way I read it, the projection is given by (x, y, z) -> (X, Y) in the plane which is what project does, and inverse project is (X, Y) -> (x, y, z). So I think it is correct the way I implemented it.

orix/plot/stereo_plot.py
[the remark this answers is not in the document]
hakonanes: Yeah, I've used a fork of the Python package licenseheaders, as I explained in this PR to diffsims: https://github.com/pyxem/diffsims/pull/151#issuecomment-756029617.

orix/projections/stereographic_projection.py
hakonanes: I guess you're right, let's leave it as you've written it!

orix/projections/stereographic_projection.py
din14970: yes those were the ideas. The main point is that vectors in the same hemisphere as the pole get projected outside the unit circle and this is very impractical in my opinion. So when I project I try to cut off all the vectors that will project outside the unit circle. This is why I want it to default to splitting by poles. The stereographic plot I figure is pretty incomplete anyway and agree with you something like an `antipodal` option makes most sense. Perhaps I can add the option to pass `pole` as well into the plot. I would have to look into it again.

hakonanes: I'm going to update the `StereographicPlot`/Axes class anyways after this PR is in, so I think we can discuss the best way to do this there. Unless you want to work some more on this to enable either plotting the northern or southern hemisphere.

orix/plot/stereo_plot.py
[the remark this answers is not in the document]
hakonanes: OK, I can commit updates to the license headers, and then we can merge.

din14970:
I'll quickly update these things and add support for the pole in the plot. Then Antipodal (showing both hemispheres) is something for later.

hakonanes (approved):
This looks good now, will merge and move on to improve upon what you've established with the `orix.projections` module, @din14970!

pc494:
Great, thanks are due to both of you, this is another nice addition!

Issue #153, opened 2021-02-12
Format of the coordinates and axes of stereographic projections of northern and southern hemispheres


hakonanes:
I'm combining the `StereographicProjection` implemented by @din14970 with the `StereographicPlot` from @bm424 and @dnjohnstone in #62, which is improving every time I understand something new about the transformation framework in Matplotlib, which is rather complex (at least for me).

However, I have a couple of questions concerning the best way to display polar coordinates and the axes in a combined plot with both the upper and lower hemispheres:

This code

```python
import matpotlib.pyplot as plt
from orix import plot, projections, vector
v = vector.Vector3d([
    [0, 0, 1],
    [1, 0, 1],
    [1, 1, 1],
    [0, 0, -1],
    [-1, 0, -1],
    [-1, -1, -1],
])

fig, ax = plt.subplots(ncols=2, subplot_kw=dict(projection="stereographic"))

ax[0].pole = -1
ax[0].scatter(v)  # Lower hemisphere points aren't plotted
ax[0].show_hemisphere()
ax[0].set_xlabel()
ax[0].set_ylabel()
for vi in v[:3]:
    ax[0].text(vi, s=vi._nice_string_repr())

ax[1].pole = 1
ax[1].scatter(v)  # Upper hemisphere points aren't plotted
ax[1].show_hemisphere()
ax[1].set_xlabel()
ax[1].set_ylabel()
for vi in v[3:]:
    ax[1].text(vi, s=vi._nice_string_repr())
```

produced this figure

![stereographic1](https://user-images.githubusercontent.com/12139781/107783525-a77e1e80-6d4a-11eb-9c4c-1651b3bbe2b1.png)

In the upper projection, the azimuthal coordinate goes counterclockwise from 0 to 2 pi, and the polar coordinate goes radially from 0 to pi/2. In the lower projection, the azimuthal coordinate still goes counterclockwise from 0 to 2 pi, but the polar coordinate goes inverse radially from pi / 2 to pi (think of the zenith angle going from northern to southern pole)...

The coordinates in the Matplotlib interactive plotting data bar when hovering over [111] in the upper projection:

![Skjermbilde fra 2021-02-12 15-59-20](https://user-images.githubusercontent.com/12139781/107784641-fbd5ce00-6d4b-11eb-9f12-732ceb74bf1e.png)

and [-1-1-1] in the lower projection:

[...]

PR #154, opened 2021-02-16
Make stereographic plot extend Matplotlib projections, doc improvements

hakonanes:
* Build upon work by @bm424 and @dnjohnstone to extend [Matplotlib's projections module](https://matplotlib.org/stable/api/projections_api.html#module-matplotlib.projections.polar), so that we can plot `orix.vector.Vector3d` in a customizable and extendable `StereographicPlot`.
* Restructure @din14970's stereographic projection class to use `SphericalRegion` to determine which vectors are to be plotted, always returning vectors inside the equatorial plane.
* Add to the stereographic projection user guide notebook showing how to customize the azimuth/polar grid, 
* Improve API documentation

Simplest example use
```python
>>> from orix import plot, vector
>>> fig, ax = plt.subplots(subplot_kw=dict(projections="stereographic"))
>>> ax.scatter(vector.Vector3d([[0, 0, 1], [1, 0, 1]]))
```

Thanks a lot to @pc494 and @din14970 for your feedback in #153! You, and everyone interested, can see more in the updated user guide notebook in the RtD docs built off of this PR in the checks below.

Close #62, #153, #91.

Todo:
* [x] Old tests work
* [x] Test new stuff
* [x] Complete API reference improvements
* [x] Add descriptions to the stereographic projection notebook
* [x] Update changelog


hakonanes:
Here's the user guide notebook built in the first PR checks: https://orix--154.org.readthedocs.build/en/154/stereographic_projection.html

hakonanes:
I just found a way to draw great and small circles on the stereographic projection, which will also enable us to draw a Wullf net, but I'll include this in a second PR.

hakonanes:
This PR is many lines... The main parts are updates to the stereographic projection and plot. Apart from this, there are many lines added to the [stereographic projection notebook](https://orix--154.org.readthedocs.build/en/154/stereographic_projection.html) describing new functionality, and the rest are improvements to the Sphinx generated API reference documentation, including [module and class overview tables](https://orix--154.org.readthedocs.build/en/154/reference.html#api-reference).

@din14970, I would be very greatful to you if you could have a quick look at the notebook above and give any comments you would have to them. I hope you continue to use the stereographic plot and can give feedback so we can develop it further.

@pc494, as always, I'm in your debt.

din14970:
This looks really really awesome! Thanks for this great addition. Also the start of a user guide I think is a really great addition. Some comments or questions:
* is the fact that you allow for 2 different hemisphere labels not a bit confusing? south/lower, north/upper? It's up to you but for me it seems that when you give too many options that do the same thing it might confuse the user. What would make maybe more sense to me if you want to have multiple options is either shortening the option, e.g. `"lower" / "l"` and `"upper" /"u"` or allowing an integer to be passed e.g. `"lower" /  -1` and `"upper" / 1`
* the default grid finesse seems kind of rough, personally I would go with something like 10 degree increments on both angles.
* with the annotate vectors, I think it would be good to parse negative numbers and change it to the  notation. I also think the way to do it is a bit weird, looping over all the points and calling a private method to get the string. 
* I think the fact that there is granularity and full control is really good, with MTEX doing something custom is always quite hacky. On the other hand I think for common workflows some shortcuts would be handy. For example, if you want to plot both hemispheres with the same plot properties (which I guess one would like most of the time), it would be handy to have a shortcut function or argument for that which does all the axes creation/setup behind the scenes. Adding labels to all points could also be a shortcut. What is your opinion on a `plot` method on `Vector3D` with arguments like `projection=stereographic`, `antipodal=True`, `show_grid=True`, `grid_polar_resolution=`, `grid_azimuthal_resolution=`, `color=(string, list of strings, ...)`, `show_labels=True`, `label_format="reciprocal"`?
* Irrelevant to this PR but on a longer timescale, I think what would be really cool would be an interactive plot that can be dragged around with the mouse. So it would be as if you are rotating the sphere and viewing the projections from any angle you like. In the end, the Wulff net is just what happens when you view the polar grid edge-on and project it on the stereographic plane, so if you were to view along X or Y.

hakonanes:
Thanks for the good feedback, @din14970!

> is the fact that you allow for 2 different hemisphere labels not a bit confusing?

I agree, I'll change this to just be "upper" and "lower". I think allowing -1 instead of "upper" and 1 instead of "lower" might confusing to users who don't get the point that -1 projects points from the upper hemisphere, and vice versa.

> the default grid finesse seems kind of rough, personally I would go with something like 10 degree increments on both angles.

Good, I'll change to 10 degrees.

> with the annotate vectors, I think it would be good to parse negative numbers and change it to the  notation. I also think the way to do it is a bit weird, looping over all the points and calling a private method to get the string.

Totally agree, the private method is strictly a temporary measure, until we get the positioning (padding the azimuthal and polar coordinates relative to the vector) and, as you say, $\bar{1}$ instead of -1. The looping is in line with the existing implementation of `Matplotlib.axes.Axes.text()`, in that it only takes in one string and it's coordinates. I have overwritten this native method to allow a Vector3d or (azimuth, polar) as coordinates for the string, but there is nothing in the way for us implementing a `StereographicPlot.annotate_vectors()` custom method. Note that `annotate()` is an existing native method working similarly to `text()`. I would like our annotation method to also allow an array of azimuthal and polar coordinates as well, to be in line with the custom methods.

I suggest we leave the annotation as it is now, and I'll implement `annotate_vectors()` (or another more descriptive name?) in a second PR I'll start on tomorrow or early next week, which will also include plotting great/small circles, Wullf net, curved lines between two vectors (part of a great circle perpendicular to a vector perpendicular to the two vectors), and perhaps more I find useful.

> I think the fact that there is granularity and full control is really good, with MTEX doing something custom is always quite hacky. On the other hand I think for common workflows some shortcuts would be handy.

Yes, the granularity must always be there, but I agree that convenience methods are always nice. MTEX has *really* strong plotting capabilities, and I think you can control most of the plots' appearance, but unfortunately it isn't as well documented as one would like.

> For example, if you want to plot both hemispheres with the same plot properties (which I guess one would like most of the time), it would be handy to have a shortcut function or argument for that which does all the axes creation/setup behind the scenes. 

The reason I haven't implemented a plot method for Vector3d (and CrystalMap) is the simple reason of circular imports, since `orix.plot.stereographic_plot.py` imports `orix.vector.Vector3d`, and we cannot go the other way as well. If we could get around that (which I'm sure is possible, I just don't know how), I'm on board with adding a `Vector3d.plot()` method. I totally agree that we should have a simpler API than the current one, for quick prototyping and for the users who just want the default appearance... I can have a look at adding this .plot() method when the underlying `StereographicPlot` class has settled a bit. If someone else want's to have a look before that, that's fine as well.

> I think what would be really cool would be an interactive plot that can be dragged around with the mouse.

This pan behaviour is part of the Matplotlib.transforms framework, and is supported for their `matplotlib.projections.polar` projection (https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/projections/polar.py#L1437). For now, `StereographicPlot.can_pan` is `False`, but this can be added at a later time.

hakonanes:
@din14970, I find the grid with 10 degrees (right) steps a bit too dense, I prefer 15 degrees (left). Do you think this is sufficient?

stereographic_projection_grids

orix/projections/stereographic.py
```diff
@@ -0,0 +1,286 @@
...
+    @property
+    def inverse(self):
+        """Return the corresponding inverse projection,
+        :class:`InverseStereographicProjection`, with the same
+        projection pole.
+        """
+        return InverseStereographicProjection(pole=self.pole)
+
+
+def _vector2xy(v, pole):
```
pc494, line 196: I think this should have docstrings, even if they only explain why the `np.errstate` block is safe

orix/vector/vector3d.py
```diff
@@ -502,3 +504,14 @@ def to_polar(self):
         theta, phi, r : Scalar
         """
         return self.theta, self.phi, self.r
+
+    def _nice_string_repr(self, parantheses="[]"):
```
pc494, line 508: Two things here, 1) this isn't user facing right? Because it's a bit weird (fine for internal use, but for example, the spaces play weird etc. etc.) and 2) it's "parenthesis" for the spelling.

orix/tests/test_vector3d.py
```diff
@@ -364,3 +331,11 @@ def test_polar_loop(self, vector):
         theta, phi, r = vector.to_polar()
         vector2 = Vector3d.from_polar(theta=theta.data, phi=phi.data, r=r.data)
         assert np.allclose(vector.data, vector2.data)
+
+
+class TestStringRepresentation:
+    def test_nice_string_representation(self):
+        v = Vector3d([[1, 2, 3], [-1, -2, -3]])
+        assert v[0]._nice_string_repr() == "[123]"
+        assert v._nice_string_repr("()") == ["(123)", "(-1-2-3)"]
```
pc494, line 340: I know it's very picky, but I think it's probably worth raising as an issue that we are planning/working on getting the bars on these.

pc494 (changes requested):
This looks really good, and the changes I want are very small

pc494:
> > For example, if you want to plot both hemispheres with the same plot properties (which I guess one would like most of the time), it would be handy to have a shortcut function or argument for that which does all the axes creation/setup behind the scenes.

This is a good feature request, endorsing.

> The reason I haven't implemented a plot method for Vector3d (and CrystalMap) is the simple reason of circular imports, since `orix.plot.stereographic_plot.py` imports `orix.vector.Vector3d`, and we cannot go the other way as well. If we could get around that (which I'm sure is possible, I just don't know how), I'm on board with adding a `Vector3d.plot()` method. I totally agree that we should have a simpler API than the current one, for quick prototyping and for the users who just want the default appearance... I can have a look at adding this .plot() method when the underlying `StereographicPlot` class has settled a bit. If someone else want's to have a look before that, that's fine as well.

I think I've worked through these sorts of circular imports before, I'm happy to prototype. Having a rough `Vector3D` plotter would be a big addition.

>> I think what would be really cool would be an interactive plot that can be dragged around with the mouse.
These are really good to have, but it's worth figuring out exactly which things you want to be able to drag around and why. We currently have interactive plotting for the `Rotations` which I think works pretty well. Right now I'm not yet sold that dragging a 3D sphere make life sufficiently easier to be worth the dev time?

orix/projections/stereographic.py
hakonanes: Agree, added this:

```python
def _vector2xy(v, pole):
    """Return stereographic coordinates (X, Y) of 3D unit vectors.

    (X, Y) is both zero for vectors with z equal to the projection
    pole, and X is zero if x is zero, and similarly for Y/y (hence the
    use of :func:`numpy.errstate`).

    Parameters
    ----------
    v : Vector3d
        If it's not a unit vector, it will be made into one.
    pole : int
        -1 or 1, where -1 (1) means the projection point of the
        stereographic transform is the south (north) pole [00-1]
        ([001]), i.e. only vectors with z > 0 (z < 0) are returned.

    Returns
    -------
    x : numpy.ndarray
        Stereographic coordinate X of shape same shape as the input
        vector shape.
    y : numpy.ndarray
        Stereographic coordinate Y of shape same shape as the input
        vector shape.

    See Also
    --------
    StereographicProjection.vector2xy
    """
    vx, vy, vz = v.unit.xyz
    with np.errstate(invalid="ignore"):
        x = -pole * vx / (vz - pole)
        y = -pole * vy / (vz - pole)
    return x, y
```

orix/vector/vector3d.py
hakonanes: I use this method in the [user guide notebook](https://orix--154.org.readthedocs.build/en/154/stereographic_projection.html#Annotate-vectors), however I didn't plan for us to release orix with this in and it was just a temporary measure before... Niels commented on it above as well. Although it is temporary, it doesn't add anything as it is now, so I will remove it and include something better when creating a `StereographicPlot.annotate_vectors()` method.

orix/tests/test_vector3d.py
hakonanes: Not picky at all, glad you're commeting on this. As mentioned above, will remove this now and reintroduce it when it's better!

orix/projections/stereographic.py
pc494: Sorry, still being picky, divide by zero is scary.
Lifting from:
https://stackoverflow.com/questions/26248654/how-to-return-0-with-divide-by-zero

would this work?
```python
vx, vy, vz = v.unit.xyz
denom = (vz - pole)
# we explicitly say x=0 when denom = 0 (with the zeros_like)
x = np.divide(-pole*vx, denom, out=np.zeros_like(vx), where=denom!=0)
y = np.divide(-pole*vy, denom, out=np.zeros_like(vy), where=denom!=0)
```

and is it harder to read (given that the np.errstate thing should always work because numpy lets you divide by zero)? I'll leave it at your discretion and merge once you're happy

hakonanes:
> These are really good to have, but it's worth figuring out exactly which things you want to be able to drag around and why.

I would much rather write a class extending Matplotlib's 3D plotting for a unit sphere, and plot vectors and circles on top of it, than figuring out how to allow panning. Unless this is easy, I don't know. I can add this thought as a #TODO in the code.

orix/projections/stereographic.py
hakonanes: It's harder to read for sure, but I've included the equation in the [API documentation](https://orix--154.org.readthedocs.build/en/154/reference.html#orix.projections.StereographicProjection.vector2xy), so there should be no doubt how this is calculated. I'll update the calculation to use `where`!

Again, picky is good.

hakonanes: This worked exactly as expected, and it replaced the use of `numpy.errstate`. Thanks a lot! I'll merge after checks pass.

hakonanes:
Thanks a lot to you, @din14970 and @pc494, for reviewing this PR.

@din14970: if you prefer 10 degrees resolution over 15 degrees, I can include that change in a later PR.

din14970:
@hakonanes reason I suggested 10 was because I thought this would allow for easy "counting" of the lines. Having the 15 degree grid makes it quite hard in my opinion to figure out what angle a line actually represents, so effectively the grid becomes a bit of a decoration. If we could have grid labels in a small font or so it would already be a different story and then one could have any type of grid spacing and have it make sense. But indeed, it is already quite a nice step as it is!

hakonanes:
@din14970, you've convinced me, I've changed the default resolution to 10 degrees for the azimuth and polar grids in #158.

PR #156, opened 2021-02-18
Rename Vector3d spherical coordinates to azimuth, polar, radial

hakonanes:
Signed-off-by: Håkon Wiik Ånes <hwaanes@gmail.com>

* As discussed in #153, the `Vector3d` spherical coordinate attributes are renamed from `Vector3d.phi`, `Vector3d.theta` and `Vector3d.r` to `Vector3d.azimuth`, `Vector3d.polar` and `Vector3d.radial`. This is done to reduce potential confusion about how the coordinates are defined. Tests and documentation are updated.
* Fix typo grid resolution in `StereographicPlot.polar_grid()` and `StereographicPlot.azimuth_grid()`.

PR #158, opened 2021-02-18
Draw circles in stereographic projection, quick Vector plotting

### LINKED ISSUE #162: Add plot methods to Vector3d
Turns out we can avoid circular imports simply by doing the following

```python
class Vector3d:
[...]
    figure = None  # Property for easy figure access
[...]
    def plot(self):
        import orix.plot

        fig, ax = plt.subplots(subplot_kw=dict(projection="stereographic"))
        ax.scatter(self)
        self.figure = fig
```

Then can we do this

```python
from orix import vector
v = vector.Vector3d([[0, 0, 1], [1, 0, 1], [1, 1, 1]])
v.scatter()
v.figure.savefig("/home/hakon/vectors.png", bbox_inches="tight")
```

which produces this

vectors

This means that we have to import the `orix.plot` module on *every* call to `.plot()`, but I think that should be OK. What do you think, @pc494 and @din14970?

If we go for this solution, the other issues we have to solve, that I can think of, are
* ~making the Matplotlib figure and axes objects easily available (in private `_figure` and `_axes` attributes?)~ already answered this above, with `Vector3d.figure`.
* allowing customization ~and adding to~ (I didn't get this to work with add_subplot or add_axes) the current figure/axes
* ~easy creation of new figures, also putting multiple side by side~ Except for having a hemisphere="both" keyword argument to `scatter()`, this quickly becomes very custom and opiniontated, so I won't spend any time on it.
* decide on which plotting methods we should make available from `Vector3d`, and what they should be called

If we go forward with this, we should make sure to make nothing available only via the class plotting methods, i.e. also make it available and customizable from the Matplotlib projections. The class plotting methods should only call the projection methods.

hakonanes:

* Add `Vector3d.get_circle(opening_angle=0.5 * np.pi, steps=100)` method to get great or small circles (depending on the opening angle, by default a great circle) about the vectors in the `Vector3d` instance. The output are a set of `steps` vectors delineating the circle per vector in the instance in its own `Vector3d` instance of shape `(n, steps)`, no matter the shape of the `Vector3d` to get circles from.
* Add `StereographicPlot.circle(*args, opening_angle=0.5 * np.pi, **kwargs)` method to draw these circles in the stereographic projection.
* Add better annotation of vectors in stereographic plots.
* Add a `Vector3d.scatter()` convenience plotting method for quick plotting of vectors in the stereographic projection
* Add a `Vector3d.draw_circle()` convenience plotting method for quick plotting of great or small circles in the stereographic projection
* Show how to create a Wulff net in the user guide (see issue #163)

Close #162.

#### Progress of the PR
- ~[ ] Annotation of vectors~ This PR is getting big enough, will leave this for another one
- [x] Complete `Vector3d.scatter()` convenience method
- [x] Wait for #156 to be merged
- [x] Add a `Vector3d.draw_circle()` method
- [x] Figure out best way to create a new figure or build upon ~the existing figure attached to the `Vector3d.figure` attribute~a returned figure

This new code snippet in the stereographic projection user guide notebook

```python
import numpy as np 
from orix import plot, vector

v1 = vector.Vector3d.from_polar(azimuth=np.deg2rad([0, 60, 180]), polar=np.deg2rad([0, 60, 60]))

c = ["C0", "C1", "C2"]
fig = v1.scatter(c=c, s=200, axes_labels=["RD", "TD", None], show_hemisphere_label=True, return_figure=True)
v1.draw_circle(color=c, linewidth=2, figure=fig)
v2 = v1[1].cross(v1[2])
v2.scatter(c="xkcd:pink", marker="p", s=250, figure=fig)
v2.draw_circle(color="xkcd:pink", linestyle="--", linewidth=3, figure=fig)
```

produces this figure

vectors_circles

#### For reviewers


hakonanes:
@din14970, I would be greatful if you could have a quick look at the final notebook updates in the PR built docs. I removed the `Vector3d.figure` attribute, as it was too much work keeping track of it when building upon the current figure... I instead opted for the solution you suggested, by providing a `return_figure` parameter, and a `figure` parameter to pass it to.

din14970:
looks very cool! I think it looks fine. As I wrote in the issue, I'm never totally sure how to best handle these matplotlib objects, so I guess this way has advantages but the old way might also have advantages. I guess if it ends up feeling unnatural it can be changed at a later point.

hakonanes:
Yeah, Matplotlib's highly customizable object oriented functionality means it might be confusing sometimes. But I think the approach of returning a figure if the user want's it, and then passing it on to build upon that figure later, is simple and intuitive, and also way easier to maintain...

orix/plot/stereographic_plot.py
```diff
@@ -403,41 +432,215 @@ def symmetry_marker(self, v, fold, **kwargs):
...
+        for k, v in new_kwargs.items():
+            kwargs.setdefault(k, v)
+
         for vec, marker, marker_size in marker:
             self.scatter(vec, marker=marker, s=marker_size, **kwargs)
 
         # TODO: Find a way to control padding, so that markers aren't
         #  clipped
 
+    def draw_circle(self, *args, opening_angle=0.5 * np.pi, steps=100, **kwargs):
```
pc494, line 456: think `np.pi/2` is easy to read as the default parameter

orix/plot/stereographic_plot.py
```diff
@@ -127,16 +127,22 @@ class StereographicPlot(Axes):
     # See this Matplotlib tutorial for explanations of methods:
     # https://matplotlib.org/stable/gallery/misc/custom_projection.html
 
+    # Elements' zorder (let's try to keep this updated!):
```
pc494, line 130: This could be done via dictionary mapping at a later point (rather than hardcoding the numbers), although that change isn't for this PR

orix/plot/stereographic_plot.py
```diff
@@ -403,41 +432,215 @@ def symmetry_marker(self, v, fold, **kwargs):
...
+            describe unit vectors.
+
+        Returns
+        -------
+        azimuth : numpy.ndarray
+            Azimuth coordiantes of unit vectors.
+        polar : numpy.ndarray
+            Polar coordinates of unit vectors.
+        """
         if len(values) == 2:
```
pc494, line 538: I feel like there is an edge case that will slip past this. The duck typing is fine, but are we sure we can't get a `Vector3D` object with a `len == 2`. Similarly, do we want a np.array of the correction dimensions to throw an error.

Does
```python
try:
    value = value.unit
    azimuth = value.azimuth.data
    polar = value.polar.data
except (AttributeError):
    try:
         azimuth,polar = values[0],values[1]
    except IndexError:
         raise ValueError(...)
```
help this? Leaving at your discretion.

orix/plot/stereographic_plot.py
```diff
@@ -403,41 +432,215 @@ def symmetry_marker(self, v, fold, **kwargs):
...
+    -------
+    numpy.ndarray
+        An array populated with `value` of a size equal to the number of
+        True elements in `visible`.
+    """
+    n = visible.size
+    if not isinstance(value, str) and hasattr(value, "__iter__") and len(value) != n:
+        value = value[0]
+    if isinstance(value, str) or not hasattr(value, "__iter__") or len(value) != n:
+        value = [value,] * n
```
pc494, line 596: Too much logic going on here for my taste. Would pull the len(value)!=n block out at the top to get:

```
if len(value) == n:
      return value
```

Then the `value = value[0] ` block is for non-strings, that I understand. But will anything that doesn't have an __iter__ have a __len__ (needed above)? If not I think you can drop "__iter__" bits of code, but let me see what else we have.

I should probably read the test cases before I write more here.

orix/plot/stereographic_plot.py
```diff
@@ -403,41 +432,215 @@ def symmetry_marker(self, v, fold, **kwargs):
...
+
+    Returns
+    -------
+    azimuth : numpy.ndarray
+    polar : numpy.ndarray
+    """
+    visible = _visible_in_hemisphere(
+        hemisphere=hemisphere, polar_cap=polar_cap, polar=polar
+    )
+    indices = np.where(visible != visible[0])[0]
```
pc494, line 641: This is hard to read, also consider from the docstrings of np.where()

```
where(condition, [x, y])

Return elements chosen from `x` or `y` depending on `condition`.

.. note::
    When only `condition` is provided, this function is a shorthand for
    ``np.asarray(condition).nonzero()``. Using `nonzero` directly should be
    preferred, as it behaves correctly for subclasses. The rest of this
    documentation covers only the case where all three arguments are
    provided.
```

Also, taking the zeroth element from this confused me...

orix/tests/test_stereographic_plot.py
```diff
@@ -309,3 +320,97 @@ def test_transform_inverted_loop(self):
...
+        ],
+    )
+    def test_sort_coords_by_shifted_bools(
+        self, hemisphere, polar_cap, azimuth, polar, desired_azimuth, desired_polar
+    ):
+        azimuth_out, polar_out = plot.stereographic_plot._sort_coords_by_shifted_bools(
+            hemisphere=hemisphere, polar_cap=polar_cap, azimuth=azimuth, polar=polar,
+        )
+        assert np.allclose(azimuth_out, desired_azimuth)
+        assert np.allclose(polar_out, desired_polar)
```
pc494, line 392: Not sure this will catch the behaviour I was talking about earlier. Could you expand on the why we shift the order around for this bit of code?

pc494 (changes requested):
I've not yet reviewed vector3d.py or test_vector3d.py ~ the functionality looks excellent, just got a couple of syntax questions to try and keep the code noise down, as my experience debugging stuff like this at a later suggests clarity is very important. Will hopefully have a finished review by the end of today.

pc494:
(also my bad, I thought I had been pinged for a review here, apparently not...)

orix/plot/stereographic_plot.py
hakonanes: I plan to mainly use that function to create an array of colors and sizes per vectors so that these can be iterated over together with the vectors in e.g. `StereographicPlot.draw_circle()`. Matplotlib singles out colors and sizes as scatter point and line styles that can be iterables, all other arguments are fixed for all scatter points and lines. In many cases, `value` is the default blue Matplotlib color "C0". If number of vectors visible in the hemisphere is 2, this would then return an array with "C0" in it, while I wanted an array with ["C0", "C0"] instead.

I agree that the logic is horrible, but if we don't check whether `value` has the `__iter__` attribute, `len(value)` throws `TypeError`. So they have to be in front of those checks.

Do you prefer try/except? I could use that.

orix/plot/stereographic_plot.py
hakonanes: The input `values` should always be a tuple (as per the docstring), either with two elements (two arrays or floats with azimuth and polar coordinates) or one (a Vector3d object). What's more, `Vector3d` has no `len()`, so if `values` was a `Vector3d` instance that if clause would raise a `TypeError`.

I suggest we leave it as it is, and revisit if we would like for `values` to not strictly be a tuple, but something else (like a `Vector3d` object).

orix/plot/stereographic_plot.py
hakonanes: Very good suggestion. I anticipate we would like to extend this class quite a bit, and it is important to keep track of what goes on top of what. Putting the `zorder`s in a dictionary was quick to do, so I did it in this PR. Better to do it right right away.

orix/plot/stereographic_plot.py
hakonanes: Yup, good suggestion, have changed this in all places in `stereographic_plot.py` and new functionality in `vector3d.py`.

orix/plot/stereographic_plot.py
hakonanes: I use numpy where here to get the first element in the boolean array not equal to the first element in the array (don't know if True or False is the first element). The use case here is as follows: when we draw a great circle in a hemisphere, only half of it will be visible, so half of the elements in `visible` are True. But these values doesn't have to always be at the start of the array or at the end, like [True True True False False False], but can be [True True False False False True]. In the latter case, Matplotlib will upon calls to `plot()` draw a line between all the azimuthal and polar coordinates, and when there is a jump around to the other side of the equatorial plane, a straight line is drawn across the plane as well. We don't want this, therefore this logic.

orix/tests/test_stereographic_plot.py
hakonanes: I agree that it isn't obvious! See my comment above https://github.com/pyxem/orix/pull/158#discussion_r588324465.

hakonanes:
Thanks for the review so far, @pc494! I hope I've answered your questions so that all the if/else clauses and other parts are more clear.

(I'm not sure about your latest comment above, because I did ping you for a review.)

orix/plot/stereographic_plot.py
hakonanes: I'll give `nonzero` a try

hakonanes: Which works like a charm, will use that instead. Thanks!

orix/plot/stereographic_plot.py
pc494: I read `tuple of numpy.ndarray or Vector3d` as (tuple of numpy.ndarray) **or** Vector3d, but given the interpretation you've just given this seems okay

orix/plot/stereographic_plot.py
pc494: On reflection I think this is giving the right answer in a fairly efficient way. I think the final `or len(value)!= n` can be dropped though. Then the four cases run:

single string: False on the first block, True on the second plot, returns ['string'] * n
single float:  False on the first block via the hasattr clause, true on the second one by extension job done
list of length n: False on the final clause for block one and False on the hassattr clause on block two
list of length !=n : block 1 gets the first element, which means block two is dealing with a single float/string (and doesn't use the final bit of or logic)

Is that at least a correct understanding of the code?

orix/plot/stereographic_plot.py
pc494: Marking as resolve, the code is more "numpy consistent" and I now know why we have to reorder the arrays

orix/plot/stereographic_plot.py
```diff
@@ -403,41 +432,215 @@ def symmetry_marker(self, v, fold, **kwargs):
...
+    if indices.size != 0:
```
pc494, line 639: Can `indices.size` be zero?

hakonanes: Yes, for the great circle about the north pole in the when the upper hemisphere is plotted and for the lower equivalent. If no elements are not visible, indices[-1] yields an IndexError.

orix/plot/stereographic_plot.py
hakonanes: Yes, you explained it much better than I did with my example. I removed the last bit of logic in block two, which as you say isn't necessary.

orix/plot/stereographic_plot.py
pc494: cool, resolved

pc494 (approved):

hakonanes:
Thanks a lot for the feedback, @din14970, and for reviewing, @pc494!

PR #164, opened 2021-03-05
Add a Miller class to handle crystal directions and planes

hakonanes:
* Add `orix.vector.Miller` class to handle crystal directions and plane normals, inheriting from `Vector3d`, initialized from a set of hkl, uvw, hkil, UVTW, or xyz, and a `orix.crystal_map.Phase` object.

This moves large parts of the `diffsims.crystallography.ReciprocalLatticePoint` class "closer" to `Vector3d`. Multiple issues with the calculation of symmetrically equivalent planes in other systems than cubic were identified in this diffsims PR https://github.com/pyxem/diffsims/pull/146, which I hope to fix here. We can then update that class in diffsims to inherit from the `Miller` class.

I've added tests with desired results from all seven crystal systems, based upon a "ground truth" produced with [EMsoft's EMstar command line program](https://github.com/EMsoft-org/EMsoft/wiki/EMstar) for those systems. ~None of these pass at the moment (for some systems the equivalent planes are correct, only the order in the arrays is wrong).~ I consider this PR more or less done when these tests pass.

#### Progress of the PR
- [x] Fix calculation of symmetrically equivalent planes
- [x] A user guide notebook explaining a bit about crystal symmetries with the `Miller` class and the stereographic projection
- [x] Determine whether to return [uvw] when doing the cross product of two (hkl) (and vice versa): *yes*
- [x] Complete overwriting of inherited, useful properties and methods
- [x] Ensure operations with rotations maintain Miller's phase property
- [x] ~Add a `Miller.in_fundamental_sector` property to get all vectors inside the fundamental sector~ This will have to wait since the `Symmetry.fundamental_sector()` cannot be used as it is for this (at least that's my experience)

Crystal directions in a tetragonal crystal aren't necessarily parallel to the crystal's planes

```python
>>> from diffpy.structure import Structure, Lattice
>>> from orix.vector import Miller
>>> tetragonal = Phase(point_group="4", structure=Structure(lattice=Lattice(0.5, 0.5, 1, 90, 90, 90)))
>>> m1 = Miller(uvw=[[1, 2, 0], [3, 1, 1]], phase=tetragonal)
>>> m1
Miller (2,), point group 4, uvw
[[1. 2. 0.]
 [3. 1. 1.]]
>>> m2 = Miller(hkl=[[1, 2, 0], [3, 1, 1]], phase=tetragonal)
>>> m2
Miller (2,), point group 4, hkl
[[1. 2. 0.]
 [3. 1. 1.]]
>>> fig = m1.scatter(return_figure=True, c=["b", "r"])
>>> m2.scatter(figure=fig, c=["y", "g"], marker="*")
```

The above code produces this figure

tetragonal

And this code produces this figure

```python
# Default structure, so we don't really need to pass this
>>> cubic = Phase(point_group="m-3m", structure=Structure(lattice=Lattice(1, 1, 1, 90, 90, 90))  
>>> Miller(hkl=[[1, 0, 0], [1, 1, 0], [1, 1, 1]], phase=cubic).symmetrise(unique=True).scatter(hemisphere="both)
```

cubic

See the notebook in this PR for more examples and explanations: https://orix--164.org.readthedocs.build/en/164/crystal_geometry.html

#### For reviewers


hakonanes:
This is still a work in progress, but getting there. I would really appreciate feedback on [this notebook](https://orix--164.org.readthedocs.build/en/164/crystal_geometry.html) from anyone who wants to have a look (@din14970, @pc494, @dnjohnstone).

pc494:
I'm very happy with what's in the notebook so far. I'm also glad to see we're making use of diffpy where appropriate.

hakonanes:
I've updated the TODO list in the top comment, and the notebook, if people wan't to have a look.

hakonanes:
Requested review, but no hurry!

pc494:
Am I right in thinking this was the larger major item before version 0.6?

hakonanes:
Yes. With this in, we can move forward with implementing pole figure plotting, inverse pole figure colouring, a better `ReciprocalLatticeVector` class in diffsims etc. It would be good to have one minor release before this functionality is added, I think, because that will take some time as well...

pc494:
Great, I'll probably be able to review this during the week and we can go from there.

hakonanes:
That would be great!

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+class Miller(Vector3d):
+    """Direct crystal lattice vectors (uvw or UVTW) and reciprocal
+    crystal lattice vectors (hkl or hkil), the latter known as Miller
+    indices, describing directions with respect to the crystal reference
+    frame defined by a phase's crystal lattice and symmetry.
+    """
+
+    def __init__(
+        self,
+        xyz: Optional[Union[np.ndarray, list, tuple]] = None,
```
pc494, line 41: I think we had given up on this kind of documentation for kwargs?

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+        hkl : numpy.ndarray, list, or tuple, optional
+            Indices of reciprocal lattice vector(s). Default is None.
+        hkil : numpy.ndarray, list, or tuple, optional
+            Indices of reciprocal lattice vector(s), often preferred
+            over `hkl` in trigonal and hexagonal lattices. Default is
+            None.
+        phase : orix.crystal_map.Phase, optional
+            A phase with a crystal lattice and symmetry. Must be passed
+            whenever direct or reciprocal lattice vectors are created.
+        coordinate_format : str, optional
```
pc494, line 78: could you explain this?

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+        hkil: Optional[Union[np.ndarray, list, tuple]] = None,
+        phase=None,
+        coordinate_format: str = None,
+    ):
+        """Create a set of direct lattice vectors (uvw or UVTW)
+        or reicprocal lattice vectors (hkl or hkil) describing
+        directions with respect to a crystal reference frame defined by
+        a phase's crystal lattice and symmetry.
+
+        Exactly one of `xyz`, `uvw`, `UVTW`, `hkl`, or `hkil` must be
```
pc494, line 54: code should probably throw an error if more than one is passed

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+            W &= w.
+            \end{align}
+        """
+        if xyz is None and phase is None:
+            raise ValueError(
+                "A phase with a crystal lattice and symmetry must be passed to create "
+                "direct  or reciprocal lattice vector(s)"
+            )
+        self.phase = phase
+
```
pc494, line 101: something like:
```python 
if np.sum([coord is not None for coord in [xyz,uvw,UVTW,hkl,hkli] < 2:
    pass 
else:
    raise ValueError

```

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+            cosines = self.dot_outer(other2).data / (
+                self.norm.data[..., np.newaxis] * other2.norm.data[np.newaxis, ...]
+            )
+            cosines = np.round(cosines, 9)
+            angles = np.min(np.arccos(cosines), axis=-1)
+            return Scalar(angles)
+        else:
+            return super().angle_with(other)
+
+    def cross(self, other):
```
pc494, line 413: is there a type constraint on `other` here?

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+        ----------
+        max_index : int
+            Maximum integer index to round to, by default 20.
+
+        Return
+        ------
+        Miller
+            Rounded set of index triplet(s) or quartet(s).
+        """
+        if self.coordinate_format == "xyz":
```
pc494, line 467: is this not slightly unexpected behaviour?

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+                vi = v2[:, i].unique()
+                l = vi.size
+                v3[i, :l] = vi
+                multiplicity[i] = l
+                idx[l_accum : l_accum + l] = i
+                l_accum += l
+            non_zero = np.sum(np.abs(v3.data), axis=-1) != 0
+            v2 = v3[non_zero]
+            idx = idx[: np.sum(non_zero)]
+
```
pc494, line 529: This is quiet hard to read, a two line statement of the algorithm as a comment at the top of the block would be much appreciated.

pc494 (commented):
Couple of small points to discuss, but the broad structure is excellent, very easy to read

orix/vector/miller.py
hakonanes: Yeah, I was just checking if you had changed your mind... Will drop them!

orix/vector/miller.py
hakonanes: All vector coordinates are converted to cartesian coordinates upon initialization. Whenever we need to return the vectors, we need to know which format they were passed as. Is that what you're asking?

orix/vector/miller.py
hakonanes: Agreed.

orix/vector/miller.py
hakonanes: We call `Vector3d.cross()`, so `Miller.cross()` takes whatever that allows. I think this is OK? If not, do you have a reasonable suggestion?

orix/vector/miller.py
pc494: yes it is. In that case I don't think you want this as a keyword argument, I think you want to just infer it internally.

Otherwise you could do:
`
Miller.(xyz=z,coordinate_format='hkl')`

which would be bad?

orix/vector/miller.py
pc494: `hkl.cross([0,0,1])` is a weird thing to do, but on reflection I think that's on the user so we can leave as is.

orix/vector/miller.py
din14970: if you would do it the single argument way as mentioned above, the other kwargs would make this obvious no?

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+        .. math::
+            U = \\frac{2u - v}{3},
+            V = \\frac{2v - u}{3},
+            T = -\\frac{u + v}{3},
+            W = w.
+        """
+        return _uvw2UVTW(self.uvw)
+
+    @UVTW.setter
```
din14970, line 250: does it make sense to be able to do this calculation for non hex or trig crystals?

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+            vectors. Default is False.
+        return_index : bool, optional
+            Whether to return the indices of the (flattened) data where
+            the unique entries were found. Default is False.
+
+        Returns
+        -------
+        Miller
+            Unique vectors.
+        idx : numpy.ndarray
```
din14970, line 562: maybe could be generalized if the array of vectors is n-d, so full coordinates as in the original array are returned. Not sure how difficult that would be to do.

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+            \end{align}
+        """
+        if xyz is None and phase is None:
+            raise ValueError(
+                "A phase with a crystal lattice and symmetry must be passed to create "
+                "direct  or reciprocal lattice vector(s)"
+            )
+        self.phase = phase
+
+        if xyz is not None:
```
din14970, line 102: I would not allow xyz as input to millers to be honest. It allows for there to be no associated phase and this will give problems with methods where you expect there to be a phase like `symmetrize`. Instead maybe there should be a method to convert Vector3d in sample coordinates to crystal coordinates and vice versa? Phase should be a mandatory argument, miller indices are not defined without.

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+        symmetry = None if self.phase is None else self.phase.point_group.name
+        coordinate_format = self.coordinate_format
+        data = np.array_str(self.coordinates, precision=4, suppress_small=True)
+        return (
+            f"{name} {shape}, point group {symmetry}, {coordinate_format}\n" f"{data}"
+        )
+
+    def __getitem__(self, key):
+        """NumPy fancy indexing of vectors."""
+        return self.__class__(
```
din14970, line 146: Not entirely sure, but will the "view/slice" nature of the data be preserved in this way? Or will it be copied?

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+
+    @property
+    def coordinate_format(self) -> str:
+        """Vector coordinate format, either "xyz", "uvw", "UVTW", "hkl",
+        or "hkil".
+        """
+        return self._coordinate_format
+
+    @coordinate_format.setter
+    def coordinate_format(self, value: str):
```
din14970, line 160: In my view this coordinate format setter is a bit confusing. One also does not need UVTW or hkil unless we deal with hexagonal or trigonal phases. I would be more inclined to somehow store that it is in reciprocal or real space units, and automatically deal with 3-4 index issues depending on the phase.

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+            raise ValueError(f"Available coordinate formats are {formats}")
+        self._coordinate_format = value
+
+    @property
+    def coordinates(self):
+        """Vector coordinates."""
+        coordinate_format = self.coordinate_format
+        if coordinate_format == "xyz":
+            coordinate_format = "data"
+        coordinates = self.__getattribute__(coordinate_format)
```
din14970, line 175: this is a cool trick but in my opinion might be prone to errors if strings or method names end up changing at some point. It may also be confusing for someone trying to understand the code and is dependent on the coordinate_format setter working properly.

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+            Index into `self` for the symmetrically equivalent vectors.
+            Returned if `return_index` is True.
+        """
+        if return_multiplicity and not unique:
+            raise ValueError("`unique` must be True when `return_multiplicity` is True")
+        elif return_index and not unique:
+            raise ValueError("`unique` must be True when `return_index` is True")
+
+        # Symmetrise directions with respect to crystal symmetry
+        operations = self.phase.point_group
```
din14970, line 510: why xyz wthout phase may be dangerous

orix/vector/miller.py
din14970: I think dot and cross may get a bit confusing? What happens when you dot or cross miller vectors representative of different phases, is this really defined? Does this properly deal with all the possible combinations (real, real), (reciprocal, real), (real, reciprocal), (reciprocal, reciprocal)?

din14970 (commented):
Just some thoughts

orix/vector/miller.py
hakonanes: As it is implemented now, we pass `coordinate_format` to `Miller.__init__()` when `Miller` methods and properties like `__getitem__()`, `unit`,  `cross()`, and `__mul__()` are passed the cartesian coordinates "xyz" (which the indexing and computations are performed with), but we want to maintain the `coordinate_format`. We can instead initialize by passing xyz, but then immediately afterwards setting the coordinate format to the correct format.

`Miller.unit`, with "old" implentation commented out:

```python
    @property
    def unit(self):
        """Unit vectors."""
#        return self.__class__(
#            xyz=super().unit.data,
#            phase=self.phase,
#            coordinate_format=self.coordinate_format,
#        )
        m = self.__class__(xyz=super().unit.data, phase=self.phase)
        m.coordinate_format = self.coordinate_format
        return m
```

@pc494, do you prefer this?

orix/vector/miller.py
hakonanes: @pc494, I now do this check, making sure that at exactly one of those parameters are passed, and raising an error if not, thanks.

@din14970, I agree that we could subclass `Miller`, and then loose the `coordinate_format`. However, I don't think we would gain that much, so I don't want to do it. Also, we wouldn't be able to do this (from the [user guide](https://orix--164.org.readthedocs.build/en/164/crystal_geometry.html#Planes-(hkl))):

bilde

orix/vector/miller.py
hakonanes: Done.

orix/vector/miller.py
hakonanes: The method rounds Miller indices, not cartesian coordinates. I think therefore the user should not use it for this purpose. This also replicates [what MTEX does](https://github.com/mtex-toolbox/mtex/blob/develop/geometry/%40Miller/round.m#L5). We could raise a warning instead?

orix/vector/miller.py
pc494: I think this might be worth a future issue.

orix/vector/miller.py
pc494: I was softly nervous about it doubly memory usage, but happy to follow MTEX on this one.

orix/vector/miller.py
pc494: My preference is for the current style, I think that "multiple - but specify only one - kwargs" is a fairly clean pattern.

orix/vector/miller.py
pc494: This makes sense to me now, I would keep it as it is & just highlight in the docstrings that users really shouldn't be doing this.

orix/vector/miller.py
pc494: Given the discussion around line 78 I think we make good use of `xyz` as an input and it's thus worth having.

orix/vector/miller.py
pc494: Probably not, but is it worth writing a check to stop users doing a fairly obviously foolish thing?

orix/vector/miller.py
hakonanes: @din14970, note that the type hints does NOT check input parameter types, they are only used for documentation purposes and by IDEs to give more notifications of possible inconsistencies. Having type hints would not catch e.g. passing a string to xyz in the highlighted code above.

orix/vector/miller.py
hakonanes: Hehe, I removed the coordinate_format parameter now, and actually prefer that and setting it afterwards... So we don't have to worry about your example above, which is now not possible.

orix/vector/miller.py
hakonanes: @din14970, good catch! There are no checks for this. From the [list of Miller methods](https://orix--164.org.readthedocs.build/en/164/reference.html#miller), I think we should make sure the phase and space in `angle_with()`, `cross()`, `dot()`, and `dot_outer()` are the same for `self` and `other`. This means we also need to overwrite `Vector3d.dot()` and `Vector3d.dot_outer()`. Can you see any other methods we should overwrite?

orix/vector/miller.py
hakonanes: Yes, it doubles memory use, but that is as expected, since you do want a new instance returned, and not just a copy.

orix/vector/miller.py
hakonanes: Totally right, I've commented it now.

orix/vector/miller.py
din14970: maybe it can be designed in a way to avoid a check. Never expect users to be reasonable
image

orix/vector/miller.py
din14970: @hakonanes yep, totally aware of this, but I was thinking along the lines of using things like [mypy](http://mypy-lang.org/)

hakonanes: Of course. I haven't used mypy, but we should discuss it in the future issue suggested.

orix/vector/miller.py
hakonanes: Haha, yeah, we can easily add a `Miller.is_hexagonal` by assuring that the lattice parameter angles are 90, 90, 120. Will think about it.

orix/vector/miller.py
hakonanes: A new instance is created. This is as designed.

orix/vector/miller.py
```diff
@@ -0,0 +1,816 @@
...
+        xyz: Optional[Union[np.ndarray, list, tuple]] = None,
+        uvw: Optional[Union[np.ndarray, list, tuple]] = None,
+        UVTW: Optional[Union[np.ndarray, list, tuple]] = None,
+        hkl: Optional[Union[np.ndarray, list, tuple]] = None,
+        hkil: Optional[Union[np.ndarray, list, tuple]] = None,
+        phase=None,
+        coordinate_format: str = None,
+    ):
+        """Create a set of direct lattice vectors (uvw or UVTW)
+        or reicprocal lattice vectors (hkl or hkil) describing
```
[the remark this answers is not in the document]
hakonanes: Great, thanks!

orix/vector/miller.py
hakonanes: The `coordinate_format` is now not a passable parameter in `Miller.__init__()`, but still setable, which is a feature. I think it is fairly robust, and I've written extensive tests, so I'm quite sure changing this in the future is no problem.

orix/vector/miller.py
hakonanes: I've implemented this class in the most flexible way from the start, I like having hkil or UVTW available whenever, not just for hexagonal or trigonal structures, and thus don't want to restrict their use unnecessarily.

orix/vector/miller.py
hakonanes: Until we have overwritten all `Vector3d` methods in `Miller`, there are some places in `Vector3d` where a `self.__class__(data)` is returned. I wanted to allow for this, so made the `phase` parameter optional. I think this is OK for now, but we should revisit this at a later time if there are obvious places where we should be more strict about this.

orix/vector/miller.py
hakonanes: I'm sorry @din14970, you're totally right, we get a view as it is now! We're now returning a deepcopy. I added a test for this. Again, thanks!

orix/vector/miller.py
hakonanes: I've now done these checks, and also overwritten the mentioned four methods. Would really appreciate if anyone could have a look at:
* `Phase.is_hexagonal` (check should live here)
* `Miller.is_hexagonal` (not used anywhere, but nice to have? Might drop.)
* `Miller._compatible_with(self, other, raise_error=False)`
* `Miller.angle_with()`, `Miller.cross()`, `Miller.dot()`, and `Miller.dot_outer()`

hakonanes: Now the methods in the last point only works with other `Miller` instances.

hakonanes:
Thanks for you feedback @din14970! I've answered everything and also implemented your suggested changes where I agree that the change improves the implementation.

orix/vector/miller.py
hakonanes: Ah, these comments were hidden. Yeah, this is the final thing we should look at, generalizing these methods to n-d arrays. I'll have a look at this.

orix/vector/miller.py
hakonanes: https://github.com/pyxem/orix/pull/164#discussion_r616673687

pc494:
If you are happy with the @hakonanes I'll get it merged in tomorrow :)

orix/vector/miller.py
hakonanes: OK, went down the rabbit hole, but I think I'm back up. Have now generalized all methods that can return multidimensional arrays to do so. This included overwriting `Object3d.flatten()` and `.reshape()` (the only ones left are `.empty()` and `.squeeze()`, which we can do at a later time).

`Miller.unique()` and `Miller.symmetrise()` always return instances with fewer and more vectors, respectively. But they should now be able to take `Miller` instances of ND shapes in (tested for shapes `(2, 3, 3)` and `(3, 2, 3)`).

orix/vector/miller.py
hakonanes: Closing this discussion, as I think we've ended up in a better place.

hakonanes:
@pc494, I'm happy with it, merging tomorrow sounds great!

Personally, I think the functionality as implemented is in quite good shape, especially after yours and @din14970's comments. It is also extensively (qualitatively and quantitatively) tested. We'll handle improvements and bugs once we start to build stuff upon it.

pc494:
Great, I've got a use case to road test some of this new code with as well.

hakonanes:
Glad to hear it!

PR #172, opened 2021-04-18
Warn when trying to create rotations from large Euler angles 

### LINKED ISSUE #168: from_euler() could warn if it looks like the input is in degrees
from_euler() could warn if it looks like the input is in degrees, eg:

```python
>>> z = [40,10,5]
>>> q = quaternion.from_euler(z)
```
`
Warning : The method "from_euler" takes angles in radians, it appears you might instead be using degrees
`

I think MTEX does this, for context such a routine would have caught https://github.com/pyxem/pyxem/issues/752 a lot quicker.

@hakonanes thoughts?

hakonanes:
* Address #168. I chose to raise the warning when any of the Euler angle values are above 9, which is [what MTEX does](https://github.com/mtex-toolbox/mtex/blob/afa615d1481174155b434a9d259afd6b1eea732e/geometry/geometry_tools/euler2quat.m#L44).
* Also touched up docstrings in `rotation.py` and in some tests.

Close #168.

#### Progress of the PR

```python
>>> r = Rotation.from_euler([90, 45, 90])
/home/hakon/kode/orix/orix/quaternion/rotation.py:354: UserWarning: Angles are assumed to be in radians, but degrees might have been passed
  warnings.warn(
```

#### For reviewers

### DIFF
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -338,13 +338,13 @@ class Rotation(Quaternion):
         direction : str
             "lab2crystal" or "crystal2lab".
         """
-        conventions = ["bunge", "krakow_hielscher"]
-        if convention.lower() not in conventions:
+        conventions = ["bunge", "Krakow_Hielscher"]
+        if convention not in conventions:
             raise ValueError(
                 f"The chosen convention is not one of the allowed options {conventions}"
             )
         directions = ["lab2crystal", "crystal2lab"]
-        if direction.lower() not in directions:
+        if direction not in directions:
             raise ValueError(
                 f"The chosen direction is not one of the allowed options {directions}"
             )
@@ -358,7 +358,7 @@ class Rotation(Quaternion):
         n = euler.shape[:-1]
         alpha, beta, gamma = euler[..., 0], euler[..., 1], euler[..., 2]
 
-        if convention.lower() == "krakow_hielscher":
+        if convention == "Krakow_Hielscher":
             # To be applied to the data found at:
             # https://www.repository.cam.ac.uk/handle/1810/263510
             alpha -= np.pi / 2
@@ -378,7 +378,7 @@ class Rotation(Quaternion):
             rot = cls(data.data)
             rot.improper = zero
             return rot
-        elif convention.lower() == "bunge":
+        elif convention == "bunge":
             # Uses A.5 & A.6 from Modelling Simul. Mater. Sci. Eng. 23
             # (2015) 083501
             sigma = 0.5 * np.add(alpha, gamma)
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -350,7 +350,7 @@ class Rotation(Quaternion):
             )
 
         euler = np.array(euler)
-        if np.any(euler > 9):
+        if np.any(np.abs(euler) > 9):
             warnings.warn(
                 "Angles are assumed to be in radians, but degrees might have been "
                 "passed"
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -279,7 +279,7 @@ class Rotation(Quaternion):
             Array of Euler angles in radians.
         """
         # TODO: other conventions
-        if convention.lower() != "bunge":
+        if convention != "bunge":
             raise ValueError(
                 "The chosen convention is not supported, 'bunge' is the only option"
             )
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -19,28 +19,29 @@
 """Point transformations of objects.
 
 Rotations are transformations of three-dimensional space leaving the
-origin in place. Rotations can be parametrized numerous ways, but in orix are
-handled as unit quaternions. Rotations can act on vectors, or other rotations,
-but not scalars. They are often most easily visualised as being a turn of a
-certain angle about a certain axis.
+origin in place. Rotations can be parametrized numerous ways, but in
+orix are handled as unit quaternions. Rotations can act on vectors, or
+other rotations, but not scalars. They are often most easily visualised
+as being a turn of a certain angle about a certain axis.
 
 .. image:: /_static/img/rotation.png
    :width: 200px
    :alt: Rotation of an object illustrated with an axis and rotation angle.
    :align: center
 
-Rotations can also be *improper*. An improper rotation in orix operates on
-vectors as a rotation by the unit quaternion, followed by inversion. Hence,
-a mirroring through the x-y plane can be considered an improper rotation of
-180° about the z-axis, illustrated in the figure below.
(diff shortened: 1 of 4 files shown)


orix/quaternion/rotation.py
```diff
@@ -260,23 +265,24 @@ def from_neo_euler(cls, neo_euler):
...
         Returns
         -------
         ndarray
             Array of Euler angles in radians.
-
         """
-        if convention != "bunge":
-            raise ValueError("The convention you have specified is not supported")
+        # TODO: other conventions
+        if convention.lower() != "bunge":
```
pc494, line 282: here

orix/quaternion/rotation.py
```diff
@@ -321,28 +327,40 @@ def to_euler(self, convention="bunge"):  # TODO: other conventions
...
         direction : str
-            'lab2crystal' or 'crystal2lab'
+            "lab2crystal" or "crystal2lab".
         """
-        if convention not in ["bunge", "Krakow_Hielscher"]:
-            raise ValueError("The chosen convention is not one of the allowed options")
-        if direction not in ["lab2crystal", "crystal2lab"]:
-            raise ValueError("The chosen direction is not one of the allowed options")
+        conventions = ["bunge", "krakow_hielscher"]
+        if convention.lower() not in conventions:
```
pc494, line 342: here

orix/quaternion/rotation.py
```diff
@@ -321,28 +327,40 @@ def to_euler(self, convention="bunge"):  # TODO: other conventions
...
+            raise ValueError(
+                f"The chosen convention is not one of the allowed options {conventions}"
+            )
+        directions = ["lab2crystal", "crystal2lab"]
+        if direction.lower() not in directions:
```
pc494, line 347: here

orix/quaternion/rotation.py
```diff
@@ -321,28 +327,40 @@ def to_euler(self, convention="bunge"):  # TODO: other conventions
...
+            raise ValueError(
+                f"The chosen direction is not one of the allowed options {directions}"
+            )
+
+        euler = np.array(euler)
+        if np.any(euler > 9):
```
pc494, line 353: probably better with `np.abs(euler)`

orix/quaternion/rotation.py
```diff
@@ -321,28 +327,40 @@ def to_euler(self, convention="bunge"):  # TODO: other conventions
...
+            warnings.warn(
+                "Angles are assumed to be in radians, but degrees might have been "
+                "passed"
+            )
+        n = euler.shape[:-1]
+        alpha, beta, gamma = euler[..., 0], euler[..., 1], euler[..., 2]
 
-        if convention == "Krakow_Hielscher":
+        if convention.lower() == "krakow_hielscher":
```
pc494, line 361: here

orix/quaternion/rotation.py
```diff
@@ -321,28 +327,40 @@ def to_euler(self, convention="bunge"):  # TODO: other conventions
...
+            )
+
+        euler = np.array(euler)
+        if np.any(euler > 9):
+            warnings.warn(
+                "Angles are assumed to be in radians, but degrees might have been "
+                "passed"
+            )
+        n = euler.shape[:-1]
+        alpha, beta, gamma = euler[..., 0], euler[..., 1], euler[..., 2]
```
pc494, line 359: nice, I approve of this.

orix/quaternion/rotation.py
```diff
@@ -360,17 +378,9 @@ def from_euler(cls, euler, convention="bunge", direction="crystal2lab"):
...
-            euler = np.array(euler)
-            n = euler.shape[:-1]
-
-            # Uses A.5 & A.6 from Modelling Simul. Mater. Sci. Eng. 23 (2015) 083501
-
-            alpha = euler[..., 0]  # psi1
-            beta = euler[..., 1]  # Psi
-            gamma = euler[..., 2]  # psi3
-
+        elif convention.lower() == "bunge":
```
pc494, line 381: here

orix/tests/test_rotation.py
```diff
@@ -190,52 +190,50 @@ def test_neg(rotation, i, expected_i):
...
 
-def test_edge_cases_to_euler():
-    x = np.sqrt(1 / 2)
-    q = Rotation(np.asarray([x, 0, 0, x]))
-    e = q.to_euler()
-    q = Rotation(np.asarray([0, x, 0, 0]))
-    e = q.to_euler()
+    def test_passing_degrees_warns(self):
+        with pytest.warns(UserWarning, match="Angles are assumed to be in radians, "):
+            r = Rotation.from_euler([90, 45, 90])
```
pc494, line 235: would maybe prefer this to be [90,0,0],convention="bunge" because then one can eyeball the quaternion.

pc494 (commented):
This is good. My only comment would be I don't think we should case check for inserted strings, it makes the code more complicated (and I have just fixed a bug stemming from it in hyperspy, so that's on my mind).

orix/quaternion/rotation.py
hakonanes: Isn't this better, because then if I pass "Bunge", it will correctly pass? Previously, if I pass "Bunge", I get the ValueError?

orix/quaternion/rotation.py
hakonanes: Agree!

orix/quaternion/rotation.py
pc494: If you're selecting a method with a string you should get the capitalisation correct. If the string should be "`bunge`" I would expect (and build/design around) "`Bunge`" throwing an error. Plus the error message makes it pretty obvious where you have gone wrong.

hakonanes: That's OK, I'll revert to how it was.

But in general, I disagree, because I believe that if we can allow for an easy mistake on the users part without too much effort, we should.

hakonanes: Done!

orix/quaternion/rotation.py
hakonanes: Yup

orix/quaternion/rotation.py
hakonanes: Done

orix/quaternion/rotation.py
hakonanes: Done

orix/quaternion/rotation.py
hakonanes: Done

orix/tests/test_rotation.py
hakonanes: Yeah, done!

hakonanes:
Thanks for the fast response and good comments, @pc494! They should be all handled now.

pc494:
Good stuff :)

Issue #175, opened 2021-04-21
orix.sampling.get_sample_local() is quite memory intensitve


hakonanes:
This code fills my 16 GB memory (from 6 GB used before running) immediately

```python
>>> import numpy as np
>>> from orix import sampling, quaternion
>>> r = quaternion.Rotation.from_euler(np.deg2rad([133.3, 88.7, 177.8]))
>>> r2 = sampling.get_sample_local(resolution=0.5, center=r, grid_width=5)
```

Am I wrong in assuming that this code would give me about 5^3 / 0.5 = 250 rotations about `r`? If I'm right, we should revisit the implementation of `get_sample_local()`.

pc494:
Resolution is a difficult concept here, so I won't speculate about how many rotations you might actually expect. But the current implementation is very naive (it samples all of orientation space then cuts it back down). 

I think I could improve upon this.

hakonanes:
Understandable. But 250 is a reasoanble estimate, give or take, I would assume.

I haven't used this function before now, but I believe it can be usefuly in our work. I would be happy to assist.

pc494:
The basic problems were:

1) Sampling uniformly while retaining a constraint (ie. rather than using a rejection based sampling).
2) Not introducing randomness (eg. Any method that involves a Gaussian)

I'm currently having a look and I think we probably could improve things

pc494:
Okay, I've now got an idea, could probably get this done for the start of next week, would that be okay?

hakonanes:
Yeah, totally, there is no rush on this!

hakonanes:
I believe `Object3d.unique()` is the source for the memory error, which I see by trying out #175 as well (will finish review on Sunday). I think there is room for improvement in that method, I will have a look at that hopefully next week.

pc494:
> I believe `Object3d.unique()` is the source for the memory error, which I see by trying out #175 as well (will finish review on Sunday). I think there is room for improvement in that method, I will have a look at that hopefully next week.

`"""Returns a new object containing only this object's unique entries."""`

Yeah that'll double the usage, but I think #175 should still make it a lot smaller.

pc494:
Closed by #177

PR #177, opened 2021-04-25
Improving the performance of the samplings utils (see #175)

pc494:
Improving the performance of the gridding utils by allowing a uniform sampling of a subspace (see #175 for details). I've also renamed a file to be more consistent with what I imagine this part of the code will look like in the future.

#### Progress of the PR

#### For reviewers

### DIFF
--- a/orix/sampling/SO3_sampling.py
+++ b/orix/sampling/SO3_sampling.py
@@ -95,8 +95,14 @@ def _three_uniform_samples_method(resolution, max_angle):
         e_1_min = np.cos(np.deg2rad(max_angle / 2))
         u_1_max = 1 - np.square(e_1_min)
         u_2_min = np.arcsin(e_1_min) / 2 / np.pi
-        u_1 = np.linspace(0, u_1_max, num=int(num_steps * (u_1_max)), endpoint=True)
-        u_2 = np.linspace(u_2_min, 1, num=int(num_steps * (1 - u_2_min)), endpoint=True)
+
+        # avoids the zero selection
+        num_1 = int(num_steps * (u_1_max)) if int(num_steps * (u_1_max)) > 1 else 2
+        num_2 = int(num_steps * (1 - u_2_min)) if int(num_steps * (1 - u_2_min)) > 1 else 2
+
+
+        u_1 = np.linspace(0, u_1_max, num=num_1, endpoint=True)
+        u_2 = np.linspace(u_2_min, 1, num=num_2, endpoint=True)
 
     u_3 = np.linspace(0, 1, num=num_steps, endpoint=True)
 
--- a/orix/sampling/SO3_sampling.py
+++ b/orix/sampling/SO3_sampling.py
@@ -98,8 +98,9 @@ def _three_uniform_samples_method(resolution, max_angle):
 
         # avoids the zero selection
         num_1 = int(num_steps * (u_1_max)) if int(num_steps * (u_1_max)) > 1 else 2
-        num_2 = int(num_steps * (1 - u_2_min)) if int(num_steps * (1 - u_2_min)) > 1 else 2
-
+        num_2 = (
+            int(num_steps * (1 - u_2_min)) if int(num_steps * (1 - u_2_min)) > 1 else 2
+        )
 
         u_1 = np.linspace(0, u_1_max, num=num_1, endpoint=True)
         u_2 = np.linspace(u_2_min, 1, num=num_2, endpoint=True)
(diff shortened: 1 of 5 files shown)


pc494:
While I'm here I'm going to tidy up the class method "random": https://github.com/pyxem/orix/blob/692c54aefae59a2c3b5d1e1454e50f361d6849d3/orix/quaternion/rotation.py#L530 which is quiet sloppy at the moment.

I'll also look into the https://github.com/pyxem/orix/blob/692c54aefae59a2c3b5d1e1454e50f361d6849d3/orix/quaternion/rotation.py#L549 functionality.

pc494:
> While I'm here I'm going to tidy up the class method "random":
> 
> https://github.com/pyxem/orix/blob/692c54aefae59a2c3b5d1e1454e50f361d6849d3/orix/quaternion/rotation.py#L530
> which is quiet sloppy at the moment.
> 
> I'll also look into the
> 
> https://github.com/pyxem/orix/blob/692c54aefae59a2c3b5d1e1454e50f361d6849d3/orix/quaternion/rotation.py#L549
> functionality.

Going to transfer this into an issue to be dealt with when the large refactor comes in at some point in the future.

orix/sampling/sample_generators.py
```diff
@@ -81,7 +82,9 @@ def get_sample_local(resolution=2, center=None, grid_width=10):
     orix.sampling_utils.uniform_SO3_sample
     """
 
-    q = uniform_SO3_sample(resolution)
+    q = uniform_SO3_sample(resolution,max_angle=grid_width)
```
hakonanes, line 85: This requires almost no memory at all, great work! I haven't looked into the actual rotations I'm getting back, other than the mean and that the misorientation to the mean is not greater than the `grid_width`.

Ah, looking at a narrow sampling, I get no rotations with a grid width of 4 degrees and a resolution of half a degree:
```python
>>> import numpy as np
>>> r = quaternion.Rotation.from_euler(np.deg2rad([133.3, 88.7, 177.8]))
>>> sampling.get_sample_local(resolution=0.5, center=r, grid_width=4)
Rotation (0,)
[]
```

Is this as expected?

Checking the sample sizes
```python
>>> widths = np.arange(1, 10)
>>> resolution = 0.5
>>> sample_sizes = np.zeros_like(widths)
>>> r = quaternion.Rotation.from_euler(np.deg2rad([133.3, 88.7, 177.8]))
>>> for i in range(widths.size):
..         r2 = sampling.get_sample_local(resolution=resolution, center=r, grid_width=widths[i])
..         print(widths[i], r2.size)
1 0
2 0
3 0
4 0
5 10
6 11
7 13
8 7924
9 18711
```

Looking into `uniform_SO3_sample()`, when `max_angle` is True and below 5, `u1` is empty because `u1max` is small, on the order of 1e-7.

orix/sampling/SO3_sampling.py
```diff
@@ -0,0 +1,158 @@
...
+
+    if max_angle is None:
+        u_1 = np.linspace(0,1,num=num_steps,endpoint=True)
+        u_2 = np.linspace(0,1,num=num_steps,endpoint=True)
+    else:
+        # e_1 = cos(omega/2) = np.sqrt(1-u_1) * np.sin(2*np.pi*u2)
+        e_1_min = np.cos(np.deg2rad(max_angle/2))
+        u_1_max = 1 - np.square(e_1_min)
+        u_2_min = np.arcsin(e_1_min) / 2 / np.pi
+        u_1 = np.linspace(0,u_1_max,num=int(num_steps*(u_1_max)),endpoint=True)
```
hakonanes, line 95: `int(num_steps * u_1_max)` becomes zero when `u_1_max` is small (see comment to `get_sample_local()`). Is there some way to handle the case where max angle is small (1-7 degrees) and a resolution of below 1 degree?

hakonanes (changes requested):
Thanks for looking into the rotation sampling, @pc494.

I haven't fully understood the maths involved here, but the sample returned with angles of about 1-7 degrees accompanied by a resolution in the order of 0.5 degrees returns an empty sampling. Could you explain this to me?

Apart from this, I like that you've renamed the utilities to SO3, so that we can easily fit SO2 in the future into this `orix.sampling` module.

pc494:
Thanks @hakonanes for this review, it's very good. It'll take me a little while to loop back to this but when I do I think most of the style changes will be simple to implement. The version in `0.6` is going to be a bit lacking (although we can fix the bug that gives back empty regions) and I've got a sense of what a more complete implementation will look like (probably for `0.7.0`)

hakonanes:
That sounds like a good plan. As mentioned in #175, there is no rush.

pc494 (commented):
Working on these locally, will be up shortly.

hakonanes (approved):

hakonanes:
Good stuff! I'll let you merge yourself, @pc494.

PR #196, opened 2021-05-23
Efficient calculation of misorientation angles via a new Orientation.get_distance_matrix() supporting Dask

### LINKED ISSUE #191: Computing misorientations
We should add `orix.quaternion.Orientation.angle_with()` to consider symmetry, i.e. overwrite `Rotation.angle_with()`. This could be done by generating all symmetrically equivalent orientations, computing the angles, and taking the minimum angle. We could add a public method `Orientation.symmetrise()` to get the symmetrically equivalent orientations as a new instance.

I haven't looked at the `_distance()` function to calculate misorientations, but I guess this approach would be an alternative?

### LINKED ISSUE #192: Going from almost equal rotations to orientations introduces a misorientation
It seems like `Orientation.set_symmetry()` with point group m-3m introduces a misorientation of about 90 degrees to two rotations which are almost identical:

```python
>>> import numpy as np
>>> from orix.quaternion import Orientation, Rotation
>>> from orix.vector import AxAngle
>>> r = Rotation.from_euler(np.deg2rad([[-89.9, 45, 134.9], [-90.0, 44.9, 45]]))
>>> print(np.rad2deg(r[0].angle_with(r[1]).data))
[89.97083784]
>>> o = Orientation(r).set_symmetry(Oh)
>>> print(np.rad2deg(o[0].angle_with(o[1]).data))
[119.88459286]

# Rotate to almost perfect alignment

>>> r[1] = Rotation.from_neo_euler(AxAngle.from_axes_angles([0, 0, 1], -np.pi / 2)) * r[1]
>>> print(np.rad2deg(r[0].angle_with(r[1]).data))
[0.12587899]

# Misorientation of about 90 degrees reintroduced?

>>> o2 = Orientation(r).set_symmetry(Oh)
>>> print(np.rad2deg(o2[0].angle_with(o2[1]).data))
[89.90008731]
```
I would expect that the last misorientation angle would be approximately 0. Am I wrong? With MTEX' `orientation.angle()` method I get a misorientation angle of about 0.

hakonanes:
Add an alternative implementation of `Orientation.distance()` in `Orientation.get_distance_matrix()`, which uses Dask for memory and time efficiency. Only misorientations to itself is supported. Misorientation angles are obtained in these steps:
1. Get misorientation quaternions: `misorientations = (~self) * other`
2. Get dot product of every misorientation to every symmetry element: `dp_all = misorientation.dot_outer(symmetry)`
3. Keep highest dot product among the symmetrically equivalent ones
4. Return misorientation angle as 2 * arccos(highest dot product)

`Orientation.get_distance_matrix()` accepts three parameters (with their defaults):
* `lazy=False`: Whether to obtain the misorientation angle with Dask (True) out-of-memory, or with NumPy in-memory. The need for Dask presented itself in step 3. above, where we get to an array of shape (100, 200, 100, 200, 48) and float64 if we want to calculate the misorientation angle between every orientation in a (100, 200) orientation instance.
* `chunk_size=20`: Number of orientations along each axis to compute the outer dot product with when using Dask. Lower chunk size takes longer time but uses less memory.
* `progressbar=True`: Whether to show a progressbar for the misorientation angle calculation when using Dask.

Additional methods using in `Orientation.get_distance_matrix()`:
* `Orientation.dot_outer()`, `Orientation._dot_outer_dask()`
* `Rotation._outer_dask()`
* `Quaternion._outer_dask()`

`Orientation.angle_with()` returning misorientation angles as with `Rotation.angle_with()`, but respecting symmetry, was implemented. This was done via implementing `Orientation.dot()` which returns the highest dot product among the symmetrically equivalent orientations.

I've added the "Clustering across fundamental region boundaries" notebook to the user guide, with some minor modifications, and added lines using `Orientation.get_distance_matrix()` but commented them out. [Documentation built from this PR](https://orix--196.org.readthedocs.build/en/196/clustering_across_fundamental_region_boundaries.html).

Close #191. Close #192.

**Note: This is not an alternative to `Misorientation.distance()`.**

#### Progress of the PR
- [x] Upload orientation clustering notebook from orix-demos (not the misorientation one, yet)
- [x] Check if we can get the misorientation angle by just considering q0 (or a): *I couldn't.*
- [x] Rename `Orientation.distance_to()` to -> `Orientation.get_distance_matrix()`, and only allow getting the distance with itself
- [x] Add all new public methods and functions to the API reference

The distance matrix `D` obtained below is the same as the pre-computed one in "ori-distance((100, 200)).npy" available [here](https://drive.google.com/drive/folders/1jJBWlOD3Q7klpEU3PvtSrp7Lmcn5o1af)

```python
>>> import numpy as np
>>> from orix.quaternion.symmetry import D6
>>> euler = np.loadtxt("Ti_orientations.ctf", skiprows=1, usecols=(0, 1, 2))
>>> o = Orientation.from_euler(np.radians(euler), convention="Krakow_Hielscher")
>>> o = o.reshape(381, 507)
>>> o = o[-100:, :200]
>>> o = o.set_symmetry(D6)
>>> D = o.get_distance_matrix(lazy=True, chunk_size=40)
[########################################] | 100% Completed |  1min  6.4s
```

This example was run on my laptop with 16 GB RAM. Memory usage fluctuated between 3,7 GB and 7,8 GB during the calculation. I've got four cores of type Intel i7-7600U CPU @ 2.80GHz.

#### For reviewers


hakonanes:
I seem to have messed up rebasing, will look at this later...

hakonanes:
The "Clustering across fundamental region boundaries" notebook can be viewed in [the documentation built from this PR](https://orix--196.org.readthedocs.build/en/196/clustering_across_fundamental_region_boundaries.html).

pc494:
This would be a huge speed up, although I think you've actually done two things here, one is using `dask` and the other is an algorithm change. I'm going to check that these are theoretically equivalent; did you get the same `D` matrix?

pc494:
Also now is the time to have the conversation about what we call these functions.

orix/quaternion/quaternion.py
```diff
@@ -229,20 +184,154 @@ def triple_cross(cls, q1, q2, q3):
...
+            + da.einsum(sum_over, d1, b2)
+            + da.einsum(sum_over, a1, c2)
+            - da.einsum(sum_over, b1, d2)
+        )
+        d = (
+            +da.einsum(sum_over, d1, a2)
+            - da.einsum(sum_over, c1, b2)
+            + da.einsum(sum_over, b1, c2)
+            + da.einsum(sum_over, a1, d2)
+        )
```
pc494, line 334: These lines aren't really a "dot product", in fact it looks like you're just providing the quaternion multiplication explicitly (possibly with vectorisation?)

pc494:
Following that comment it looks like what you are doing is:

qmult
 
Which adjusts the algorithm statement to:

> 
> Add an alternative implementation of `Misorientation.distance()` in `Orientation.distance_to()`, which uses Dask for memory and time efficiency. Misorientation angles are obtained in these steps:
>     1. Combine the symmetries of both orientation instances: `symmetry = self.symmetry.outer(other.symmetry).unique()`
>     2. Get misorientation quaternions: `misorientations = (~self) * other`
>     3. Get dot product of every misorientation to every symmetry element: `dp_all = misorientation.dot_outer(symmetry)`

Find the product of these two Orientations, which corresponds to all symmetrically equivalent misorinetations. This is returned by the statement in `_outer_dask`:`
return da.stack((a, b, c, d), axis=-1).rechunk(new_chunks)`

>
>4. Keep highest dot product among the symmetrically equivalent ones

Command is : `dp = da.max(abs(dp_all), axis=-1)`

This gives the correct answer most of the time (as our terms are cos(omega/2), sin(omega/2) [nx,ny,nz]) but in occasional cases (angles near 90 degrees etc.) it will give the wrong answer. Also you just need the first column anyway so it's no more difficult to do it correctly: 

```python

[4]: z = np.asarray([[1,0,0,0],[0.5,0.5,0,0]])                               

In [5]: z                                                                       
Out[5]: 
array([[1. , 0. , 0. , 0. ],
       [0.5, 0.5, 0. , 0. ]])

In [6]: np.max(z,axis=-1)                                                       
Out[6]: array([1. , 0.5])

In [7]: z = np.asarray([[1,0,0,0],[0,1,0,0]])                                   

In [8]: np.max(z,axis=-1)                                                       
Out[8]: array([1, 1])

```

>     5. Return misorientation angle as 2 * arccos(highest dot product)
this is actual just an inversion of the equation that `q_0 = cos(omega/2)` as far as I can tell.

=============================

A lot there, but if you agree with this assessment the code basically follows the route described in the density based clustering manuscript, but is much much faster, which would be excellent news!

orix/quaternion/quaternion.py
hakonanes: Yes, this is a Dask implementation of `Quaternion.outer()`, nothing less.

hakonanes: I added it as a private method because I don't want us to support returning Dask arrays instead of (Scalar with) NumPy arrays. If you want we could add a `lazy` boolean parameter to `Quaternion.outer()`, instead of having this separate method? Then we can only allow `other` to be of type `Quaternion` or a derived type when `lazy=True`.

pc494: This is fine as is. I will have another read of the code here.

pc494:
I'm still softly of the belief that you one shouldn't use dot products for this, see https://stackoverflow.com/questions/21513637/dot-product-of-two-quaternion-rotations

Probably easier to discuss over zoom though, with key points reproduced here.

hakonanes:
First, thank you for giving your thoughts right away @pc494.

As stated above, `Quaternion._outer_dask()` does exactly what's done in `Quaternion.outer()` and presented in equation (1.4) in https://github.com/pyxem/orix/pull/196#issuecomment-847082249. Where's the screenshot from by the way?

What I gather from [the SO discussion](https://stackoverflow.com/questions/21513637/dot-product-of-two-quaternion-rotations) you linked is that as long as the quaternions are normalized, the dot product between them can be considered as a safe measure of their similarity?

Lastly, I should say that I don't consider this PR done, but a work in progress! Instead of using the WIP tag, I prefer marking the PR as a "draft", and pressing the "Ready for review" when it's ready for review. I'm sorry if my adding and then removing the tag gave the impression that the PR is ready for review.

hakonanes:
I should have specified before that `Orientation.distance_to()` is only an alternative to `Orientation.distance()`, not `Misorientation.distance()`. We cannot get the misorientation distance matrix from `distance_to()`.

Implementation of this computation with Dask should be done in another PR.

hakonanes:
The "Clustering orientations" notebook, with cell output uploaded from running on my laptop (i.e. not executed by Read The Docs), using Orientation.distance_to(), can be viewed [here](https://orix--196.org.readthedocs.build/en/196/clustering_orientations.html).

pc494:
> As stated above, `Quaternion._outer_dask()` does exactly what's done in `Quaternion.outer()` and presented in equation (1.4) in [#196 (comment)](https://github.com/pyxem/orix/pull/196#issuecomment-847082249). Where's the screenshot from by the way?

Current draft of the relevant chapter of my thesis, the equation is lifted from wikipedia though.

> What I gather from [the SO discussion](https://stackoverflow.com/questions/21513637/dot-product-of-two-quaternion-rotations) you linked is that as long as the quaternions are normalized, the dot product between them can be considered as a safe measure of their similarity?

We can discuss this in a meeting; it suffices to say I'm not convinced.

> Lastly, I should say that I don't consider this PR done, but a work in progress! Instead of using the WIP tag, I prefer marking the PR as a "draft", and pressing the "Ready for review" when it's ready for review. I'm sorry if my adding and then removing the tag gave the impression that the PR is ready for review.

Yup, I figured as much, I was trying to steer clear of the code and mainly focus on confirming that the we understood what the implementation was doing.

hakonanes:
I've now specified in the *Notes* of the `Orientation.distance_to()` docstring how the smallest angle is found (maths and wording can be improved):

bilde

hakonanes:
OK, so from our discussion off-GitHub @pc494, I believe this was what we converged on:
- Check if we can get the misorientation angle by just considering q0 (a)
- Rename `Orientation.distance_to()` to -> `Orientation.get_distance_matrix()`, and only allow getting the distance with itself

Please remind me if I'm missing other concrete points.

pc494:
Yes, this is as I recall it.

hakonanes:
I can't seem to get the misorientation angle by just considering q0. Most distances are identical by using the full quaternion or just q0, but I also get quite a lot of differences up to a maximum of 2. I'll leave it for someone else to try this some more.

So, this is now ready for review.

orix/quaternion/orientation.py
```diff
@@ -243,12 +247,21 @@ class Orientation(Misorientation):
...
-            misorientation = Misorientation(self * ~other).squeeze()
-            return misorientation.set_symmetry(self.symmetry, other.symmetry)
-        return NotImplemented
+    @property
+    def symmetry(self):
+        """Symmetry."""
+        return self._symmetry[1]
+
+    @property
+    def unit(self):
```
pc494, line 256: physically what is this?

orix/quaternion/orientation.py
```diff
@@ -326,6 +338,114 @@ def from_neo_euler(cls, neo_euler, symmetry=None):
...
+        misorientation = (~self).outer(other)
+        all_dot_products = Rotation(misorientation).dot_outer(symmetry).data
+        highest_dot_product = np.max(all_dot_products, axis=-1)
+        return Scalar(highest_dot_product)
+
+    def get_distance_matrix(self, lazy=False, chunk_size=20, progressbar=True):
+        r"""The symmetry reduced smallest angle of rotation transforming
+        each orientation in this instance to every other orientation.
+
+        This is an alternative implementation of
```
pc494, line 391: Should we mark the old route as deprecated?

orix/quaternion/rotation.py
```diff
@@ -581,6 +582,30 @@ def antipodal(self):
...
+        -------
+        dask.array.Array
+
+        Notes
+        -----
+        To get a new rotation from the returned array `rarr`, do
+        `r = Rotation(rarr.compute())`.
+        """
+        r = super()._outer_dask(other, chunk_size=chunk_size)
+        #        r[..., -1] = np.logical_xor.outer(self.improper, other.improper)
```
pc494, line 606: drop this line

pc494 (commented):
This is basically good to go, I'm happy to merge in with this distance metric, and will raise an issue to remind us to potentially improve it at a later date.

hakonanes:
Thanks for the quick response!

I would be very much appreciate your comments, if you have any, on:
* The updated API for `Orientation`: https://orix--196.org.readthedocs.build/en/196/reference.html#orientation
* The "Clustering across fundamental region boundaries" notebook added to the orix user guide: https://orix--196.org.readthedocs.build/en/196/reference.html#orientation
* The "Clustering orientations" notebook also added to the user guide: https://orix--196.org.readthedocs.build/en/196/clustering_orientations.html

I haven't changed any of the points made or the end result in the notebook, mostly changed some wording to make things a little bit more clear (in my opinion) and variable naming.

orix/quaternion/orientation.py
hakonanes: I can't answer that...

Is it by design, that when you do `Orientation.unit`, you get an orientation instance with no symmetry? If so, I can remove it.

orix/quaternion/orientation.py
hakonanes: Yes, why not, I can make such a warning pop up upon calls to `Orientation.distance()`.

orix/quaternion/rotation.py
hakonanes: Done.

orix/quaternion/orientation.py
pc494: Let's address this when we have a broader sweep of the code.

pc494 (approved):
The documentation changes are good - I think this only needs one deprecation warning to be added (see comment) and then it's good to merge.

orix/quaternion/orientation.py
hakonanes: I ended up writing a `deprecated` decorator based on similar decorators in scikit-image and matplotlib (see links in the decorator docstring):

```python
    @deprecated(
        since="0.7",
        alternative="orix.quaternion.Orientation.get_distance_matrix",
        removal="0.8",
    )
    def distance(self, verbose=False, split_size=100):
        return super().distance(verbose=verbose, split_size=split_size)
```

```python
>>> Orientation.random_vonmises(10).distance()
/home/hakon/kode/orix/orix/quaternion/orientation.py:389: VisibleDeprecationWarning: Function `distance()` is deprecated and will be removed in version 0.8. Use `orix.quaternion.Orientation.get_distance_matrix()` instead.
  since="0.7",
```

hakonanes: I think for orix 0.8, we can change the above to:

```python
def distance(self, *args, **kwargs):
    return NotImplemented
```

What do people think? We have to overwrite it, since `Orientation` inherits from `Misorientation`.

pc494: yup, that works for me

PR #201, opened 2021-06-09
Euler angle color key

### LINKED ISSUE #244: Rotation.to_euler() shouldn't return negative Euler angles
`Rotation.to_euler()` should return Euler angles in the ranges [0, 2pi], [0, pi], [0, 2pi] according to the first author of the 2015 paper ["Consistent representations of and conversions between 3D rotations"](http://dx.doi.org/10.1088/0965-0393/23/8/083501), Dave Rowenhorst, if we follow the definitions of that paper, which we say we do [in the code](https://github.com/pyxem/orix/blob/master/orix/quaternion/rotation.py#L378). However, I get negative Euler angles:

```python
>>> import numpy as np
>>> rot = Rotation.from_euler(np.radians((45, 10, 45)))
>>> np.degrees(rot.to_euler())
array([[-45.,  10.,  45.]])
```

After a quick check, it doesn't seem like MTEX ever returns negative Euler angles, either.

I think we should always return positive Euler angles.

hakonanes:
Add Euler angle coloring which looks like so compared to the IPF coloring for the MTEX dataset "olivine"

```python
>>> from orix import io, plot
>>> xmap = io.load("olivineopticalmap.ang")  # From MTEX
>>> xmap
Phase    Orientations          Name  Space group  Point group  Proper point group       Color
    1   44953 (90.4%)       olivine         None          222                 222     tab:red
    2     1370 (2.8%)      Dolomite         None            3                   3   tab:green
    3     2311 (4.6%)     Enstatite         None          222                 222  tab:orange
    4     1095 (2.2%)  Chalcopyrite         None          422                 422    tab:blue
Properties: iq, ci, unknown1, fit, unknown2, unknown3, unknown4, unknown5
Scan unit: um
>>> pg = xmap.phases["olivine"].point_group
>>> ori = xmap.phases["olivine"].orientations
>>> ckey1 = plot.EulerColorKey(pg)
>>> ckey1.plot()
>>> rgb1 = ckey.orientation2color(ori)
>>> xmap["olivine"].plot(rgb)
>>> ckey2 = plot.IPFColorKeyTSL(pg)
>>> ckey2.plot()
>>> rgb2 = ckey2.orientation2color(ori)
>>> xmap["olivine"].plot(rgb2)
```

Euler | IPF
:--:|:---:
euler_key|ipf_key
euler_map|ipf_map

I've added ~two~ one "naive" color key~s~:
* `EulerColorKey` as described by Gert Nolze in https://doi.org/10.1002/crat.201400427. I've based the implementation on that paper and the implementation in MTEX ([fundamentalRegionEuler](https://github.com/mtex-toolbox/mtex/blob/develop/geometry/%40symmetry/fundamentalRegionEuler.m) and [project2EulerFR](https://github.com/mtex-toolbox/mtex/blob/develop/geometry/%40quaternion/project2EulerFR.m))
* ~`AxisColorKey` as implemented here: https://github.com/usnistgov/OOF2/blob/master/SRC/engine/angle2color.C#L52~ Dropped this as it is highly discontinuous.

~The latter suffers from color jumps, which is as expected. The former shouldn't, but does as I've implemented it now. The main issue is the "projection" of orientations into a fundamental Euler region defined by angle ranges for the three angles, which is implemented in `Orientation.in_euler_fundamental_region()`. The projection is not correct, and I believe it is related to the differing definitions of Euler angles employed in orix and in Gert's paper. MTEX has its own definition, so I cannot get their approach to work with our Euler angles either. This PR is therefore a work in progress~...This should now be solved, and Euler angles for all (proper) point groups should be colored using the full RGB range of the Euler angle range.

Other changes:
* Ensure only positive Euler angles are returned from `Rotation.to_euler()` following Rowenhorst et al. (2015). This was necessary for Euler angle coloring to work. Fixes #244.
* Touched up some docstrings, other minor clarifications

#### Progress of the PR
- [x] Correct projection of Euler angles into their fundamental region
- [x] User guide notebook explaining orientation coloring
- [x] Plotting of color key

#### For reviewers


SteffenBrinckmann:
Hey, 
This code I had used in the past with an non-orix project to calculate the IPF-color from the quaternions. Since orix also uses quaternions, this code should apply with modifications:

    def IPFcolor(self,axis, proper=False):
         """color of inverse pole figure for given axis

        Args:
           axis: axis of pole figure (ND=001)
           proper: considers only vectors with z >= 0, hence uses two neighboring SSTs to determine if in SST. i.e. allows more positive results, rgb-value does not depend on this

        Returns:
           vector of color (rgb)
        """
        color = np.zeros(3,'d')
        for q in self.symmetry.equivalentQuaternions(self.quaternion):
          pole = q.conjugated()*axis                                                                    # align crystal direction to axis
          inSST,color = self.inSST(pole,color=True,proper=proper)
          if inSST: break
        return color

and

`  def inSST(self, vector, proper = False,  color = False):
    '''
    Check whether given vector falls into standard stereographic triangle of own symmetry.

    Args:
       vector: hkl vector tested/converted to rbg; considers only this vector, no crystal orientation
       proper: considers only vectors with z >= 0, hence uses two neighboring SSTs to determine if in SST. i.e. allows more positive results, rgb-value does not depend on this
       color: if true, return also color

    Returns:
       whether in SST
       (if requested) inverse pole figure color if in SST in RGB values
    '''
    if self.lattice == 'cubic':
      basis = {'improper':np.array([ [-1.            ,  0.            ,  1. ],
                                   [ np.sqrt(2.)   , -np.sqrt(2.)   ,  0. ],
                                   [ 0.            ,  np.sqrt(3.)   ,  0. ] ]),
             'proper':np.array([ [ 0.            , -1.            ,  1. ],
                                   [-np.sqrt(2.)   , np.sqrt(2.)    ,  0. ],
                                   [ np.sqrt(3.)   ,  0.            ,  0. ] ]),
              }
    elif self.lattice == 'hexagonal':
      basis = {'improper':np.array([ [ 0.            ,  0.            ,  1. ],
                                   [ 1.            , -np.sqrt(3.),     0. ],
                                   [ 0.            ,  2.            ,  0. ] ]),
             'proper':np.array([ [ 0.            ,  0.            ,  1. ],
                                   [-1.            ,  np.sqrt(3.)   ,  0. ],
                                   [ np.sqrt(3)    , -1.            ,  0. ] ]),
              }
    elif self.lattice == 'tetragonal':
      basis = {'improper':np.array([ [ 0.            ,  0.            ,  1. ],
                                   [ 1.            , -1.            ,  0. ],
                                   [ 0.            ,  np.sqrt(2.),     0. ] ]),
             'proper':np.array([ [ 0.            ,  0.            ,  1. ],
                                   [-1.            ,  1.            ,  0. ],
                                   [ np.sqrt(2.)   ,  0.            ,  0. ] ]),
              }
    elif self.lattice == 'orthorhombic':
      basis = {'improper':np.array([ [ 0., 0., 1.],
                                   [ 1., 0., 0.],
                                   [ 0., 1., 0.] ]),
             'proper':np.array([ [ 0., 0., 1.],
                                   [-1., 0., 0.],
                                   [ 0., 1., 0.] ]),
              }
    else:
      basis = {'improper':np.zeros((3,3),dtype=float),
             'proper':np.zeros((3,3),dtype=float),
              }
    #theComponents = color components
    #inSST = vector of true/false for each data-point
    if np.all(basis == 0.0):  #default in basis not set
      theComponents = -np.ones(3,'d')
      inSST = False   #np.all(theComponents >= 0.0), this statement is always false
    else:
      v = np.array(vector,dtype = float)
      if proper:                                                                                  # check both improper ...
        theComponents = np.dot(basis['improper'],v)
        inSST = np.all(theComponents >= 0.0, axis=0)
        if not np.any(inSST):                                                                               # ... and proper SST
          theComponents = np.dot(basis['proper'],v)
      else:
        v[2] = abs(v[2])                                                                            # z component projects identical for positive and negative values
        theComponents = np.dot(basis['improper'],v)
      inSST = np.all(theComponents >= 0.0, axis=0)
    if color:                                                                                       # have to return color array
      if np.any(inSST):
        theComponentsNorm = theComponents/np.linalg.norm(theComponents, axis=0)
        rgb = np.power(np.abs(theComponentsNorm),0.5)                             # smoothen color ramps
        if rgb.ndim>1: rgb[:,~inSST] = 0.0
        rgb[rgb>1.0] = 1.0                                                        # limit to maximum intensity
        if rgb.ndim>1: rgb[:,inSST] /= np.max(rgb[:,inSST], axis=0)                                                                             # normalize to (HS)V = 1
        else         : rgb          /= np.max(rgb)
      else:
        rgb = np.zeros(3,'d')
      return (inSST,rgb)
    else:
      return inSST`

hakonanes:
@SteffenBrinckmann, thanks for this, it's from https://github.com/SteffenBrinckmann/pythonEBSD, correct? We should compare orientation colors produced in orix, when we get to implementing that, to what pythonEBSD produces for the same point groups. In this PR I only try to establish an orientation color framework similar to MTEX', with Euler angle coloring as per Gert Nolze's paper https://doi.org/10.1002/crat.201400427.

SteffenBrinckmann:
Ok, I see the difference. My bad.

hakonanes:
Not at all, I'm sure your code snippet will be helpful in the future!

pc494:
Coming at this from somewhere else - I think the difference in conventions might be resolved by using the kwarg `direction="crystal2lab"` in `from_euler()` - there is some discussion of this here https://mtex-toolbox.github.io/MTEXvsBungeConvention.html

hakonanes:
I've confirmed that by bypassing `Orientation.set_symmetry()` in `CrystalMap.orientations` and just do `ori._symmetry = (C1, symmetry)`, we get a reasonable color map similar to MTEX' for the `mtexdata olivine` dataset:

MTEX | with set_symmetry | without set_symmetry
:--:|:---:|:---:
olivine_mtex|olivine_orix1|olivine_orix2

pc494:
Has this been killed off by #235?

hakonanes:
No, I suggest we finish this, as it might be useful when people don't want or need IPF colours? Also, definitions of symmetry's fundamental Euler region and projection of Euler angles to this region might be useful in other situations.

The major thing left is to plot the colour key, which I'm not entirely sure of how to do (three bars RGB with angle ranges?), but will try to do before v0.8.

pc494:
Okay, I'm happy to trust your judgement here.

din14970:
Typically this kind of coloring is used as a qualitative method to distinguish grains. The goal is simple visual separation. To get a good idea of different grains with IPF, you really need at least two of them and compare both. In a highly textured material, the IPF-Z map might show everything in the same color; one can not infer anything about how crystals are rotated around Z. With the Euler coloring all information is retained. It's nearly impossible to interpret quantitatively but you'll be quite sure that grains next to each other with different orientation will also be a different color.

pc494:
> Typically this kind of coloring is used as a qualitative method to distinguish grains. The goal is simple visual separation. To get a good idea of different grains with IPF, you really need at least two of them and compare both. In a highly textured material, the IPF-Z map might show everything in the same color; one can not infer anything about how crystals are rotated around Z. With the Euler coloring all information is retained. It's nearly impossible to interpret quantitatively but you'll be quite sure that grains next to each other with different orientation will also be a different color.

Thanks Niels, I had missed this application as (one has to plug ones own work from the time to time!) I would normally apply the density based clustering method. But you're right, a simpler, quicker, easier to understand approach is obviously an improvement for the code.

hakonanes:
> Typically this kind of coloring is used as a qualitative method to distinguish grains.

Very good point, thank you @din14970! Perhaps the colour key isn't crucial since this colouring is considered qualitative...

pc494:
I guess this is also in a tidying up phase?

hakonanes:
I'm finishing this as we speak, will mark ready for review hopefully in a few hours!

hakonanes:
> I'm finishing this as we speak, will mark ready for review hopefully in a few hours!

This needed more work than I thought, will see if I spend some more time on this within the weekend to solve the cubic case (without brute force), or just treat the cubic case as all the other ones. By "cubic case" I mean that all symmetrically equivalent Euler angles of proper point groups cannot be found just by rotating about the primary axis and any perpendicular, but also about [111].

hakonanes:
This is now on the brink of completion, only tests remain. See the updated OP.

pc494:
will review this on Monday.

pc494:
@hakonanes can you confirm that you checked these results are the same as those produced by MTEX as quality assurance?

pc494 (approved):
Happy with these, just want verbal confirmation of the above and then will merge.

hakonanes:
I'll have a quick look and "fix" the code style.

hakonanes:
The implementations of the coloring of Euler angles are identical in orix and MTEX, it's the definition of the fundamental Euler region that differs. Thus, they produce different results. The most important thing is that the colors in orix are without color jumps and that the color key is correct. This should be the case.

hakonanes:
Made a *final* update to the fundamental Euler regions so that they correspond to the ones in MTEX and not the ones in Table 5 in Nolze's paper, which they didn't anyway because I couldn't get a reduction of Phi to the ranges [0, 45] and [0, 54.74] for point groups 23 and 432, respectively. Ranges for these are now [0, 90].

hakonanes:
Will fix the tests, sorry.

PR #202, opened 2021-06-09
Add Orientation.scatter() method for simpler plotting in axis-angle and Rodrigues space

hakonanes:

We can now call the existing functions for plotting orientations in axis-angle and Rodrigues fundamental zone space (which are of great visual aid, by the way) more easily via a `Orientation.scatter()` method

```python
>>> rot = sampling.get_sample_fundamental(resolution=15, point_group=symmetry.D6)
>>> ori = Orientation(rot).set_symmetry(symmetry.D6)
>>> ori.scatter("axangle")  # Default
```

sample_d6_10deg

We can also do

```python
>>> fig = plt.figure(figsize=(10, 5))
>>> Orientation.random_vonmises(100).scatter(figure=fig, position=(1, 2, 1), c="r")
>>> fig.axes[0].axis("on")
>>> Orientation.random_vonmises(100).scatter(figure=fig, position=122, c="b")
```

multiple_plots

Restricting the number of orientations to plot by drawing a random sample of a given size

```python
>>> rot = sampling.get_sample_fundamental(resolution=6, point_group=symmetry.D6)
>>> ori = Orientation(rot).set_symmetry(symmetry.D6)
>>> sizes = [1000, 2000, ori.size]
>>> fig = plt.figure(figsize=(15, 5))
>>> for i in range(len(sizes)):
...     ori.scatter(figure=fig, position=(1, 3, i + 1), size=sizes[i])
...     fig.axes[i].set_title(f"{sizes[i]} orientations", y=0.9)
```

sample_sizes

Summary of additions:
* I've also added a `np.errstate(divide="ignore")` to `OrientationRegion.get_plot_data()` because I couldn't find another way to silence the warning.
* `Object3d.get_random_sample()`, a thin wrapper for [numpy.random.Generator.choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html#numpy.random.Generator.choice), used in the latest example above.

#### Progress of the PR
- [x] Figure out best use of `axes.set_xlim()` (and ylim and zlim)
- [x] Select a sample of points to plot when too many orientations are to be plotted (like MTEX does), and warn that we do this
- [x] Silence Matplotlib warning relating to Axes3D by passing parameter to Matplotlib added after oldest supported Matplotlib version (so we must do some version check). This is why the build with Python 3.6 and oldest dependencies fail.

#### For reviewers


hakonanes:
If we want to silence the mentioned warning in a v0.6.1 release, I must separate it out in another commit...

pc494:
> If we want to silence the mentioned warning in a v0.6.1 release, I must separate it out in another commit...

Not important enough to bother doing for a patch release as far as I am concerned. It is good to have it for `0.7` though.

hakonanes:
Great, thanks @pc494.

Question related to this PR: why does the rotation sampling return rotations and not orientations?

hakonanes:
I've requested a review, but there's no rush on this.

hakonanes:
I haven't used the `Misorientation` class at all, that's why I didn't write a `Misorientation.scatter()` instead of `Orientation.scatter()`. If there is sufficient interest in moving the method to the misorientation class instead, I can do that.

pc494:
> Great, thanks @pc494.
> 
> Question related to this PR: why does the rotation sampling return rotations and not orientations?

I think this is correct. From a python perspective `Orientation` is a subclassing of `Rotation` so given that both do apply it makes sense to choose the super class.

From a physics perspective, a **Rotation** (ie. the transformation) is being sampled, you could use such a sampling to find all the `Orientations` of an object of unknown symmettry (by multiplying in the `Rotation`s). It's marginal though & not helped by the absence of Rotation vs Orientation documentation at present.

pc494:
> I haven't used the `Misorientation` class at all, that's why I didn't write a `Misorientation.scatter()` instead of `Orientation.scatter()`. If there is sufficient interest in moving the method to the misorientation class instead, I can do that.

We can upstream it at a later date, plotting `Orientation`s is the priority. 

I'll review this now.

pc494 (approved):
Looks good to me :D

hakonanes:
> I think this is correct. From a python perspective `Orientation` is a subclassing of `Rotation` so given that both do apply it makes sense to choose the super class.

Good point, I agree. I think some users might be confused though: They supply a point group to sample, but this crucial information is lost in the returned `Rotation` instance. We should add a note and example to the docstring, showing how to convert the returned rotations to orientations.

> From a physics perspective, a **Rotation** (ie. the transformation) is being sampled, you could use such a sampling to find all the `Orientations` of an object of unknown symmettry (by multiplying in the `Rotation`s). It's marginal though & not helped by the absence of Rotation vs Orientation documentation at present.

Agree again. Such a documentation page (user guide notebook?) would be very helpful, and could utilize both these 3D projections and pole figures to highlight differences.

pc494:
This generalised up to `Misorientation` with less than 10 minutes work from me. It also looks incredible and runs really fast. A big win for when 0.7 comes out.

PR #209, opened 2021-06-23
add transpose class method

harripj:
Hi there, thanks for all the work put into Orix! I have found it very useful. I was wondering whether you would be interested in adding a transpose class method given that there is also a reshape method. I have found myself coding this up quite a few times recently and found it to be quite useful when comparing arrays of Orientations, so thought about creating a PR.
Cheers!

#### Progress of the PR

```python
>>> from orix import Orientation
>>> import numpy as np
>>> ori = Orientation(np.random.rand(5, 3, 4))
>>> oriT = ori.transpose()
>>> ori.shape, oriT.shape

>>> v = Vector3d(np.random.rand(10, 3, 4, 3))
>>> vT = v.transpose(2, 0, 1)
>>> v.shape, vT.shape
```

#### For reviewers

### DIFF
--- a/orix/base/__init__.py
+++ b/orix/base/__init__.py
@@ -54,9 +54,7 @@ class DimensionError(Exception):
 
 
 class Object3d:
-    """Base class for 3d objects.
-
-    """
+    """Base class for 3d objects."""
 
     dim = None
     """int : The number of dimensions for this object."""
@@ -205,6 +203,24 @@ class Object3d:
         obj = self.__class__(self.data.reshape(*shape, self.dim))
         obj._data = self._data.reshape(*shape, -1)
         return obj
+
+    def transpose(self, *order):
+        """Returns a new object containing the same data transposed.
+        If ndim is originally 2, then order may be undefined.
+        In this case the first two dimensions will be transposed."""
+        if not len(order):
+            if len(self.shape) != 2:
+                raise ValueError("order must be defined for more than two dimensions.")
+            else:
+                # swap first two axes
+                order = (1, 0)
+
+        if len(order) != len(self.shape):
+            raise ValueError(
+                f"Number of axes is ill-defined: {tuple(order)} does not fit with {self.shape}."
+            )
+
+        return self.__class__(self.data.transpose(*order, -1))
 
     def get_plot_data(self):
         return self


hakonanes:
Hi @harripj, and thanks for this PR!

> I have found myself coding this up quite a few times recently and found it to be quite useful when comparing arrays of Orientations

Do you mean computing the misorientation angle (or geodesic distance), like you can do with `Orientation.get_distance_matrix()` ([API](https://orix.readthedocs.io/en/latest/reference.html#orix.quaternion.Orientation.get_distance_matrix), [user guide](https://orix.readthedocs.io/en/latest/clustering_orientations.html#Compute-distance-matrix))?. If not, what do you mean?

Don't know what @pc494 thinks about transposing the shape of orientations, vectors etc., but I think it should be supported.

harripj:
Hi @hakonanes,

Thanks for your help, I wasn't familiar with that function, is it recent? And yes you are right, I should have mentioned misorientations. I don't think however this solves my problem, I will try to elaborate.

Given two sets of orientations (related, but the symmetries are not consistent) I would like to find the smallest angular distance, for example:
```python
# some random orientations
data = np.random.rand(11, 4)
ori1 = Orientation(data)

# choose a random symmetry for second set, we then know which index should be found
idx = np.random.randint(Oh.shape[0], size=ori1.shape[0])
# compute second set of orientations
ori2 = ori1.outer(Oh)[np.arange(ori1.shape[0]), idx]

# compute misorientation and angular distance matrix
dist = (ori2 * ~ori1).set_symmetry(Oh).get_distance_matrix()
dist.data.argmin(0), dist.data.argmin(1), idx

>>> (array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10], dtype=int64),
     array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10], dtype=int64),
     array([37,  9, 39,  7, 29, 17, 20, 18, 26, 45, 44]))
```

For this example the fn ```get_distance_matrix()``` appears to always give the diagonal (perhaps I am misusing or misunderstanding it), whereas I would expect the final array:

```python
(ori2 * ~ori1.outer(Oh)[np.arange(ori1.shape[0]), idx]).angle.data
>>> array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
```

For the transpose I have been using it to compute the 1-to-1 misorientation between two sets of components. For example if ```ori1.shape = (11,)``` and ```ori2.shape = (11, 48)```, I am not able to compute the 1-to-1 misorientation ```ori2 * ~ori1```, however I can compute ```ori2.transpose() * ~ori1``` to achieve the same thing.

hakonanes:
> I wasn't familiar with that function, is it recent?

Yeah, got merged into v0.7.dev0 (v0.6 is the current release) a couple weeks back (see the [changelog](https://orix.readthedocs.io/en/latest/changelog.html) for details).

> For this example the fn `get_distance_matrix()` appears to always give the diagonal (perhaps I am misusing or misunderstanding it), whereas I would expect the `idx` array, eg.:

The method is a replacement for `Orientation.distance()` used to compute the misorientation angles of a set of orientations to themselves. It might be that you find an interest in `Misorientation`s ([API](https://orix.readthedocs.io/en/latest/reference.html#misorientation), [example notebook](https://nbviewer.jupyter.org/github/pyxem/orix-demos/blob/master/03%20-%20Clustering%20Misorientations.ipynb)). To obtain misorientations: `m = ori1 - ori2`.

> For the transpose I have been using it to compute the 1-to-1 misorientation between two sets of components. For example if `ori1.shape = (11,)` and `ori2.shape = (11, 48)`, I am not able to compute the 1-to-1 misorientation `ori2 * ~ori1`, however I can compute `ori2.transpose() * ~ori1` to achieve the same thing.

Do you get the expected misorientation angles this way?

We also have the `Orientation.angle_with(other)` ([API](https://orix.readthedocs.io/en/latest/reference.html#orix.quaternion.Orientation.angle_with)) which accounts for symmetry. I think I understood your example, although I'm not 100% sure... Does this do what you want?

```python
>>> import numpy as np
>>> ori1 = Orientation.random_vonmises((11,))
>>> ori1
Orientation (11,) 1
[[-0.9413 -0.0469 -0.0772 -0.3254]
 [-0.3402 -0.3925  0.8197  0.2415]
 [-0.945   0.0203  0.042  -0.3236]
 [ 0.8793  0.4569  0.0877  0.1016]
 [ 0.8183 -0.2241 -0.5288 -0.0214]
 [-0.9152  0.0532 -0.0685 -0.3935]
 [-0.8145  0.1133  0.5511 -0.1418]
 [-0.5641  0.0947  0.7001  0.4274]
 [-0.9104  0.1918  0.2258 -0.2887]
 [-0.8875 -0.1225 -0.0059 -0.4441]
 [-0.2406  0.3532 -0.8038  0.4138]]
>>> idx = np.random.randint(Oh.size, size=ori1.size)
>>> ori2 = Orientation(ori1.outer(Oh))
>>> ori2 = ori2[np.arange(ori1.size), idx]
>>> ori2 = ori2.set_symmetry(Oh)
>>> ori2
Orientation (11,) m-3m
[[-0.9413 -0.0469 -0.0772 -0.3254]
 [ 0.897  -0.2629 -0.3152 -0.1642]
 [ 0.945  -0.0203 -0.042   0.3236]
 [-0.9449  0.2987  0.0098 -0.1338]
 [-0.9526  0.1433 -0.2047  0.1736]
 [ 0.9254 -0.0861  0.0108 -0.3689]
 [ 0.9656 -0.1804  0.1862  0.0201]
 [ 0.8939  0.2353 -0.0962 -0.3692]
 [ 0.9104 -0.1918 -0.2258  0.2887]
 [-0.9416 -0.0824 -0.0908  0.3135]
 [ 0.9057  0.1387  0.2514 -0.3119]]
>>> ori1.angle_with(ori2)
Scalar (11,)
[0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
```

pc494:
As transpose is changing the "navigation" placements of an orientation, great, put it in (which I think is what you are describing given the illusion to `reshape`).

I think the adjacent conversation is happening because if we are working with rotation matrices "transpose" does the same thing as "inversion". In this context we tend to avoid the word "transpose". Does that help?

Also, for clarity could you branch this PR request off. ie. go to your git client and run:

`git checkout -b transpose_feature`

then PR that, as working off a forks `master` is often problematic.

pc494:
Also, comments above aside, this PR looks pretty good and a solid first contribution! Thanks :)

harripj:
@hakonanes thanks for the clarification! I was not familiar with ```angle_with```, but it works well.

Just of note I have compared the three methods:
```python
(ori2.set_symmetry(C1) - ori1.set_symmetry(Oh)).angle.data
>>> array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
%timeit (ori2.set_symmetry(C1) - ori1.set_symmetry(Oh)).angle.data
>>> 400 ms ± 11.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

ori2.angle_with(ori1.set_symmetry(Oh)).data
>>> array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
%timeit ori2.angle_with(ori1.set_symmetry(Oh)).data
>>> 466 ms ± 10.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

(ori2 * ~ori1.outer(Oh).transpose()).angle.data.min(0)
>>> array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
%timeit (ori2 * ~ori1.outer(Oh).transpose()).angle.data.min(0)
>>> 1.02 ms ± 3.83 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
I am not sure why there is such a large time difference between these calculations.

@pc494 yes I was just thinking about reshaping the navigation axes rather than any orientation inversion. I will create a new PR from a new branch, thanks!

hakonanes:
> I am not sure why there is such a large time difference between these calculations.

`Orientation.set_symmetry()` ([implementation](https://github.com/pyxem/orix/blob/master/orix/quaternion/orientation.py#L157)) is not optimized, so I think most of the computation time comes from that. I think you will see a drastic time reduction in your second example with `angle_with()` ([implementation](https://github.com/pyxem/orix/blob/master/orix/quaternion/orientation.py#L343)) if you don't do `set_symmetry()` in the same line.

It might be that your third example is still faster after that, and if we get the same results from both routes, we should rewrite `angle_with()` to use the approach in the third example instead! Will have a look.

PR #210, opened 2021-06-23
Add transpose method to all 3D classes

harripj:
Duplicate of #209 with a new branch name.

#### Progress of the PR

#### For reviewers


orix/base/__init__.py
```diff
@@ -206,6 +204,24 @@ def reshape(self, *shape):
...
+            else:
+                # swap first two axes
+                order = (1, 0)
+
+        if len(order) != len(self.shape):
+            raise ValueError(
+                f"Number of axes is ill-defined: {tuple(order)} does not fit with {self.shape}."
+            )
+
+        return self.__class__(self.data.transpose(*order, -1))
```
hakonanes, line 223: An issue with orix' class inheritance is that necessary properties, like `Orientation.symmetry`, isn't carried over. Could you overwrite this method in the `Misorientation` class like so

```python
class Misorientation(Rotation):
[...]
    def transpose(self, *axes):
        mori = super().transpose(*axes)
        sym1, sym2 = self.symmetry
        mori.set_symmetry(sym1, sym2)
        return mori
```

and in `Orientation`

```python
class Orientation(Misorientation):
[...]
    def transpose(self, *axes):
        ori = super().transpose(*axes)
        ori = ori.set_symmetry(self.symmetry)
        return ori
```

I think that should work, not 100% sure. Please tell me if you encounter some issues.

orix/base/__init__.py
```diff
@@ -206,6 +204,24 @@ def reshape(self, *shape):
...
+
```
hakonanes, line 224: Regarding tests. Have you written those using pytest before? If you want to write them, I think adding some in `orix/tests/test_orientation.py` and `orix/tests/test_vector3d.py` should be OK. In the former, you should also check that the symmetry carries over.

hakonanes (changes requested):
As @pc494 mentioned, thanks for this PR! It adds functionality that definitely should be there.

I've made some requests regarding style, tests, and overwriting the method in `Orientation` and `Misorientation` to carry over `Orientation.symmetry`. Please tell me if you need any assistance with anything.

orix/base/__init__.py
hakonanes: If you don't want to write them, I could do that.

orix/base/__init__.py
```diff
@@ -206,6 +204,24 @@ def reshape(self, *shape):
         obj._data = self._data.reshape(*shape, -1)
         return obj
 
+    def transpose(self, *order):
+        """Returns a new object containing the same data transposed.
```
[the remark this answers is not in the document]
harripj: No problem!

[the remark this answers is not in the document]
hakonanes: Great! If you could just change to `axes : int, optional` that would be perfect.

hakonanes:
@harripj you can see how the docstring looks on the Read the Docs build: https://orix--210.org.readthedocs.build/en/210/reference.html#orix.base.Object3d.transpose

orix/base/__init__.py
harripj: I've just been trying to implement this, but in the case of '''Orientation''' this doesn't work as far as I can tell as '''super''' will call the method from '''Misorientation''' which will fail as '''Misorientation''' has two symmetries compared to '''Orientation''''s one. We could possibly just call '''Object3D''' init and then cast after transpose?

Another comment would be whether symmetries should or could be assumed to maintain order after transpose? Not sure on this point either, any thoughts?

harripj: Sorry for the misformatting, not currently at a computer!

hakonanes: > I've just been trying to implement this, but in the case of '''Orientation''' this doesn't work as far as I can tell as '''super''' will call the method from '''Misorientation''' which will fail as '''Misorientation''' has two symmetries compared to '''Orientation''''s one.

Pulled your changes locally and found out you can do:

```python
        mori = super().transpose(*axes)
        mori.set_symmetry(self.symmetry)
        return mori
```

> Another comment would be whether symmetries should or could be assumed to maintain order after transpose? Not sure on this point either, any thoughts?

I'm comfortable with leaving the `Symmetry` class be, since I cannot think of a relevant use case at the moment.

orix/base/__init__.py
```diff
@@ -206,6 +204,37 @@ def reshape(self, *shape):
...
+            May be undefined if self only contains two navigation dimensions.
+
+        Returns
+        -------
+        obj
+            The transposed object.
+
+        """
+        if not len(axes):
+            if len(self.shape) != 2:
```
hakonanes, line 225: I think this error is misleading when an instance is 1D. I suggest you just let a 1D array pass through to the final return.

orix/base/__init__.py
hakonanes: A (hopefully!) final request regarding overriding: the `Miller` class should pass along the `Miller.phase` and `Miller.coordinate_format`, like is done for [`Miller.flatten()`]: https://github.com/pyxem/orix/blob/dffb68e0619d8b2381de629efbc2962b8e81c284/orix/vector/miller.py#L602-L605

`Miller` stores vectors in cartesian coordinates internally as `Miller.data`, that's why `xyz` is used in initialization.

orix/base/__init__.py
harripj: Yes, I have added an initial pass through for 1d data.

orix/quaternion/orientation.py
```diff
@@ -228,6 +228,37 @@ def distance(self, verbose=False, split_size=100):
...
+        -------
+        Misorientation
+            The transposed Misorientation.
+
+        """
+        mori = super().transpose(*axes)
+        if isinstance(self.symmetry, Symmetry):
+            # only one symmetry provided, eg. Orientation
+            mori = mori.set_symmetry(self.symmetry)
+        elif isinstance(self.symmetry, tuple) and all(
```
hakonanes, line 252: Could you explain to me why my suggestion in https://github.com/pyxem/orix/pull/210#discussion_r658124898 isn't enough?

harripj: #210 works as expected for ```Misorientation```, but not so for ```Orientation``` as ```Orientation``` (as far as I am aware) has one symmetry attribute whereas ```Misorientation``` has two. In this case when ```super``` is called from ```Orientation.transpose``` it calls ```Misorientation.transpose``` which expects to unpack two symmetries but can only unpack one and crashes. I also tried this with a vararg but couldn't get it to work. Commit #e5e15e2 attempts to handle this case and in fact this means that ```Orientation.transpose``` can be handled by ```Misorientation.transpose``` (commit b1b5cb2).

harripj: I believe ```Misorientation.set_symmetry``` accepts two arguments rather than a tuple of symmetries. If I try to unpack ```*self.symmetry``` it will work for ```Misorientation``` but not ```Orientation``` (as it then unpacks the single symmetry).

hakonanes: Ah yes, sorry! I suggest that you just set the private `mori._symmetry = self._symmetry` attribute, to avoid going via `Misorientation.set_symmetry()`. This should be safe to do, since `transpose()` should never change the crystal symmetry.

Also, the `Misorientation` class can assume that `Misorientation._symmetry` is a valid `Symmetry` instance, so these checks are unnecessary.

What do you think, @harripj?

harripj: Makes sense! I will update this in the next commit.

hakonanes: Looking better! What I meant was to remove *all* the checks, and just do

```python
def transpose(self, *axes):
    mori = super().transpose(*axes)
    mori._symmetry = self._symmetry
    return mori
```

This way you dont need `Orientation.transpose()`, or the checks.

harripj: Ideal! Updated in latest commit.

hakonanes:
Note that coverage can be checked locally with the coverage.py package as explained in the [contributing guide](https://orix.readthedocs.io/en/latest/contributing.html#run-and-write-tests).

hakonanes (changes requested):
Great stuff @harripj, thank you! In my opinion this is nearly ready to be merged. @pc494 might have something to add.

I've made some final suggestions. In addition to these, I would very much appreciate it if you could:
1. Add a one-line explanation of the addition (e.g. the PR title) to the changelog (CHANGELOG.rst in top directory) under Unreleased/Added, at the bottom of the list (newest additions).
2. Update the API reference (which is not 100% auto generated) with the transpose method in the list of class methods under Orientation/auto_summary (alphabetically), and likewise for Misorientation, Rotation, Miller and Vector3d.

orix/tests/test_vector3d.py
```diff
@@ -316,6 +349,84 @@ def test_mean_xyz():
...
+def test_transpose_2d_data_shape(shape, expected_shape):
+    v1 = Vector3d(np.random.rand(*shape))
+    v2 = v1.transpose()
+
+    assert v2.data.shape == tuple(expected_shape)
+
+
+def test_transpose_3d_no_axes():
+    v1 = Vector3d(np.random.rand(5, 4, 2, 3))
+    try:
```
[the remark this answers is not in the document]
harripj: Also incorporated for test_transpose_3d_wrong_number_of_axes

hakonanes (approved):

hakonanes:
Again, thanks for this @harripj! I'd appreciate it if either @dnjohnstone or @pc494 could merge this, to get another pair of eyes.

pc494:
I'll review this and fix the conflict :)

pc494 (approved):
I am happy with this, will deal with the conflict and merge. Many thanks @harripj

Issue #218, opened 2021-08-05
Refactor the "random" method of Rotations


pc494:
It currently uses rejection based sampling, I would prefer the method described in:

https://mathworld.wolfram.com/HyperspherePointPicking.html

hakonanes:
I haven't used `Rotation.random()`, does it not produce the results you would like, is it inefficient, or both?

https://github.com/pyxem/orix/blob/05311f6a5623453a8f50a9308066d2a3e66f6d29/orix/quaternion/rotation.py#L531-L547

I'm not entierly sure how to implement the equations in the link, but I guess you do. The hypersphere point picking should be added as an alternative method to the existing one, I assume.

pc494:
The current version is correct, it's doing the Marsaglia (1972) method from:

https://mathworld.wolfram.com/SpherePointPicking.html

which gives the correct answer. However it's slower than the version I'll replace it with because it involves discarding entries (also it's a `while` loop).

Consider this nothing more than house keeping that came up while I was working with some of these bits of the code. I'll implement the fix in my big bundle of things.

argerlt:
I haven't done a PR in a bit, so I'm practicing with an easier one first.

Implementation of this was easy enough, it's a change in `orix\_base.py` from this:
```
    @classmethod
    def random(cls, shape: Union[int, tuple] = 1) -> Object3d:
        n = int(np.prod(shape))
        obj = []
        while len(obj) < n:
            r = np.random.uniform(-1, 1, (3 * n, cls.dim))
            r2 = np.sum(np.square(r), axis=1)
            r = r[np.logical_and(1e-9**2 < r2, r2 <= 1)]
            obj += list(r)
        obj = cls(np.array(obj[:n]))
        obj = obj.unit
        obj = obj.reshape(shape)
        return obj
```
to this 
```
    @classmethod
    def random(cls, shape: Union[int, tuple] = 1) -> Object3d:
        n = int(np.prod(shape))
        obj = np.random.normal(size = [n, cls.dim])
        obj = cls(np.array(obj[:n]))
        obj = obj.unit
        obj = obj.reshape(shape)
        return obj
```

Then I did some speed tests to ensure it was, in fact, faster
```
# speed testing for a few different sizes:
for n in np.array([500,5000,50000,500000]):
    tic = time.time()
    for i in range(5):
        r_new = Rotation.random_new(n)
    toc_new = time.time()-tic
    tic = time.time()
    for i in range(5):
[...]

Issue #216, opened 2021-07-21
Generalising the .scatter() method


pc494:
The incoming `.scatter()` method can be moved up one level in the inheritance chain (see https://github.com/pyxem/orix/pull/202#issuecomment-884286379). This issue much like #215 in that is really a reminder for me when I get the chance to write new code.

pc494:
closed by #222

Issue #215, opened 2021-07-19
Streamline the from_euler method to better work with MTEX


pc494:
As discussed https://github.com/pyxem/orix/pull/201#issuecomment-880784662 in I've recently discovered how MTEX's convention differs from Bunge's. Currently we use the (undocumented) convention keyword '`Krakow_Hielschler`' to deal with this. But testing suggests we could simply reroute that through direction kwarg.

This would simplify the API somewhat and is also a lot easier to document.

hakonanes:
great. would this be a patch or minor release change?

pc494:
it's patch level (non-breaking and mainly documentation), but I think realistically it won't happen until I've got a fair bit of time to code (October or so) at which point I'll try and push all of the minor release elements through as well. I've slightly gone off patching except for urgent cases.

PR #221, opened 2021-09-04
Uniform sampling of orientation space using cubochoric coordinates

### LINKED ISSUE #219: We should start to use Numba
After optimizing generation of a dictionary of simulated EBSD patterns from a master pattern in kikuchipy ([code in open PR](https://github.com/pyxem/kikuchipy/pull/405/files#diff-600ddf9a30919bf272396e055cbbac22ce79d8b443ce705524e7d3e22e29a14fR19)), I'm starting to think we should start to use Numba in orix. This can sometimes mean moving away from always using vectorized operations, and instead using a for loop in a function decorated with Numba.

I've for example written a [Numba version](https://github.com/pyxem/kikuchipy/pull/405/files#diff-600ddf9a30919bf272396e055cbbac22ce79d8b443ce705524e7d3e22e29a14fR289) of the orix quaternion multiplication of one quaternion with many vectors, as part of the mentioned PR. That gave me a huge speed up over doing it vectorized.

If we find the time, I suggest we start small by providing Numba versions of calculations like this, and move on from there. How this would be available via the public API, I don't have a suggestion for yet.

hakonanes:
This PR adds sampling of a point group's Rodrigues Fundamental Zone (RFZ) using cubochoric coordinates as an alternative method in the `orix.sampling.get_sample_fundamental()` function. The sampling is presented in a paper by [Singh and De Graef (2016)](dx.doi.org/10.1088/0965-0393/24/8/085013) and is available in EMsoft's `EMsampleRFZ` command line program.

I've checked the quaternions returned in this PR's implementation to the quaternions from `EMsampleRFZ` for all proper point groups (1, 2, 222, 4, 422, 3, 32, 6, 622, 23, 432) for *N* from 10 to 100 in steps of 10 (*N* is the number of steps of the semi-length of the cubochoric cell), and get identical quaternions for all samplings apart from for 2 and 32. For these two, I get the same number of quaternions, but they are slightly different. Plotting the RFZ it seems like either (1) orix' `OrientationRegion` is defined differently than in EMsoft, or (2) the EMsoft sampling for these two point groups aren't correct... Because plotting the orientations returned from EMsoft within the RFZ defined in orix shows that there are parts of the zone not filled by EMsoft, while they are by the implementation in this PR. I've looked at the implementation in EMsoft, and the only difference is in the final selection of which rotations calculated from cubochoric coordinates are inside the RFZ: In EMsoft, this selection is done with Rodrigues vectors before final conversion to quaternions, while in orix this selection is done after the final conversion to quaternions.

Suggestions (would like your input on, @pc494):
1. I've used the sampling from EMsoft with great success previously. I suggest the cubochoric sampling becomes the default sampling in `get_sample_fundamental()`.
2. I want to use this sampling in the kikuchipy documentation, and thus would like to release v0.7 not long after this is in, hopefully within next weekend. Is this possible from your end? I can do the release if you don't have time or don't want to.

I've also allowed keyword arguments to be passed to `get_sample_fundamental()`, which are passed on to the selected sampling method. This way, instead of passing the "resolution", users of the cubochoric sampling can pass *N* as in EMsoft (see explanation of *N* above).

The sampling is implemented using Numba, so that package is added to the dependencies. With Numba we get quite nice speeds with regular for loops (which are much simpler than trying to do it vectorized!), but I couldn't find a good way to parallelize the loops... For all thinkable use cases, it shouldn't be necessary to go any faster than with one core (see timing example under Examples).

Many of the conversions in the [Rowenhorst et al. (2015)](http://dx.doi.org/10.1088/0965-0393/23/8/083501) are needed in the sampling and are thus implemented. I might move these to their own private module in `orix.quaternion`, where they belong.

Close #219.

#### Progress of the PR
- [x] Orientation sampling notebook showing the orientations produced from different sampling methods in RFZs and rotations of the Z vector in pole figures
- [x] Update API reference for `get_sample_fundamental()`.
- [x] Changelog update

Sampling the RFZ for point group *432* with a "resolution" of about 1.5 degrees (takes ~18 s on my laptop using 1 Intel i7-7600 core, and almost no RAM)
```python
>>> pg = quaternion.symmetry.O
>>> rot = sampling.get_sample_fundamental(method="cubochoric", semi_edge_steps=100, point_group=pg)
>>> ori = quaternion.Orientation(rot).set_symmetry(pg)
>>> ori
Orientation (333227,) 432
[[ 0.8562 -0.3474 -0.3474 -0.1595]
 [ 0.8562 -0.3491 -0.3491 -0.1519]
 [ 0.8562 -0.3508 -0.3508 -0.1442]
 ...
 [ 0.8562  0.3508  0.3508  0.1442]
 [ 0.8562  0.3491  0.3491  0.1519]
 [ 0.8562  0.3474  0.3474  0.1595]]
>>> ori.get_random_sample(10000).scatter()
```

cubochoric_sampling_432_rfz

#### For reviewers


pc494:
Broadly:

This is a good addition to the code base, I'm happy to review it, once we've looped round and got it merged we can set to work on a release version of `orix` (it's about time) - if we miss some stuff for lack of time (my end), eg. #215,#216,#218 that's okay.

Just a note, I think this is going to be slower than the 'quaternion' method for sampling small fundamental zones. Not really sure that's relevant, but thought it was worth mentioning.

Expect a first review back from me by Monday.

hakonanes:
> I'm happy to review it, once we've looped round and got it merged we can set to work on a release version of `orix` (it's about time) - if we miss some stuff for lack of time (my end), eg. #215,#216,#218 that's okay.

Great, thanks! Yeah, that sounds good. I think the IPF functionality has to wait for v0.8.

> Just a note, I think this is going to be slower than the 'quaternion' method for sampling small fundamental zones. Not really sure that's relevant, but thought it was worth mentioning.

You're right in that it takes longer. I just checked now, without a resolution of 1.5 degrees for point group *432* "cubochoric" took 19.3 s and produced 251 249 quaternions, while "quaternion" took 9 s and produced 290 847 quaternions. However, "quaternion" used 2.8 GB of RAM while "cubochoric" used 0.3 GB of RAM.

> Expect a first review back from me by Monday.

Thanks, but I would appreciate it if you could wait until I've removed the "Draft" labeling of this PR and marked it "Ready for review". I hope to have it ready by Monday.

pc494:
> > I'm happy to review it, once we've looped round and got it merged we can set to work on a release version of `orix` (it's about time) - if we miss some stuff for lack of time (my end), eg. #215,#216,#218 that's okay.
> 
> Great, thanks! Yeah, that sounds good. I think the IPF functionality has to wait for v0.8.
> 
> > Just a note, I think this is going to be slower than the 'quaternion' method for sampling small fundamental zones. Not really sure that's relevant, but thought it was worth mentioning.
> 
> You're right in that it takes longer. I just checked now, with a resolution of 1.5 degrees for point group _432_ "cubochoric" took 19.3 s and produced 251 249 quaternions, while "quaternion" took 9 s and produced 290 847 quaternions. However, "quaternion" used 2.8 GB of RAM while "cubochoric" used 0.3 GB of RAM.
>
Ah great, that means we've introduced more flexibility for users.
 
> > Expect a first review back from me by Monday.
> 
> Thanks, but I would appreciate it if you could wait until I've removed the "Draft" labeling of this PR and marked it "Ready for review". I hope to have it ready by Monday.

Ooops, my bad, would have noticed that when I started trying to review it, just ping me when you're ready.

hakonanes:
I'm working on getting 100% coverage and a short notebook titled "Orientation sampling" to be included in the user guide. The notebook will show the fundamental zone for some point groups from the side and top, and perhaps also Z vectors in the stereographic projection rotated by the orientations.

pc494:
Yup, would just change it directly, annoying but it is unlikely to cause anyone huge issues.

hakonanes:
I've added a user guide comparing the three sampling methods. These results clearly show that cubochoric sampling produces the most uniform sampling with the fewest samples. I've therefore set this method as the default one in `get_sample_fundamental()` and `get_sample_local()`. Any input on the wording or presentation in the notebook is appreciated.

hakonanes:
The user guide notebook added: https://orix--221.org.readthedocs.build/en/221/uniform_sampling_of_orientation_space.html

pc494 (commented):
I'm happy with the code, I think the final block of the documentation isn't true though

> Based on these plots we can say that the Cubochoric sampling method covers orientation space uniformly, while the Haar Euler and Quaternion methods have trouble sampling close to the identity orientation and an orientation located perpendicular to this orientation in the RFZ, respectively.

All three samples should be uniform. Possible explanations:

1) Bug in the other two samplings
2) Bug/problem with the way plotting works when one takes subsamples
3) Visualizing in the RF zone gives confusing results because orientation spaces are weird.

Any thoughts?

hakonanes:
> Any thoughts?

I think it must be a bug, as you say. We've use the "haar_euler" method in the kikuchipy documentation, and @friedkitteh found in his master thesis that using that sampling method matched poorly with grains close to the identity orientation. So based on this experience I'd say it samples poorly close to the identity orientation.

I did a quick check now, where I calculated the disorientation (smallest misorientation under symmetry) using `Orientation.angle_with()` to `Orientation.identity()`, and got this:

```python
>>> import numpy as np
>>> from orix.sampling import get_sample_fundamental
>>> pg = symmetry.O
>>> res = 2
>>> ori_identity = Orientation.identity().set_symmetry(pg)
>>> rot_euler = get_sample_fundamental(method="haar_euler", point_group=pg, resolution=res)
>>> ori_euler = Orientation(rot_euler).set_symmetry(pg)
>>> mori_euler = np.degrees(ori_euler.angle_with(ori_identity).data)
>>> rot_quat = get_sample_fundamental(method="quaternion", point_group=pg, resolution=res)
>>> ori_quat = Orientation(rot_quat).set_symmetry(pg)
>>> mori_quat = np.degrees(ori_quat.angle_with(ori_identity).data)
>>> bins = np.linspace(0, 62, 100)
>>> fig, ax = plt.subplots(figsize=(9, 5))
>>> _ = ax.hist(mori_euler, bins, label="haar_euler", alpha=0.5)
>>> _ = ax.hist(mori_quat, bins, label="quaternion", alpha=0.5)
>>> ax.set_xlabel(r"Misorientation [$^{\circ}$]")
>>> ax.set_ylabel("Frequency")
>>> ax.legend()
>>> fig.savefig("/home/hakon/kode/orix_test/gridding/haar_measure_identity1.png")
```

haar_measure_identity1

After including the cubochoric one:

sampling_close_to_identity

hakonanes:
Should I drop the two other methods in the user guide then, and just present the cubochoric sampling?

pc494:
No, I think leave the notebook as is, I would just drop the final paragraph; users can see what the deal is from the plots. Given that this PR doesn't introduce problems (only identifies them) I'll be happy to merge early tomorrow. 

NB: I ran some brief checks of my own, I'm pretty sure it's a bug/oddity somewhere with the existing sampling methods, but will get this merged, use the new code to confirm the finding and raise a suitable warning/issue. This will help us keep to the release schedule.

hakonanes:
The .gitignore entry of .ipynb files is a problem...

hakonanes:
Scratch that, the push went OK.

hakonanes:
The updated notebook is here: https://orix--221.org.readthedocs.build/en/221/uniform_sampling_of_orientation_space.html

pc494:
> The .gitignore entry of .ipynb files is a problem...

did you end up using the `-f` flag?

hakonanes:
> did you end up using the `-f` flag?

Hehe, yup. It's just that we don't have .ipynb in the kikuchipy .gitignore, so I only add it when pushing orix stuff, which has been a while since I've done.

PR #223, opened 2021-09-05
Simplify the Rotation.from_euler() API

### LINKED ISSUE #215: Streamline the from_euler method to better work with MTEX

pc494:
See discussion in #215

#### Progress of the PR

#### For reviewers

close #215

### DIFF
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -335,7 +335,7 @@ class Rotation(Quaternion):
         direction : str
             "lab2crystal" or "crystal2lab", ignored if "MTEX" convention is in use
         """
-        conventions = ["bunge","MTEX","Krakow_Hielscher"]
+        conventions = ["bunge", "MTEX", "Krakow_Hielscher"]
 
         if convention not in conventions:
             raise ValueError(
@@ -367,7 +367,6 @@ class Rotation(Quaternion):
             # see - https://mtex-toolbox.github.io/MTEXvsBungeConvention.html
             # and orix issue #215
             direction = "lab2crystal"
-
 
         # Uses A.5 & A.6 from Modelling Simul. Mater. Sci. Eng. 23
         # (2015) 083501
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -335,17 +335,37 @@ class Rotation(Quaternion):
         direction : str
             "lab2crystal" or "crystal2lab", ignored if "MTEX" convention is in use
         """
-        conventions = ["bunge", "MTEX", "Krakow_Hielscher"]
+        directions = ["lab2crystal", "crystal2lab"]
+        conventions = ["bunge", "mtex"]
+
+        # processing directions
+        if direction not in directions:
+            raise ValueError(
+                f"The chosen direction is not one of the allowed options {directions}"
+            )
+
+        # processing the convention chosen
+
+        convention = convention.lower()
+
+        if convention == "krakow_hielscher":
+            # To be applied to the data found at:
+            # https://www.repository.cam.ac.uk/handle/1810/263510
+            warnings.warn(
+                "This method is deprecated and will be removed in v0.8 use 'MTEX' instead"
+            )
+            convention = "mtex"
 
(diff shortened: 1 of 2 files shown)


orix/quaternion/rotation.py
```diff
@@ -358,49 +359,40 @@ def from_euler(cls, euler, convention="bunge", direction="crystal2lab"):
...
-
-            data = Quaternion(q)
-
-            if direction == "lab2crystal":
-                data = ~data
-
-            rot = cls(data.data)
-            rot.improper = np.zeros((n))
-            return rot
+            warnings.warn("This method is deprecated, use 'MTEX' instead")
```
hakonanes, line 362: "This convention is deprecated, use 'MTEX' instead".

Since it is bunge in lowercase and MTEX in uppercase, could we support mtex as well? As in `if convention.lower() == "mtex":`. I believe I've asked this question before, and you turned it down, but I can't remember why, and I think both "bunge", "BUNGE" and "Bunge" should be accepted.

orix/quaternion/rotation.py
```diff
@@ -358,49 +359,40 @@ def from_euler(cls, euler, convention="bunge", direction="crystal2lab"):
...
+            convention = "MTEX"
+
+        if convention == "MTEX":
+            # MTEX uses bunge but with lab2crystal referencing:
+            # see - https://mtex-toolbox.github.io/MTEXvsBungeConvention.html
+            # and orix issue #215
+            direction = "lab2crystal"
```
hakonanes, line 369: Which tests have you done to ensure that this is correct? I did a simple check.

MTEX: 
```matlab
>> startup_mtex
>> plotx2east
>> plotzOutOfPlane
>> rot = rotation.byEuler([30, 60, 70] * degree)
>> disp([rot.a rot.b rot.c rot.d])
   -0.5567   -0.4698    0.1710   -0.6634
>> figure, plot(rot * vector3d.Z) % Produces a pole figure
```

orix:
```python
>>> from orix import quaternion, vector
>>> rot = quaternion.Rotation.from_euler(np.radians([30, 60, 70]), direction="lab2crystal")
>>> rot
Rotation (1,)
[[ 0.5567  0.4698 -0.171   0.6634]]
>>> (rot * vector.Vector3d.zvector()).scatter()  # Produces the same pole figure
```

I'm satisfied with this result.

hakonanes (approved):
Great that we can remove code but not functionality! I just have a couple questions I would like answered if you find the time.

orix/quaternion/rotation.py
pc494: MTEX is (historically) an acronym which is why I capitalised it. I think words (even proper nouns) should be lower case. Generally, I don't like string checking, it's developer clutter for an extremely marginal increase in usability. Compromise might be to run `.lower()` at the top?

orix/quaternion/rotation.py
pc494: I ran the old implementation against the new one with random rotations (where #218 came from as it happens) and got an agreement. I also replicated a figure produced in MTEX in orix (fundamental zone plotting). It's also consistent with what we've got written down, MTEX should only differ from 'bunge' by the frame change.

orix/quaternion/rotation.py
hakonanes: > Compromise might be to run `.lower()` at the top?

That is *exactly* what I'm after :)

orix/quaternion/rotation.py
hakonanes: Thanks, good to know, I agree.

orix/quaternion/rotation.py
pc494: okay, I'm going to leave the documentation as is, but I'll implement the change

Issue #227, opened 2021-09-08
On the use of data_dim


harripj:
Just wanted to open a discussion about navigation dimensions and access.

Personally I have found myself trying to call the attribute ```.ndim``` on almost all instances (```Orientation```, ```Vector3d```) to get the number of navigation dimensions (not the full data ndim as last dimension ```dim``` is hidden from the user for the most part) and have been considering adding this to ```Object3d```.

I have found that ```.data_dim``` produces this result, but personally, coming from NumPy, I was searching for ```.ndim``` in the docs. I am also not sure whether ```.data_dim``` is the clearest attribute name for this, as its doctsring doesn't reference that it provides access to the number of navigation dimensions.

I would like to get other peoples feelings on this, and whether ```.data_dim``` is well used within your workflows?

I would have a couple the suggestion to add ```.ndim``` attribute to ```Object3d``` and either:

1. maintain alongside ```.data_dim```
2. replace ```.data_dim```.

My feeling is that ```.ndim``` is natural for those familiar with NumPy. The drawback to this would be that ```Object3d.data.ndim``` and ```Object3d.ndim``` give different values, but this is also the case with ```.shape```, so I think it is consistent behaviour.

In any case I think the docstring for both attributes should describe that they provide access to just the navigation dimensions.

hakonanes:
> I was searching for `.ndim` in the docs.

The API reference aren't perfect, and they have great room for improvement in my opinion. I'm slowly updating docstrings in the parts of the code I touch in PRs. Please feel free to make PRs with just docstring updates if you find something lacking and you find the time.

> I would like to get other peoples feelings on this, and whether `.data_dim` is well used within your workflows?

I always use `len(rotations.shape)` to get number of dimensions ([example use in kikuchipy](https://github.com/pyxem/kikuchipy/blob/develop/kikuchipy/generators/ebsd_simulation_generator.py#L87)), but that is because I haven't considered using the `data_dim` property at all!

> I would have a couple the suggestion to add `.ndim` attribute to `Object3d` and either:
>     1. maintain alongside `.data_dim`
>     2. replace `.data_dim`.

[...]

PR #228, opened 2021-09-08
Deprecate Object3d.data_dim property in favour of Object3d.ndim

harripj:
As discussed in #227 we will aim for the addition of a the attribute ```ndim``` on the ```Object3d``` class to be more consistent with NumPy. This will lead to the eventual deprecation of ```data_dim```.

The docstring example of ```data_dim``` has been updated for clarity.

#### Progress of the PR

```python
>>> v = Orientation.random_vonmises((100, 200))
>>> v.ndim
>>> 2
```

#### For reviewers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,11 @@ this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.htm
 Unreleased
 ==========
 
+Added
+-----
+- ndim attribute to Object3d which returns number of navigation dimensions
+- data_dim attribute of Object3d to be deprecated in 0.9.0 with a warning from 0.8.0
+
 
 2021-09-07 - version 0.7.0
 ==========================
--- a/orix/base/__init__.py
+++ b/orix/base/__init__.py
@@ -111,12 +111,7 @@ class Object3d:
         return self.data.shape[:-1]
 
     @property
-    @deprecated(
-        "0.8.0",
-        message="The use of 'data_dim' is deprecated.",
-        alternative="ndim",
-        removal="0.9.0",
-    )
+    @deprecated(since="0.8", alternative="ndim", removal="0.9", object_type="property")
     def data_dim(self):
         """int : The dimensions of the data.
 
--- a/orix/base/__init__.py
+++ b/orix/base/__init__.py
@@ -112,15 +112,15 @@ class Object3d:
 
     @property
     @deprecated(
-        "0.7.1",
+        "0.8.0",
         message="The use of 'data_dim' is deprecated.",
         alternative="ndim",
-        removal="0.8.0",
+        removal="0.9.0",
     )
     def data_dim(self):
         """int : The dimensions of the data.
 
-        For example, if `data` has shape (4, 4, 3), `data_dim` is 3.
+        For example, if `data` has shape (4, 5, 6), `data_dim` is 3.
 
         """
         return self.ndim
--- a/orix/base/__init__.py
+++ b/orix/base/__init__.py
@@ -17,6 +17,7 @@
 # along with orix.  If not, see <http://www.gnu.org/licenses/>.
 
 import numpy as np
+from .._util import deprecated
 
 
 # Lists what will be imported when calling "from orix.base import *"
@@ -110,12 +111,23 @@ class Object3d:
         return self.data.shape[:-1]
 
     @property
+    @deprecated(
+        "0.7.1",
+        message="The use of 'data_dim' is deprecated.",
+        alternative="ndim",
+        removal="0.8.0",
+    )
     def data_dim(self):
         """int : The dimensions of the data.
 
         For example, if `data` has shape (4, 4, 3), `data_dim` is 3.
 
         """
+        return self.ndim
+
+    @property
+    def ndim(self):
+        """int: The number of navigation dimensions of the object."""
         return len(self.shape)
 
     @property
--- a/orix/tests/test_object3d.py
+++ b/orix/tests/test_object3d.py
@@ -158,6 +158,10 @@ def test_data_dim(object3d):
     assert object3d.data_dim == len(object3d.data.shape[:-1])
 
 
+def test_ndim(object3d):
+    assert object3d.ndim == len(object3d.data.shape[:-1])
+
+
 def test_size(object3d):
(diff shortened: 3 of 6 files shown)


hakonanes:
Could you pull the latest changes from `master` and rebase on that? Should be `git rebase <latest-commit-hash-from-master-log>`.

hakonanes:
Or just merge master into your branch, as you just did now.

hakonanes:
Looks like you haven't configured pre-commit? It's explained in [the contributing guide](https://orix.readthedocs.io/en/stable/contributing.html#code-style)). This way, black is run before commiting, so that only blacked code is commited.

hakonanes:
I'll review after all checks pass.

orix/base/__init__.py
```diff
@@ -17,6 +17,7 @@
 # along with orix.  If not, see <http://www.gnu.org/licenses/>.
 
 import numpy as np
+from .._util import deprecated
```
hakonanes, line 20: I think it's best to stay with complete imports `from orix._util import deprecated`. We don't use relative imports anywhere else.

hakonanes: Also, as per [the contributing guide](https://orix.readthedocs.io/en/stable/contributing.html#code-style), we distinguish between stanard library imports, 3rd party imports, and orix imports by a blank line, as in

```python
import numpy as np

from orix._util import deprecated
```

orix/base/__init__.py
```diff
@@ -110,12 +111,23 @@ def shape(self):
         return self.data.shape[:-1]
 
     @property
+    @deprecated(
+        "0.8.0",
+        message="The use of 'data_dim' is deprecated.",
```
hakonanes, line 116: The `message` field isn't actually used. If it's OK with you, I'll remove it from the `deprecated` class and relevant tests and push to your branch. The message would now be (see [the class](https://github.com/pyxem/orix/blob/master/orix/_util.py#L69)): "Function `data_dim()` is depreated and will be removed in version 0.9.0. Use `ndim()` instead".

orix/base/__init__.py
```diff
@@ -110,12 +111,23 @@ def shape(self):
...
 
-        For example, if `data` has shape (4, 4, 3), `data_dim` is 3.
+        For example, if `data` has shape (4, 5, 6), `data_dim` is 3.
 
         """
+        return self.ndim
+
+    @property
+    def ndim(self):
+        """int: The number of navigation dimensions of the object."""
```
hakonanes, line 130: Add the example from the `data_dim` docstring here as well.

Also, I would actually say "... navigation dimensions of the instance.", because it can vary from instance to instance of the object.

orix/tests/test_object3d.py
```diff
@@ -158,6 +158,10 @@ def test_data_dim(object3d):
     assert object3d.data_dim == len(object3d.data.shape[:-1])
 
 
+def test_ndim(object3d):
```
hakonanes, line 161: Could you add a test to ensure that the warning is raised as expected? See [this test in kikuchipy](https://github.com/pyxem/kikuchipy/blob/v0.4.0/kikuchipy/signals/tests/test_ebsd.py#L1246) as a reference.

hakonanes (changes requested):

orix/base/__init__.py
harripj: Fine with me!

CHANGELOG.rst
```diff
@@ -10,6 +10,11 @@ this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.htm
 Unreleased
 ==========
 
+Added
+-----
+- ndim attribute to Object3d which returns number of navigation dimensions
+- data_dim attribute of Object3d to be deprecated in 0.9.0 with a warning from 0.8.0
```
hakonanes, line 16: Move the deprecation notice to its own "Deprecated" header below "Added" as per [Keep a Changelog](https://keepachangelog.com/en/1.1.0/#how), that we follow.

hakonanes (changes requested):

CHANGELOG.rst
```diff
@@ -12,8 +12,11 @@ Unreleased
 
 Added
 -----
-- ndim attribute to Object3d which returns number of navigation dimensions
-- data_dim attribute of Object3d to be deprecated in 0.9.0 with a warning from 0.8.0
+- "ndim" attribute to Object3d which returns number of navigation dimensions
+
+Deprecated
+----------
+- "data_dim" attribute of Object3d to be deprecated in 0.9.0 with a warning from 0.8.0
```
hakonanes, line 19: It's deprecated from 0.8.0, and will be removed in 0.9.0.

hakonanes (changes requested):

hakonanes:
So, I pulled your changes and found out that:
1. By placing the `@deprecated` decorator below the `@property` decorator made `deprecated` think `data_dim` was a function, and thus the warning is displayed correctly.
2. If `@deprecated` is placed above `@property`, an error is thrown because we cannot access `Object3d.data_dim.__name__`, for example, among other issues.

I consider 1. a hack, while 2. isn't an option, at least I cannot find a solution to wrap a property on the web. I'm comfortable with the hack, so I'll just provide a parameter to `@deprecated()` to specify whether it's a function or property.

harripj:
Yes, I found exactly the same and thought about implementing exactly what you have described, Happy for you to go ahead with updating the ```deprecated``` function in the way you suggested if you can! Based on the value of ```is_property``` (or similar) argument for this function it would make sense to remove (or not) the brackets shown in the warning.

orix/base/__init__.py
harripj: Updated.

hakonanes:
Almost done with this, thanks for the patience.

hakonanes:
Didn't find a way to add my commit on top of yours... Bad gitting, I assume. Anyways, I added the test to ensure that an expected warning was raised from accessing `Object3d.data_dim`. So this should be mergable now, I think.

hakonanes:
I forgot to remove the old test for the warning, I'll push that change now.

hakonanes (approved):
Since I contributed to this PR, we'll have to wait for another maintainer to have a look before merging.

pc494 (approved):
Good, sensible clean up - I am happy with the code as is, merging.

Issue #217, opened 2021-07-27
Interactive unit cell display


hakonanes:
After discussions with @IMBalENce, we agree that an interactive unit cell display, e.g. by changing three Euler angles, would be nice both for analysis of orientations, or for pedagogical illustrations.

We can track any interest and progress towards implementing this in this issue.

Some examples from Matplotlib that might be useful:
* https://matplotlib.org/stable/gallery/mplot3d/wire3d.html#sphx-glr-gallery-mplot3d-wire3d-py
* https://matplotlib.org/stable/gallery/mplot3d/voxels_numpy_logo.html#sphx-glr-gallery-mplot3d-voxels-numpy-logo-py

Note that we also have wireframe plotting of the Rodrigues fundamental zone and axis angle space: https://github.com/pyxem/orix/blob/master/orix/plot/rotation_plot.py.

hakonanes:
We can close this once #234 is merged.

PR #234, opened 2021-10-15
Simple plotting of unit cells

hakonanes:

* Plot quick and dirty single rotated unit cell with respect to the sample reference frame

Address parts of #217.

I needed this yesterday, found a quick solution on [the internet](https://stackoverflow.com/a/18790732), and added use of orix' vectors and rotations. Don't have any immediate plans to finish this myself: if you have a use for it, please contribute to this PR!

#### Progress of the PR
- [x] Attach crystal reference frame
- [x] Attach sample reference frame
- [n/a] ~Add new simple user guide notebook explaning the orientation definition in orix using this unit cell and IPF~ Left for a later PR when #249 is concluded.
- [x] Add support for orthorhombic unit cell plotting 

Secondary goals for another PR:
- [n/a] ~Provide all *basic* unit cells (not just cubic)~
- [n/a] ~Account for lattice parameters in non-cubic unit cells~

```python
>>> import numpy as np
>>> ori = Orientation.from_euler(np.radians(((0, 0, 0), (45, 0, 0))), symmetry=symmetry.Oh)
>>> ori[0].plot_unit_cell(c="b")
```

See https://github.com/pyxem/orix/pull/234#issuecomment-986977558 for how the unit cells plots look.

#### For reviewers


harripj:
I like this idea and actually it is very intuitive for the user to visualize what is going on. What I think is important for this is to show the reference frame correctly, ie. the inclusion of a small xyz axis somewhere on the plot. As in the case above the lattice points are being rotated, this would mean that these axes can remain fixed (for now).

It would also be nice to add `Structure` in here somewhere to change the visualization of the unit cell, but for now cubic is a good starting point.

I also think that the default projection of the unit cell should be consistent, from my understanding of the reference frames here this would mean that `+z` is out-of-page (/screen) by default, am I right in thinking this?

hakonanes:
> What I think is important for this is to show the reference frame correctly, ie. the inclusion of a small xyz axis somewhere on the plot.

Yes, there should be connected XYZ axes colored RGB that rotates with the unit cell here in this case, since we plot a fixed orientation. Adding a crystal reference frame attached to the crystal should be even easier. (There are TODO points for both of these tasks.)

> As in the case above the lattice points are being rotated, this would mean that these axes can remain fixed (for now).

We could create a simple widget providing the user with possibilities to interactively explore how the cell is oriented given a Euler angle triplet, axis/angle representation, or a quaternion. I assume this is what you're thinking of with "for now"?

> It would also be nice to add `Structure` in here somewhere to change the visualization of the unit cell, but for now cubic is a good starting point.

We could use diffpy.structure.Lattice here (which is stored in a `orix.crystal_map.Phase`'s `structure.lattice` attribute). I would keep this as simple as possible, though, because there are great existing tools for drawing unit cells in Python.

> I also think that the default projection of the unit cell should be consistent, from my understanding of the reference frames here this would mean that `+z` is out-of-page (/screen) by default, am I right in thinking this?

I agree. This is part of a broader discussion about default reference frame orientation, which we haven't had in orix yet. MTEX uses the map axes x-east, y-south, z-into plane throughout, which makes sense. We can adopt this, but then we should consider to change the stereographic projection axes as well (change from y-north and z-out of plane).

harripj:
> We could create a simple widget providing the user with possibilities to interactively explore how the cell is oriented given a Euler angle triplet, axis/angle representation, or a quaternion. I assume this is what you're thinking of with "for now"?

This would be a nice idea. What I meant was whether `Orientation` describes a rotation of the lattice or the `lab reference frame. At the moment in the PR the lattice is rotating (makes sense to me), but it could be the other way around.

> I agree. This is part of a broader discussion about default reference frame orientation, which we haven't had in orix yet. MTEX uses the map axes x-east, y-south, z-into plane throughout, which makes sense. We can adopt this, but then we should consider to change the stereographic projection axes as well (change from y-north and z-out of plane).

The current reference frame makes sense to me intuitively, but I agree this should be defined more concretely either way.

@hakonanes if you don't mind I can fix the default projection and add the reference frame axes by pushing to this PR?

hakonanes:
Just rebased this PR on the latest commit in `master`. Feel free to commit to my branch directly (you should be able to do this since you have write access to this repo). Your suggestions of what's needed to complete this PR is good, feel free to update the top comment as well (the todo list, examples etc.). I moved accounting for lattice parameters and different unit cells to a separete list of "secondary goals", as I think that can prove challenging, and for now a quick and dirty unit cell is something to build upon.

hakonanes:
I think we should finally add a short user guide notebook explaining and showing the definition of an orientation in orix, and perhaps the different options available in `Orientation.from_euler()`, by plotting unit cells and IPFs (and perhaps PFs). I added this as a thing to do in this PR. What do you think @harripj?

harripj:
> I think we should finally add a short user guide notebook explaining and showing the definition of an orientation in orix, and perhaps the different options available in Orientation.from_euler(), by plotting unit cells and IPFs (and perhaps PFs). I added this as a thing to do in this PR. What do you think @harripj?

This sounds like an excellent idea and very much needed!

hakonanes:
Great! For now, we can add the notebook to the top of the list in `index.rst`. ".ipynb" files are listed in ".gitignore", so we have to `git add -f notebook.ipynb` before committing (see [contribution guide](https://orix.readthedocs.io/en/stable/contributing.html#build-and-write-documentation)).

hakonanes:
> What I meant was whether `Orientation` describes a rotation of the lattice or the lab reference frame.  At the moment in the PR the lattice is rotating (makes sense to me), but it could be the other way around.

orix uses the Bunge convention for Euler angles and the definition of an `Orientation` (lab/sample to crystal), which is the opposite of MTEX. But we should still display the crystal rotated, with the sample reference frame fixed. Is that what you mean? If so, I agree.

harripj:
I have added the crystal and lab reference frames to the plots, perhaps these should be subscripted `_L` and `_C` to properly distinguish them in the future. The colour scheme for the reference frame axes comes from ParaView, but it strikes me that the yellow doesn't stand out well on the white plot background. These axes are also arrows and can be given arrowheads by setting the `arrowstyle` argument in `plot_unit_cell`: https://matplotlib.org/stable/api/_as_gen/matplotlib.patches.FancyArrowPatch.html

However under the `+z` projection the `z` arrow head doesn't show quite right, it should be a circle if looking straight into a cone, so for now the default is `'-'`, ie. no arrow head.

For the two example plots above:
image
image

hakonanes:
This looks promising!

Could you make the arrows compatible with Matplotlib >= 3.5 (as per [this comment in the SO post](https://stackoverflow.com/questions/22867620/putting-arrowheads-on-vectors-in-matplotlibs-3d-plot/22867877#comment123731049_22867877))?

I suggest using some easily recognizable RGB colors for the axes, and yes some subscripts to distinguish sample/crystal would be an improvement.

Looking at the (45, 0, 0) unit cell, it seems to me like we have to apply the inverse rotation to the unit cell vectors to get the expected orientation if we use the Bunge definition?

hakonanes:
Another thing, is there a particular reason for not attaching the crystal coordinate system to the cube origin?

harripj:
> Looking at the (45, 0, 0) unit cell, it seems to me like we have to apply the inverse rotation to the unit cell vectors to get the expected orientation if we use the Bunge definition?

It looks like it, but this is somewhat confusing for me given https://github.com/pyxem/orix/pull/234#issuecomment-986941838.

> Another thing, is there a particular reason for not attaching the crystal coordinate system to the cube origin?

They are currently centered at `(0, 0, 0)` and the lattice spans from `-1..1`. I will update it to start at the lattice origin and see how it looks.

hakonanes:
When I interpret Bunge Euler angles, I start out with the two coordinate systems in parallel, and then rotate the crystal about Zc, Xc and then about Xc again. This is according to Bunge's definition in Figure 2.2 of [his book](http://www.lem3.fr/docs/Bunge_TextureAnalysis.pdf#page=13). Given this interpretation, the unit cell above should have orientation (-45, 0, 0), right?

hakonanes:
Not 100% sure, but it seems to me like `Orientation.from_euler(np.radians((45, 0, 0,)), direction="lab2crystal")` is what we want. Actually, passing `convention="mtex"` in provides the same result.

@pc494, does Bunge's figure in [the reference above](http://www.lem3.fr/docs/Bunge_TextureAnalysis.pdf#page=13) correspond to the definition of an orientation in orix?

harripj:
`direction="lab2crystal"` ah yes, good spot. This makes sense and I forgot about it in the examples above, too. For `from_euler` we must use `lab2crystal` as we are rotating the crystal in the lab reference frame (this is what I believe is done in the reference too).

hakonanes:
My understanding is that Bunge's definition is `lab2crystal`, not `crystal2lab`, which MTEX uses, and is the default in orix. This has confused me before... I actually think we should switch the default to `Rotation.from_euler(*, direction="lab2crystal")`. I'll make a PR stating this.

harripj:
> My understanding is that Bunge's definition is `lab2crystal`, not `crystal2lab`

I agree.

> I actually think we should switch the default to Rotation.from_euler(*, direction="lab2crystal"). I'll make a PR stating this.

Personally when using `Orientation.from_euler` I am always using the `lab2crystal` kwarg. I realise that MTEX uses the opposite convention. I am working on a notebook to demonstrate this as mentioned here, in any case I think that it should be made clear to the users.

hakonanes:
I'm glad that we agree. Looking forward to the notebook.

The issue in `Rotation.from_euler()` is that `convention="bunge"` corresponds to `direction="crystal2lab"` instead of "lab2crystal", which is correct. Now that we have removed the "krakow_hielscher" convention (necessary for the 2020 clustering paper of Phillip++), the `convention` parameter is redundant, as we can control the two possible behaviours with `direction`. We should remove either `convention` or `direction`. I think we should keep `direction`.

hakonanes:
Looking at the [Rowenhorst et al. (2015](http://dx.doi.org/10.1088/0965-0393/23/8/083501), I'm not so sure again. They use passive rotations, that is of reference frames, not objects. This is Bunge's definiton. Might it be that the naming "crystal2lab" should be changed to "lab2crystal", and vice versa?

hakonanes:
> Looking at the (45, 0, 0) unit cell, it seems to me like we have to apply the inverse rotation to the unit cell vectors to get the expected orientation if we use the Bunge definition?

Orientations in *orix* are coordinate transformations from the sample reference frame to the crystal reference frame, which follows Bunge. Therefore, we must apply the inverse rotation when rotating crystal coordinates into sample coordinates. The unit cell plot shows the orientation of the crystal reference frame with respect to the sample reference frame, so we must apply the inverse rotation here (on the crystal reference frame vectors as well as the vectors delineating the crystal).

Following this logic, I've created an "Orientations in orix" notebook to understand orientations in orix:
[PDF](https://github.com/pyxem/orix/files/7667688/orientations_in_orix.pdf). Note that I've applied the inverse rotation `rotation = ~rotation` in `unit_cell_plot.py` before plotting the unit cell.

What do you think, @harripj?

I find that  the notebook corresponds well with the following MTEX documentation:
* [MTEX vs. Bunge Convention](https://mtex-toolbox.github.io/MTEXvsBungeConvention.html)
* [Pole Figures](https://mtex-toolbox.github.io/OrientationPoleFigure.html)
* [Inverse Pole Figure](https://mtex-toolbox.github.io/OrientationInversePoleFigure.html)

harripj:
I like the notebook and am still processing this, but one thing that catches my eye is that we should be using `lab2crystal` when using `from_euler` to be consistent with Bunge. In the PDF I don't believe this is currently the case and this would seem to me to cancel out the need for the inversion in `unit_cell_plot.py`. What do you think @hakonanes?

hakonanes:
My understanding is that the direction arguments crystal2lab (default) and lab2crystal should be switched, that they mean the opposite of what they do now. With this change and the rotation inversion in cell plotting, I think everything is correct. I'd like to get @pc494's opinion on this as well.

hakonanes:
In `unit_cell_plot.py` we keep the sample reference frame stationary while rotating the crystal reference frame by `rotation`. The vectors to rotate are crystal directions (reference frame and vectors delineating the unit cell), hence we must apply the inverse rotation (`m-3m -> xyz`), not the rotation (`xyz -> m-3m`). This is the point I'm making in the notebook. On the other hand, when rotating sample directions to find out which crystal direction a certain sample direction points in, we apply the rotation. In MTEX it's the opposite, which aligns with their documentation stating that they do the opposite of what Bunge does.

As stated above, `Orientation.from_euler(..., direction="lab2crystal")` and `Orientation.from_euler(..., convention="mtex")` produce the same orientations. Therefore, we should switch `lab2crystal` and `crystal2lab`.

I have always found these conventions and operations tricky to pin down across softwares... However, remembering that applying the Bunge orientation to a sample direction returns the crystal direction, and applying the inverse Bunge orientation to a crystal direction to get the sample direction, helps me to understand when to use which. This definition and its use in orix is in line with the conversions and vector rotations in the tutorial paper by [Rowenhorst et al. (2015)](http://dx.doi.org/10.1088/0965-0393/23/8/083501).

Another thing we should mention in the notebook is that rotations in orix are passive, so if two rotations q_A and q_B should be done, first A, then B, then they must be combined q_A * q_B.

hakonanes:
> Another thing we should mention in the notebook is that rotations in orix are passive, so if two rotations q_A and q_B should be done, first A, then B, then they must be combined q_A * q_B.

Again, I'm confused. The following tells me rotations in orix are *active*:

```python
>>> import numpy as np
>>> rot1 = Rotation.from_euler(np.radians((90, 0, 0)))
>>> rot2 = Rotation.from_euler(np.radians((0, 60, 0)))
>>> rot3 = Rotation.from_euler(np.radians((90, 60, 0)))
>>> np.allclose(rot3.data, (rot1 * rot2).data)
False
>>> np.allclose(rot3.data, (rot2 * rot1).data)
True
```

This is *not* consistent with convention 3 in Rowenhorst et al. (2015) (reference above), but with MTEX (see [this part](https://mtex-toolbox.github.io/RotationOperations.html#4) of their documentation). More surprisingly, it *is* consistent with the results I obtain by using the functions [eu2qu()](https://github.com/drowenhorst-nrl/PyEBSDIndex/blob/main/pyebsdindex/rotations.py#L234) and [quat_multiply1()](https://github.com/drowenhorst-nrl/PyEBSDIndex/blob/main/pyebsdindex/rotlib.py#L1569) from the supplementary material to the Rowenhorst paper.

hakonanes:
Never mind, I don't think [nbsphinx](https://nbsphinx.readthedocs.io/en/0.8.7/code-cells.html#Interactive-Widgets-(HTML-only)) works with a Matplotlib widget.

harripj:
> @harripj, could you add your notebook to the notebook gallery in index.rst, so that it is included in the docs built by RtD in this PR?

Yes no problem, I will do it now. In fact I am still trying to get my head around all of this as we have discussed here. The notebook itself operates on the unit cell vertices _actively_. This may need to be addressed given the outcome of this discussion. For now the notebook should run and at least give some insight as to what is happening to the relative crystal orientation. It would be great to hear another voice on this subject, if @pc494 has some time?

For the notebook case if rotations in orix are passive, then the sample reference frame should be animated to rotate around a fixed crystal, instead of as it is now. This is not very intuitive to the user of course, so it leads to the interpretation that the global reference frame is attached to the sample reference frame? In that case a passive sample reference frame rotation must be equivalent to an active inverse rotation of the crystal. I think it is necessary to tidy this up in any case.

hakonanes:
Forgot to mention that `nbsphinx`/RtD doesn't build notebooks with cell output stored. If we want the interactive unit cell plotting in the documentation, the notebook must be uploaded without cell output. We should add a point to the [contributing guide](https://orix--234.org.readthedocs.build/en/234/contributing.html#build-and-write-documentation) about this!

hakonanes:
> The notebook itself operates on the unit cell vertices _actively_.

Using `convention="mtex"` (same result as "lab2crystal") gives the same quaternion as with MTEX' `orientation.byEuler()`, which describes a coordinate transformation from the crystal to the sample reference frame (inverse of Bunge's definition). This produces the expected unit cell when directly multiplying the unit cell vertices (which one would do in MTEX as well), because we find which sample direction the unit cell vertices point in. If we do `convention="bunge"`, we get a different quaternion from MTEX (Bunge's definition), which describes a coordinate transformation from the sample to the crystal reference frame, and we must thus multiply with the inverse rotation to find which sample direction the unit cell vertices point in. These two approaches gives the same result when plotting the unit cell.

However, when using `convention="mtex"`, we must also use MTEX' way when multiplying sample (rotation) or crystal directions (inverse rotation) to find which crystal and sample directions they point in, respectively. If we would have switched which time we inversed the rotation, we wouldn't get results consistent with the unit cell plot.

This tells me that lab2crystal (consistent with MTEX) should be named crystal2lab, and that we must always multiply with the rotation when transforming sample directions, and the inverse rotation when transforming crystal directions.

> For the notebook case if rotations in orix are passive, then the sample reference frame should be animated to rotate around a fixed crystal, instead of as it is now.

I agree that the way it is plotted now is the best. The way I interpret the unit cell plot is to start out with the crystal and sample reference frames parallel, and then rotating the unit cell by applying the three Euler angles consecutively about crystal Z, new crystal X, and then new crystal Z. Is this not how you interpret it, @harripj?

hakonanes:
I'd like to suggest the following conventions for orix:
1. Orientations describe coordinate transformations from the sample reference frame to the crystal reference frame (Bunge's definition)
2. Sample coordinates r are translated into crystal coordinates h by multiplying by the orientation: h = g * r
3. Crystal coordinates h are translated into sample coordinates r by multiplying by the inverse orientation: r = g^-1 * h
4. Consequtive rotations g1, g2, g3, are combined by multiplying in reverse order: g4 = g3 * g2 * g1

The existing conversions between rotation representations and rotation of vectors by multiplication of quaternions are consistent with these conventions. Convention 2. is also used in the unreleased inverse pole figure (IPF) functionality.

From the notebook PDF [linked above](https://github.com/pyxem/orix/pull/234#issuecomment-987812565), we get:

```python
>>> import numpy as np
>>> from orix.vector import Miller, Vector3d
>>> euler = (45, 90, 0)
>>> ori = Orientation.from_euler(np.radians(euler), symmetry=cubic.point_group, convention="bunge")
>>> ori.plot_unit_cell()
```

image

The crystal directions corresponding to the three sample directions (x, y, z) in an inverse pole figure

```python
>>> r1 = Vector3d((1, 0, 0))
>>> h1 = ori * r1
>>> h1 = Miller(uvw=h1.data, phase=cubic)
>>> h1
Miller (1,), point group m-3m, uvw
[[ 0.7071 -0.      0.7071]]
>>> ori.scatter("ipf", direction=Vector3d(((1, 0, 0), (0, 1, 0), (0, 0, 1))))
```

image

The sample directions corresponding to all crystal directions symmetrically equivalent to [100] in a pole figure

```python
>>> h2 = Miller(uvw=(1, 0, 0), phase=cubic)
>>> r2 = ~ori * h2
>>> r2 = Vector3d(r2)
>>> r2
Vector3d (1,)
[[0.7071 0.7071 0.    ]]
>>> (~ori * h1.symmetrise()).scatter(axes_labels=["x", "y", None], hemisphere="both")
```

image

Combining rotations

```python
>>> ori1 = Orientation.from_euler(np.radians((45, 0, 0)))
>>> ori2 = Orientation.from_euler(np.radians((0, 90, 0)))
>>> ori3 = ori2 * ori1  # Correct
>>> ori4 = ori1 * ori2  # Wrong
>>> ori3.plot_unit_cell()
```

image

```python
>>> ori4.plot_unit_cell()
```

image

harripj:
> I agree that the way it is plotted now is the best. The way I interpret the unit cell plot is to start out with the crystal and sample reference frames parallel, and then rotating the unit cell by applying the three Euler angles consecutively about crystal Z, new crystal X, and then new crystal Z. Is this not how you interpret it, @harripj?

This is how I interpret it too, and how I think is intuitive for most people. However this is an active rather than passive rotation.

I think if orix defines that rotations are passive, ie. transformations from the sample reference frame to crystal reference frame, then that means that the transformations happening in the notebook (rotating the unit cell) must be inverted as they are active. ...I think we are gradually coming to an agreement here!

harripj:
> I'd like to suggest the following conventions for orix:
> Orientations describe coordinate transformations from the sample reference frame to the crystal reference frame (Bunge's definition)
> Sample coordinates r are translated into crystal coordinates h by multiplying by the orientation: h = g * r
> Crystal coordinates h are translated into sample coordinates r by multiplying by the inverse orientation: r = g^-1 * h
> Consequtive rotations g1, g2, g3, are combined by multiplying in reverse order: g4 = g3 * g2 * g1

I agree with all of these statements. The final thing then would be to correct `lab2crystal`, as it should describe the sample to crystal reference frame transformation.

hakonanes:
> This is how I interpret it too, and how I think is intuitive for most people. However this is an active rather than passive rotation.
> 
> I think if orix defines that rotations are passive, ie. transformations from the sample reference frame to crystal reference frame, then that means that the transformations happening in the notebook (rotating the unit cell) must be inverted as they are active.

Hm, I confused myself when stating that passive rotations should be combined in order, not in reverse order, which was stated in the Rowenhorst et al. (2015) paper... I guess we can combine passive rotations in reverse order, which gives expected results with existing functionality.

> I agree with all of these statements. The final thing then would be to correct `lab2crystal`, as it should describe the sample to crystal reference frame transformation.

Puh, I'm relieved and happy that we agree :) Yes, as long as @pc494 gives his blessing for the conventions above and interchanging lab2crystal/crystal2lab, I think we have a consistent framework.

hakonanes:
@IMBalENce, you've said previously that an (interactive) unit cell display would be nice: do you have any input on the functionality added in this PR shown in https://github.com/pyxem/orix/pull/234#issuecomment-988136832 above?

pc494:
> I'd like to suggest the following conventions for orix:
> 
>1. Orientations describe coordinate transformations from the sample reference frame to the crystal reference frame (Bunge's definition)

I don't have a strong opinion about which way this operation works internally, but the io needs to account for both options (this was discussed in review of Density Based Clustering...)

>2. Sample coordinates r are translated into crystal coordinates h by multiplying by the orientation: h = g * r

This follows from 1 as far as I can tell.

>3. Crystal coordinates h are translated into sample coordinates r by multiplying by the inverse orientation: r = g^-1 * h

This follows from 1 as far as I can tell.

>4. Consequtive rotations g1, g2, g3, are combined by multiplying in reverse order: g4 = g3 * g2 * g1

This follows from 2 when working with the sample to crystal coordinates case. There is a bunch of inversions needed when going the other way.

===========

I think my feeling is that although this seems like a huge oversight that we've missed, as most work orix was able to do (prior to the recent features additions) was misorientation based, these inconsistencies cancelled out.

Broadly we have followed Rowenhorst et al's:
_"Therefore, for a crystal whose standard Cartesian reference frame is rotated with respect
to the specimen reference frame, the orientation of the crystal will be described as a passive
rotation of the sample reference frame to coincide with the crystal’s standard reference frame"_

I was going to check this against the OP, but ironically 45 degree rotations hide a plus minus error

hakonanes:
Thank you for your input @pc494.

> I don't have a strong opinion about which way this operation works internally, but the io needs to account for both options (this was discussed in review of Density Based Clustering...)

Eq. (1) in the clustering paper aligns with MTEX, i.e. opposite to Bunge's, as far as I can tell (opposite of convention 3 above). We could add a notice in a future "Orientation" user guide notebook stating that the convention in orix is different from the one used in the paper.

> This follows from 2 when working with the sample to crystal coordinates case. There is a bunch of inversions needed when going the other way.

Thanks for this point... I've always combined rotations this way in orix, the same way as done in MTEX. However, re-reading the Rowenhorst et al. (2015) paper, I got confused by the following statement (pp. 14-15):

*In quaternion form, the combined rotation C can be derived by consecutive application of the quaternion operators L_qA(r) and L_qB(r) in the correct order; since these quaternions, according to convention 3, represent passive rotations, we must combine them in the order qAqB , similar to the order of the passive matrices above.*

> I think my feeling is that although this seems like a huge oversight that we've missed, as most work orix was able to do (prior to the recent features additions) was misorientation based, these inconsistencies cancelled out.

With this unit cell plot, the conventions written out in an "Orientations" notebook, and the switch of the meaning of the `direction` parameter in `Rotation.from_euler()`, I think this oversight is handled.

> Broadly we have followed Rowenhorst et al's: _"Therefore, for a crystal whose standard Cartesian reference frame is rotated with respect to the specimen reference frame, the orientation of the crystal will be described as a passive rotation of the sample reference frame to coincide with the crystal’s standard reference frame"_

Good to hear! Their paper seems to be the source for rotation representation conversions, vector rotations by a quaternion, and quaternion multiplication. Is anything in orix based on Morawiec's book "Orientations and rotations"? If so, it should be stated where relevant. So far his book is not referenced.

> I was going to check this against the OP, but ironically 45 degree rotations hide a plus minus error

I believe the plots in https://github.com/pyxem/orix/pull/234#issuecomment-988136832 illustrate the conventions nicely?

pc494:
Although I think we're making progress, a new issue detailing exactly what needs to be done to correctly document (and apply) these (or a suitably adjusted version of these) changes seems like a good idea. Then we could make a (basically internal) rewrite the goal of v0.9.0

hakonanes:
@harripj, feel free to mark this PR as ready for review (I think you can do this?), or just say so, when you're happy with the functionality. I'll then review it. As mentioned, I have some thoughts on the expansion of the user guide, so I might push some changes myself, if that's OK?

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,195 @@
...
+def _plot_unit_cell(
+    rotation, c="tab:blue", axes_length=0.5, structure=None, **arrow_kwargs
+):
+    """Plot unit cell orientation.
+    TODO: define rotation reference frame (active vs passive).
+
+    Parameters
+    ----------
+    rotation : orix.quaternion.Rotation
+        Rotation of the unit cell (TODO).
```
harripj, line 88: We should specify as `'Active rotation of the unit cell.'`

orix/quaternion/orientation.py
```diff
@@ -620,6 +620,54 @@ def get_distance_matrix(self, lazy=False, chunk_size=20, progressbar=True):
...
+        ValueError
+            If self.size > 1.
+
+        """
+        if self.size > 1:
+            raise ValueError("Can only plot a single unit cell, so *size* must be 1")
+
+        from orix.plot.unit_cell_plot import _plot_unit_cell
+
+        fig = _plot_unit_cell(
```
harripj, line 664: This would also become `~self`.

harripj (commented):
Given the comments in this PR and in expectation of #249, we define orientations in orix as passive rotations of the sample reference frame. In this case then the rotation of the unit cell calculated in this PR is active and so I believe that the rotation applied to the unit cell vectors in this PR should be inverted. What do you think @hakonanes @pc494?

orix/plot/unit_cell_plot.py
hakonanes: I agree.

orix/quaternion/orientation.py
hakonanes: Agree here as well!

pc494: yup, looks correct to me

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+import numpy as np
+
+from orix.vector import Vector3d
+
+from matplotlib.patches import FancyArrowPatch
+from mpl_toolkits.mplot3d import proj3d
+import numpy as np
+from diffpy.structure import Structure, Lattice
+
+from orix.vector import Vector3d
```
hakonanes, line 35: See note on listing of package imports in the [contributing guide](https://orix.readthedocs.io/en/stable/contributing.html#code-style).

```suggestion
from itertools import combinations, product

from diffpy.structure import Structure, Lattice
from matplotlib.patches import FancyArrowPatch
from mpl_toolkits.mplot3d import proj3d
import matplotlib.pyplot as plt
import numpy as np

from orix.vector import Vector3d
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+
+# taken from SO post https://stackoverflow.com/a/22867877/12063126
+class Arrow3D(FancyArrowPatch):
+    "Matplotlib arrows in 3D."
```
hakonanes, line 39: These arrows can be useful in the fundamental zone plots as well (see for example the [orientation clustering user guide](https://orix.readthedocs.io/en/stable/clustering_orientations.html#Visualisation)), I suggest moving the class to a private `orix/plot/_util.py` file for now.

```suggestion
class Arrow3D(FancyArrowPatch):
    """Matplotlib arrows in 3D with arrowheads.
    
    Taken from this Stackoverflow answer by user CT Zhu
    https://stackoverflow.com/a/22867877/12063126.
    """
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def __init__(self, xs, ys, zs, *args, **kwargs):
+        FancyArrowPatch.__init__(self, (0, 0), (0, 0), *args, **kwargs)
+        self._verts3d = xs, ys, zs
+
+    def draw(self, renderer):
+        xs3d, ys3d, zs3d = self._verts3d
+        xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, self.axes.M)
+        # no in plane projection, ie. vector is vertical
```
hakonanes, line 48: When is this case relevant? I'd remove the commented lines now if you don't anticipate them to become relevant.

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+    verts = np.array(list(product(*zip((0, 0, 0), (a1, a2, a3)))))
+    center = verts.mean(axis=0)
+    return verts - center  # center on (0, 0, 0)
+
+
+def _calculate_basic_unit_cell_edges(verts, a1, a2, a3):
+    """Calculate valid unit cell edges for orthorhombic until cells."""
+    verts = _calculate_basic_unit_cell_vertices(a1, a2, a3)
+    # get valid edges from all unit cell egde possibilities unit cell
+    edges_valid = [
```
hakonanes, line 71: Can you build this list by appending in a loop instead of list comprehension? I think it will increase readability.

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+
+def _plot_unit_cell(
+    rotation, c="tab:blue", axes_length=0.5, structure=None, **arrow_kwargs
+):
+    """Plot unit cell orientation.
+
+    Parameters
+    ----------
+    rotation : orix.quaternion.Rotation
+        Active rotation of the unit cell.
```
hakonanes, line 87: In anticipation of [#249](https://github.com/pyxem/orix/issues/249#issuecomment-991249451), I think we shouldn't mention active/passive rotation. Write "Transformation of the sample reference frame to the crystal reference frame", or just "Rotation of the unit cell" for now?

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+        for s, e in combinations(verts, 2)
+        if np.isclose((a1, a2, a3), np.linalg.norm(s - e)).any()
+    ]
+    return np.array(edges_valid)
+
+
+def _plot_unit_cell(
+    rotation, c="tab:blue", axes_length=0.5, structure=None, **arrow_kwargs
+):
+    """Plot unit cell orientation.
```
hakonanes, line 82:
```suggestion
    """Plot a unit cell orientation, also showing the sample and crystal reference frames.
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+    rotation : orix.quaternion.Rotation
+        Active rotation of the unit cell.
+    c : str, optional
+        Unit cell edge color.
+    axes_length : float, optional
+        Length of the reference axes in Angstroms, by default 0.5.
+    structure : diffpy.structure.Structure or None, optional
+        Structure of the unit cell, only orthorhombic lattices are currently
+        supported. By default None, in which case a cubic unit cell with lattice
+        parameter a=2 will be plotted.
```
hakonanes, line 95:
```suggestion
        supported. If not given, a cubic unit cell with a lattice parameter of 2
        will be plotted.
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+    **arrow_kwargs : dict, optional
```
hakonanes, line 96:
```suggestion
    arrow_kwargs : dict, optional
        Keyword arguments passed to
        :class:`matplotlib.patches.FancyArrowPatch`, for example"arrowstyle".
```

I believe intersphinx will point to Matplotlib's API reference with this syntax.

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+        Passed to matplotlib.patches.FancyArrowPatch, for example 'arrowstyle'.
+
+    Returns
+    -------
+    fig: matplotlib.figure.Figure
+        The plotted figure.
+
```
hakonanes, line 103: Remove these two blank lines.

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+    structure : diffpy.structure.Structure or None, optional
+        Structure of the unit cell, only orthorhombic lattices are currently
+        supported. By default None, in which case a cubic unit cell with lattice
+        parameter a=2 will be plotted.
+    **arrow_kwargs : dict, optional
+        Passed to matplotlib.patches.FancyArrowPatch, for example 'arrowstyle'.
+
+    Returns
+    -------
+    fig: matplotlib.figure.Figure
```
hakonanes, line 101:
```suggestion
    fig : matplotlib.figure.Figure
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+        structure, Structure
+    ), "Structure must be diffpy.structure.Structure."
+    lattice = structure.lattice
+    if not (lattice.alpha == lattice.beta == lattice.gamma == 90):
+        raise ValueError("Only orthorhombic lattices are currently supported.")
+    a1, a2, a3 = lattice.a, lattice.b, lattice.c
+
+    verts = _calculate_basic_unit_cell_vertices(a1, a2, a3)
+    edges = _calculate_basic_unit_cell_edges(verts, a1, a2, a3)
+    edges_rotated = rotation * Vector3d(edges)
```
hakonanes, line 121: As per #249, this transformation is from the crystal reference frame to the sample reference frame, hence the inverse rotation must be applied. This is cleaner design than requiring an inverted rotation as input. The inverted rotation could be its own variable within this function, since it's used multiple times.

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+
+    fig, ax = plt.subplots(subplot_kw=dict(projection="3d"))
+    ax.axis("off")
+    ax.set_box_aspect((1, 1, 1))  # equal aspect
+
+    # xrange, yrange, zrange = np.ptp(verts, axis=0)
```
hakonanes, line 127: Remove this line if unused?

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+    colors = ("tab:red", "yellow", "tab:green")
+    labels = ("x", "y", "z")
+
+    arrow_kwargs.setdefault("mutation_scale", 20)
+    arrow_kwargs.setdefault("arrowstyle", "-")
+    arrow_kwargs.setdefault("linewidth", 2)
+
+    # add lab reference frame axes and labels
+    for i in range(3):
+        _data = np.full((3, 2), -1.4 * lim)  # less padding than axlim
```
hakonanes, line 152: Why the preceding underscore?

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+            *_data[:, 1], f"${_label}_s$", label=f"Sample reference axes label {_label}"
+        )  # _s for sample
+
+    for i, (v1, v2) in enumerate(edges_rotated.data):
+        ax.plot3D(*zip(v1, v2), c=c, label=f"Lattice edge {i}")
+
+    # add crystal reference frame axes and labels
+    for i, v in enumerate(Vector3d(np.eye(3))):
+        # rotate vector
+        v1 = (rotation * v).data.ravel() * axes_length
```
hakonanes, line 172: This could be done outside the loop, right?

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,194 @@
...
+            color=colors[i],
+            label=f"Crystal reference axes {_label}",
+            **arrow_kwargs,
+        )
+        ax.add_artist(arrow)
+        ax.text3D(
+            *_data[:, 1],
+            f"${_label}_c$",
+            label=f"Crystal reference axes label {_label}",
+        )  # _c for crystal
```
hakonanes, line 192: Place comment above line so the arguments can go on one line?

orix/quaternion/orientation.py
```diff
@@ -620,6 +620,56 @@ def get_distance_matrix(self, lazy=False, chunk_size=20, progressbar=True):
...
 
+    def plot_unit_cell(
+        self,
+        c="tab:blue",
+        return_figure=False,
+        axes_length=0.5,
+        structure=None,
+        **arrow_kwargs,
+    ):
+        """Plot unit cell orientation.
```
hakonanes, line 631: Same docstring as the private function?

orix/quaternion/orientation.py
```diff
@@ -620,6 +620,56 @@ def get_distance_matrix(self, lazy=False, chunk_size=20, progressbar=True):
...
+        Returns
+        -------
+        fig: matplotlib.figure.Figure
+            The plotted figure.
+
+        Raises
+        ------
+        ValueError
+            If self.size > 1.
+
```
hakonanes, line 657: Remove blank line, unless there's a reason for it?

orix/quaternion/orientation.py
```diff
@@ -620,6 +620,56 @@ def get_distance_matrix(self, lazy=False, chunk_size=20, progressbar=True):
...
+        """
+        if self.size > 1:
+            raise ValueError("Can only plot a single unit cell, so *size* must be 1")
+
+        from orix.plot.unit_cell_plot import _plot_unit_cell
+
+        # _plot_unit_cell requires active rotation of the unit cell
+        # => invert passive orientation (~self)
+        fig = _plot_unit_cell(
+            ~self, c=c, axes_length=axes_length, structure=structure, **arrow_kwargs
```
hakonanes, line 667: As previously said, I think it's cleaner design to invert the rotation inside the function.

orix/tests/plot/test_unit_cell_plot.py
```diff
@@ -0,0 +1,62 @@
...
+# You should have received a copy of the GNU General Public License
+# along with orix.  If not, see <http://www.gnu.org/licenses/>.
+
+from packaging import version
+
+import pytest
+from matplotlib import __version__ as _MPL_VERSION
+
+from diffpy.structure import Structure, Lattice
+from orix.quaternion import Orientation
```
hakonanes, line 25:
```suggestion
from diffpy.structure import Lattice, Structure
from matplotlib import __version__ as _MPL_VERSION
import pytest

from orix.quaternion import Orientation
```

hakonanes (commented):
You've greatly improved the plotting functionality here, @harripj, thanks!

I've several suggestions for improvements and some questions.

For nbsphinx to format the notebook correctly we need to do some changes, I'll test it out myself locally. I'm not sure if we will get interactive plotting via ReadtheDocs, and if not so I'm not so sure we should include this notebook in the user guide, but instead a separate "Examples" section with examples that don't fit in the user guide (think tutorials). Will check.

orix/plot/unit_cell_plot.py
harripj: In the case where the projection is parallel to the arrow, the arrow head is not perfectly represented. I will think about how to better implement this.

orix/plot/unit_cell_plot.py
harripj: I will go for the latter whilst we work on #249. I think that it would actually be rotation of the crystal to sample reference frame in this case (opposite of Bunge).

orix/plot/unit_cell_plot.py
harripj: Yes, you're right!

orix/plot/unit_cell_plot.py
harripj: There must have been a variable name conflict when I was working on this- removed the underscore.

orix/plot/unit_cell_plot.py
harripj: For now, for the default case of no arrow head, this works fine. It would be better to improve this by showing a dot or cross within a circle (depending on arrow direction) automatically, however this would also need to be removed if the plot is in interactive mode and rotated (as the arrow would no longer per parallel to the screen normal). This seems to be a lot of effort for not much gain and I think the current functionality is okay for what we are trying to show here (basic unit cell orientation representation).

hakonanes: Thank you for clarifying. These are good points to keep in mind for later improvements, if we find time as you say.

harripj:
In fact I have built upon this a bit further to use the lattice basis vectors rather than just the vector lengths. In principle now this can plot hexagonal unit cells (I had to do a bit of a workaround to stop the (0000)-(11-20) edge being plotted). It would be important to define the reference frame for the hexagonal unit cell (at the moment it is still cubic).

```python
from diffpy.structure import Structure, Lattice

ori = Orientation.identity()
lattice = Lattice(1, 1, 2, 90, 90, 120)
structure = Structure(lattice=lattice)
fig = ori.plot_unit_cell(return_figure=True, structure=structure)
```
image

orix/tests/plot/test_unit_cell_plot.py
```diff
@@ -0,0 +1,121 @@
...
+        ValueError, match="Crystal_reference_frame_axes_position must be either"
+    ):
+        ori.plot_unit_cell(crystal_reference_frame_axes_position="test")
+
+
+def test_arrow3D():
+    fig, ax = plt.subplots(subplot_kw=dict(projection="3d"))
+    a = Arrow3D((0, 1), (0, 1), (0, 1), arrowstyle="-|>", mutation_scale=20)
+    ax.add_artist(a)
+    plt.draw()
```
hakonanes, line 121: This command is what emits the mentioned warning

```python
  /home/hakon/kode/orix/orix/tests/plot/test_unit_cell_plot.py:121: MatplotlibDeprecationWarning: The 'renderer' parameter of do_3d_projection() was deprecated in Matplotlib 3.4 and will be removed two minor releases later.
    plt.draw()
```

After thinking about it, I think it is OK to leave it as a reminder that this might be something we should fix when Matplotlib 3.6 comes.

orix/tests/plot/test_unit_cell_plot.py
```diff
@@ -0,0 +1,121 @@
...
+    ori = Orientation.random((2,))
+    with pytest.raises(ValueError, match="Can only plot a single unit cell"):
+        ori.plot_unit_cell()
+
+
+def test_unit_cell_plot_orthorhombic():
+    ori = Orientation.random()
+    lattice = Lattice(1, 2, 3, 90, 90, 90)
+    structure = Structure(lattice=lattice)
+    fig = ori.plot_unit_cell(return_figure=True, structure=structure)
```
hakonanes, line 57:
```suggestion
    _ = ori.plot_unit_cell(return_figure=True, structure=structure)
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,199 @@
...
+    center = verts.mean(axis=0)
+    return verts - center  # center on (0, 0, 0)
+
+
+def _calculate_basic_unit_cell_edges(verts, vectors):
+    """Calculate valid unit cell edges for unit cells."""
+    vectors = np.asarray(vectors)
+    assert vectors.shape == (3, 3), "Vectors must be (3, 3) array."
+    a1, a2, a3 = np.linalg.norm(vectors, axis=-1)
+    # get valid edges from all unit cell egde possibilities unit cell
```
hakonanes, line 49:
```suggestion
    # get valid edges from all unit cell edge possibilities unit cell
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,199 @@
...
+
+def _plot_unit_cell(
+    rotation,
+    c="tab:blue",
+    axes_length=0.5,
+    structure=None,
+    crystal_reference_frame_axes_position="origin",
+    **arrow_kwargs,
+):
+    """Plot the unit cell orientation, showing the sample and crystal reference frames.
```
hakonanes, line 72: As per the [code style section](https://orix.readthedocs.io/en/stable/contributing.html#code-style) in the contributing guide, which references the [Python style guide](https://www.python.org/dev/peps/pep-0008/#maximum-line-length), I suggest keeping comment lines limited to 72 characters.

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,199 @@
...
+    fig : matplotlib.figure.Figure
+        The plotted figure.
+    """
+    # requires active rotation of the lattice in the sample reference frame
+    rotation = ~rotation
+
+    crystal_reference_frame_axes_position = (
+        crystal_reference_frame_axes_position.lower()
+    )
+    if not crystal_reference_frame_axes_position in ("origin", "center"):
```
hakonanes, line 104: PyCharm informs me that this should be

```suggestion
    if crystal_reference_frame_axes_position not in ("origin", "center"):
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,199 @@
...
+
+    for i, (v1, v2) in enumerate(edges_rotated.data):
+        ax.plot3D(*zip(v1, v2), c=c, label=f"Lattice edge {i}")
+
+    # add crystal reference frame axes and labels
+    v_ref_ax = rotation * Vector3d(np.eye(3))
+    if crystal_reference_frame_axes_position == "origin":  # cell origin
+        crys_ref_ax_origin = Vector3d(verts[0])
+    else:
+        crys_ref_ax_origin = Vector3d(np.zeros(3))
```
hakonanes, line 174: There's a method for this!

```suggestion
        crys_ref_ax_origin = Vector3d.zero(1)
```

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,199 @@
...
+        Keyword arguments passed to
+        :class:`matplotlib.patches.FancyArrowPatch`, for example "arrowstyle".
+
+    Returns
+    -------
+    fig : matplotlib.figure.Figure
+        The plotted figure.
+    """
+    # requires active rotation of the lattice in the sample reference frame
+    rotation = ~rotation
```
hakonanes, line 99: To make this fact very clear, I suggest naming the new variable something like `inv_rotation`.

orix/tests/plot/test_unit_cell_plot.py
```diff
@@ -0,0 +1,121 @@
...
+def test_unit_cell_plot_default():
+    ori = Orientation.random()
+    fig = ori.plot_unit_cell(return_figure=True)
+    assert len(fig.axes) == 1
+    axes = fig.axes[0]
+    assert len(axes.lines) == 12  # 12 edges in orthorhombic unit cell
+    # 6 Arrow3D -> 3 for both sample and crystal reference frames
+    if version.parse(_MPL_VERSION) >= version.parse("3.4"):  # pragma: no cover
+        assert len(axes.patches) == 6
+    else:
```
hakonanes, line 40: I've got 3.5 on my machine, so these lines aren't covered

```suggestion
    else:  # pragma: no cover
```

hakonanes (commented):
Great extension with allowing to set the location of the crystal reference frame axes location! I've added some suggestions for improvements.

I cannot get nbsphinx to create an interactive notebook, so I suggest to leave the notebook out of this PR and instead include uses of the unit cell in later PRs following our discussion on conventions. What do you think?

I have to look a bit at the hexagonal unit cell you just added. I think the PR has reached the initial goal of "simple plotting of unit cells", and suggest that we merge after considering the my suggestions and after Phillip approves.

hakonanes:
> I have to look a bit at the hexagonal unit cell you just added.

Looks good to me!

orix/plot/unit_cell_plot.py
```diff
@@ -0,0 +1,190 @@
...
+        Rotation of the unit cell.
+    c : str, optional
+        Unit cell edge color.
+    axes_length : float, optional
+        Length of the reference axes in Angstroms, by default 0.5.
+    structure : diffpy.structure.Structure or None, optional
+        Structure of the unit cell, only orthorhombic lattices are currently
+        supported. If not given, a cubic unit cell with a lattice parameter of
+        2 Angstroms will be plotted.
+    crystal_reference_frame_axes_position : str, optional
```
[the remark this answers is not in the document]
harripj: Yes it is very long, that caught my eye too. Let's go with `crystal_axes_loc`.

orix/tests/plot/test_unit_cell_plot.py
harripj: I run 3.5 too. There is a python 3.6 build on CI, which uses Matplotlib 3.3 I believe and is why this is needed.

orix/plot/unit_cell_plot.py
harripj: Thanks, didn't know this!

orix/plot/unit_cell_plot.py
harripj: Yep thanks, I was too busy looking at my 88 char line limit!

orix/tests/plot/test_unit_cell_plot.py
harripj: In fact this `plt.draw()` is only necessary for coverage. Without this `Arrow3D.draw` was not being called. Perhaps this is because the testing is not actually rendering the plots. Would you happen to know?

harripj:
> I cannot get nbsphinx to create an interactive notebook, so I suggest to leave the notebook out of this PR and instead include uses of the unit cell in later PRs following our discussion on conventions. What do you think?

No problem, I will have a look into this at some point too, but for now it's fine if we leave it out. I will remove it from the changelog.

> I have to look a bit at the hexagonal unit cell you just added. I think the PR has reached the initial goal of "simple plotting of unit cells", and suggest that we merge after considering the my suggestions and after Phillip approves.

Yes, I think that we have given the users enough to play with here!

orix/plot/_util.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def __init__(self, xs, ys, zs, *args, **kwargs):
+        FancyArrowPatch.__init__(self, (0, 0), (0, 0), *args, **kwargs)
+        self._verts3d = xs, ys, zs
+
+    def draw(self, renderer):
+        xs3d, ys3d, zs3d = self._verts3d
+        xs, ys, zs = proj3d.proj_transform(xs3d, ys3d, zs3d, self.axes.M)
+        self.set_positions((xs[0], ys[0]), (xs[1], ys[1]))
+        FancyArrowPatch.draw(self, renderer)
```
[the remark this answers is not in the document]
harripj: In fact it looks like this change breaks the CI testing. I will revert this and perhaps we deal with the deprecation later.

orix/tests/plot/test_unit_cell_plot.py
hakonanes: That must be the reason. Rendering plots during tests is undesirable, so this is seems to be the best approach.

orix/tests/plot/test_unit_cell_plot.py
hakonanes: Yes, that's the [minimal version we support](https://github.com/pyxem/orix/blob/master/setup.py#L50).

orix/plot/_util.py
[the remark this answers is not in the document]
hakonanes: I can see that ([failing test](https://github.com/pyxem/orix/runs/4512152126?check_suite_focus=true#step:8:119)). Keep it as it is.

doc/index.rst
```diff
@@ -29,6 +29,7 @@ Work using orix
     clustering_across_fundamental_region_boundaries.ipynb
     clustering_orientations.ipynb
     uniform_sampling_of_orientation_space.ipynb
+    interactive_euler_angles.ipynb
```
hakonanes, line 32:
```suggestion
```

pc494:
Can I confirm this is waiting on my review (and probable merge)?

hakonanes:
This PR has converged to a review-and-merge state, yes!

pc494 (approved):
Happy with this, hadn't seen the use of `assert` in production code before but it works obviously enough. Will merge once #252 is in.

harripj:
> Happy with this, hadn't seen the use of assert in production code before but it works obviously enough. Will merge once #252 is in.

You're right, these should be changed to `ValueError` and `TypeError` before merging. I should be able to do this before Monday!

hakonanes:
I'm changing the three `assert condition, message` to `if not condition: raise ValueError(message)`, as per [this SO answer](https://stackoverflow.com/questions/40182944/whats-the-difference-between-raise-try-and-assert) and [the Python docs](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement), which state that the latter statement should be used if the error should be raised in all cases the condition is not met.

hakonanes:
@harripj, not of importance now, but why does `Orientation.plot_unit_cell()` accept a `Structure` and not just a `Lattice`, since you only use that anyway? Do you anticipate adding plotting of atoms from `Structure.atoms` in the future?

hakonanes:
I'll not commit further here. It can be merged when the checks pass.

pc494:
Thanks @hakonanes I'll merge this on pass

harripj:
> @harripj, not of importance now, but why does `Orientation.plot_unit_cell()` accept a `Structure` and not just a `Lattice`, since you only use that anyway? Do you anticipate adding plotting of atoms from `Structure.atoms` in the future?

@hakonanes yes in the future atoms may also be displayed within the plot but it is out of the scope of this PR. I agree with you that for this PR a `Lattice` would suffice and be more intuitive as other aspects of the `Structure` are not currently used. This has now been merged so we can address it in the future.

PR #235, opened 2021-10-22
Inverse pole figure plotting and coloring

hakonanes:
#### Description of the change
New functionality:
1. Rotations of directions in the sample coordinate system by orientations to find crystal coordinate directions parallel to these sample directions. This is available via `Orientation.scatter()` with the `projection="ipf"` option, which defaults to out-of-plane direction (IPFZ). Closes #122.
2. Coloring of orientations by specifying a direction in the sample reference frame and a crystal symmetry. Only distinguish between Laue groups so far, based on the TSL coloring in MTEX. Closes #166.
3. Projection of vectors, typically crystal coordinates, into the fundamental sector in the stereographic projection of a symmetry, via `Vector3d.in_fundamental_sector(sector)`. This is based on `vector3d.project2fundamentalRegion` in `MTEX`. Closes #169.

This functionality uses a new `orix.plot.InversePoleFigurePlot` class which inherits from `orix.plot.StereographicPlot`.

Other changes which were necessary but need some consideration:
1. All use of `Orientation.set_symmetry()` internally (via __invert__, __neg__, etc) is removed, because otherwise I couldn't get any directions to rotate where I wanted them. **All tests depending on indirect calls to `set_symmetry()` will fail for now**. See #233.
2. Took the *S2* cube mesh generation in `diffsims` implemented by @din14970. It would be best if you could make a PR with this code, perhaps to my branch `add-inverse-pole-figure-plot`, so that the contribution carries your name. As mentioned before, I think these functions belong in orix, since diffsims depends on orix.

#### Credit
The orientation coloring is based on the procedure explained in section 2.4 in Gert Nolze's and Ralf Hielscher's "Orientations – perfectly colored" [paper](https://doi.org/10.1107/S1600576716012942). MTEX's source code has also been of much help, specifically the functions [project2fundamentalRegion](https://github.com/mtex-toolbox/mtex/blob/develop/geometry/%40vector3d/project2FundamentalRegion.m) and [polarCoordinates](https://github.com/mtex-toolbox/mtex/blob/develop/geometry/%40sphericalRegion/polarCoordinates.m). Credit is made in the source code and documentation to these references.

The interpolation between colored vectors used in the IPF color key uses the `scipy.interpolate.griddata` approach suggested by @din14970 in #166.

#### Thoughts on patching this post-release
[...]

Issue #233, opened 2021-10-01
Assign symmetry to an orientation without calling Orientation.set_symmetry()


hakonanes:
I don't think `Orientation.set_symmetry()` should be the canonical way to assign a symmetry to an orientation. I suggest allowing to set the symmetry via a `property` setter instead.

`Orientation.set_symmetry()`, which calls `Misorientation.set_symmetry()`, was according to its docstring implemented to ensure a misorientation has the smallest angle of rotation. Is it strictly necessary for orientations to be stored with a smallest angle of rotation with respect to the sample reference frame? Couldn't this instead be an option for `Orientation`, perhaps as a method?

The reason I'm asking is that I find it difficult to debug IPF functionality etc. against MTEX, which doesn't change given quaternion or Euler angles when Orientation symmetries are assigned or changed.

If we make it possible to set the symmetry of an orientation without going via `Misorientation.set_symmetry()`, we should of course make sure that it is called wherever necessary, like when plotting in orientation space etc.

hakonanes:
If I find the time I'll make a PR allowing to set an orientation's symmetry without `set_symmetry()`.

harripj:
I have just come across this when testing #235, and completely agree with everything you have said here @hakonanes. For an ```Orientation``` ```set_symmetry``` finds the smallest angle with ```C1``` which I don't think is necessary.

I would favour a ```symmetry``` property and ```getter``` and ```setter``` methods as implemented in #235 for both ```Orientation``` and ```Misorientation``` methods which do not perform any calculations by default. Then another function similar to ```get_distance_matrix``` which will calculate the symmetry reduced orientations as required.

I was testing to see whether IPF (#235) will show agreement after a 360 degree rotation about a given axis. Using the ```symmetry``` property, ie. without calculating symmetry reduced orientations:

```python
%matplotlib widget
from orix import plot
from orix.vector import Vector3d
from matplotlib import pyplot as plt
from orix.quaternion import Orientation
import numpy as np
from KED.orientations import convert_to_orix, convert_to_scipy
from scipy.spatial.transform import Rotation
from orix.vector import AxAngle

o1 = Orientation.random()
# o1 = Orientation([-0.1528,  0.4705,  0.234,   0.837 ])
# o1 = Orientation([1, 0.1, 0, 0])

# rotation axis angle
axis = (0.3, 0.4, 0.5)
# axis= (1, 1, 1)
angle = 0.5

axis = np.array(axis) / np.linalg.norm(axis)

[...]

PR #240, opened 2021-11-10
Add symmetry property to Orientation-based classes in favour of set_symmetry

harripj:
As discussed in #235 the use of ```set_symmetry``` produces behaviour which may be undesired in certain circumstances. As such a ```symmetry``` property has been introduced for ```Orientation``` and ```Misorientation``` classes which sets the instances ```_symmetry``` attribute without further calculations.

These extra calculations, in the case of the ```Orientation``` class, take the form of finding the symmetry component with the smallest angular distance to the ```C1``` symmetry, which is the identity orientation. In this PR no calculations are automatically performed when the ```symmetry``` property is set. A new function currently named ```compute_symmetry_reduced_orientations``` performs the calculations previously defined within ```set_symmetry```. In this case, the old functionality afforded by ```set_symmetry```:
```python
o = Orientation.random((5, 3))
o.set_symmetry(Oh)
```
can still be achieved as:

```python
o = Orientation.random((5, 3))
o.symmetry = Oh
o = o.compute_symmetry_reduced_orientations()
```

This would be in favour of possibly deprecating ```set_symmetry``` in a future release.

The tests have been updated to use the new syntax above and are passing.

#### Progress of the PR

```python
o = Orientation.random((5, 3))
o.symmetry = Oh
o.symmetry
>>> Symmetry (48,) m-3m
    ...
```

#### For reviewers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,12 +33,9 @@ Deprecated
 ----------
 - The `data_dim` attribute of Object3d and all derived classes is deprecated from 0.8
   and will be removed in 0.9. Use `ndim` instead.
-- `set_symmetry` will be deprecated from 0.8 in favour of a `symmetry` property for
-  `Orientation` and `Misorientation` classes. `set_symmetry` will be removed in 0.9.
-  The behaviour of `set_symmetry` can still be achieved by using the following snippet:
-  >>> o = Orientation(data)
-  >>> o.symmetry = symmetry
-  >>> o = o.compute_symmetry_reduced_orientations()
+- Setting (Mis)Orientation symmetry via `set_symmetry()` is deprecated in 0.8, in favour of
+  setting it directly via a `symmetry.setter`, and will be removed in 0.9. Use
+  `assign_smallest_angle()` instead.
 
 Removed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -36,7 +36,9 @@ Deprecated
 - Setting (Mis)Orientation symmetry via `set_symmetry()` is deprecated in 0.8, in favour of
   setting it directly via a `symmetry.setter`, and will be removed in 0.9. Use
   `assign_smallest_angle()` instead.
-
+- `Orientation` class methods `from_euler`, `from_matrix`, and `from_neo_euler` no longer 
+  return the smallest angle rotation when `symmetry` argument is provided and not None.
+   
 Removed
 -------
 - `StereographicPlot` methods `azimuth_grid()` and `polar_grid()`.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,8 @@ Changed
 - `StereographicPlot` doesn't use Matplotlib's `transforms` framework anymore, and
   (X, Y) replaces (azimuth, polar) as internal coordinates.
 - Renamed `Symmetry` method `fundamental_sector()` to `fundamental_zone()`.
+- `Orientation` class methods `from_euler`, `from_matrix`, and `from_neo_euler` no longer 
+  return the smallest angle rotation when `symmetry` argument is provided and not None.
 
 Deprecated
 ----------
@@ -36,9 +38,7 @@ Deprecated
 - Setting (Mis)Orientation symmetry via `set_symmetry()` is deprecated in 0.8, in favour of
   setting it directly via a `symmetry.setter`, and will be removed in 0.9. Use
   `assign_smallest_angle()` instead.
-- `Orientation` class methods `from_euler`, `from_matrix`, and `from_neo_euler` no longer 
-  return the smallest angle rotation when `symmetry` argument is provided and not None.
-   
+ 
 Removed
 -------
 - `StereographicPlot` methods `azimuth_grid()` and `polar_grid()`.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,7 +21,7 @@ Added
 - Sampling of vectors in UV mesh on a unit sphere (*S2*).
 - `ndim` attribute to Object3d and derived classes which returns number of navigation
   dimensions.
-- `symmetry` property for `Orientation` and `Misorientation` classes.
+- Setting the symmetry of a (Mis)Orientation via a `symmetry.setter`.
 
 Changed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,7 @@ Changed
 - Renamed `Symmetry` method `fundamental_sector()` to `fundamental_zone()`.
 - `Orientation` class methods `from_euler`, `from_matrix`, and `from_neo_euler` no longer 
   return the smallest angle rotation when `symmetry` argument is provided and not None.
+- `CrystalMap.orientations` no longer returns smallest angle rotation.
 
 Deprecated
 ----------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ Added
 - Sampling of vectors in UV mesh on a unit sphere (*S2*).
 - `ndim` attribute to Object3d and derived classes which returns number of navigation
   dimensions.
+- `symmetry` property for `Orientation` and `Misorientation` classes.
 
 Changed
 -------
@@ -32,6 +33,12 @@ Deprecated
 ----------
 - The `data_dim` attribute of Object3d and all derived classes is deprecated from 0.8
   and will be removed in 0.9. Use `ndim` instead.
+- `set_symmetry` will be deprecated from 0.8 in favour of a `symmetry` property for
+  `Orientation` and `Misorientation` classes. `set_symmetry` will be removed in 0.9.
+  The behaviour of `set_symmetry` can still be achieved by using the following snippet:
+  >>> o = Orientation(data)
+  >>> o.symmetry = symmetry
+  >>> o = o.compute_symmetry_reduced_orientations()
 
 Removed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,7 +38,7 @@ Deprecated
   and will be removed in 0.9. Use `ndim` instead.
 - Setting (Mis)Orientation symmetry via `set_symmetry()` is deprecated in 0.8, in favour of
   setting it directly via a `symmetry.setter`, and will be removed in 0.9. Use
-  `assign_smallest_angle()` instead.
+  `map_into_symmetry_reduced_zone()` instead.
  
 Removed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,8 +29,8 @@ Changed
   (X, Y) replaces (azimuth, polar) as internal coordinates.
 - Renamed `Symmetry` method `fundamental_sector()` to `fundamental_zone()`.
 - `Orientation` class methods `from_euler`, `from_matrix`, and `from_neo_euler` no longer 
-  return the smallest angle rotation when `symmetry` argument is provided and not None.
-- `CrystalMap.orientations` no longer returns smallest angle rotation.
+  return the smallest angle orientation when a `symmetry` is given.
+- `CrystalMap.orientations` no longer returns smallest angle orientation.
 
 Deprecated
 ----------
--- a/orix/crystal_map/crystal_map.py
+++ b/orix/crystal_map/crystal_map.py
@@ -331,9 +331,9 @@ class CrystalMap:
                 rotations = self.rotations[:, 0]
             else:
                 rotations = self.rotations
-            o = Orientation(rotations)
-            o.symmetry = phases[:].point_group
-            return o.compute_symmetry_reduced_orientations()
+            orientations = Orientation(rotations)
+            orientations.symmetry = phases[:].point_group
+            return orientations
         else:
             raise ValueError(
                 f"Data has the phases {phases.names}, however, you are executing a "
--- a/orix/crystal_map/crystal_map.py
+++ b/orix/crystal_map/crystal_map.py
@@ -331,7 +331,9 @@ class CrystalMap:
                 rotations = self.rotations[:, 0]
             else:
                 rotations = self.rotations
-            return Orientation(rotations).set_symmetry(phases[:].point_group)
+            o = Orientation(rotations)
+            o.symmetry = phases[:].point_group
+            return o.compute_symmetry_reduced_orientations()
         else:
             raise ValueError(
                 f"Data has the phases {phases.names}, however, you are executing a "
--- a/orix/plot/rotation_plot.py
+++ b/orix/plot/rotation_plot.py
@@ -33,7 +33,7 @@ class RotationPlot(Axes3D):
     def transform(self, xs, fundamental_region=None):
         from orix.quaternion import Rotation, Misorientation, OrientationRegion
 
-        # check whether xs is in fundamental region
+        # Project rotations into fundamental zone if necessary
         if isinstance(xs, Misorientation):
             if fundamental_region is None:
                 if isinstance(xs.symmetry, tuple):
--- a/orix/plot/rotation_plot.py
+++ b/orix/plot/rotation_plot.py
@@ -30,22 +30,22 @@ class RotationPlot(Axes3D):
     name = None
     transformation_class = None
 
-    def transform(self, xs, fundamental_region=None):
+    def transform(self, xs, fundamental_zone=None):
         from orix.quaternion import Rotation, Misorientation, OrientationRegion
 
         # Project rotations into fundamental zone if necessary
         if isinstance(xs, Misorientation):
-            if fundamental_region is None:
+            if fundamental_zone is None:
                 if isinstance(xs.symmetry, tuple):
-                    fundamental_region = OrientationRegion.from_symmetry(
+                    fundamental_zone = OrientationRegion.from_symmetry(
                         xs.symmetry[0], xs.symmetry[1]
                     )
                 else:
-                    fundamental_region = OrientationRegion.from_symmetry(xs.symmetry)
-            # check fundamental_region is properly defined
-            if not isinstance(fundamental_region, OrientationRegion):
-                raise TypeError("Fundamental_region is not OrientationRegion.")
-            region_edge = fundamental_region.get_plot_data().angle.data.max()
+                    fundamental_zone = OrientationRegion.from_symmetry(xs.symmetry)
+            # check fundamental_zone is properly defined
+            if not isinstance(fundamental_zone, OrientationRegion):
+                raise TypeError("Fundamental_zone is not OrientationRegion.")
+            region_edge = fundamental_zone.get_plot_data().angle.data.max()
             # if xs is out of fundamental region, calculate symmetry reduction
             if xs.angle.data.max() > region_edge:
                 xs = xs.map_into_symmetry_reduced_zone()
@@ -57,8 +57,8 @@ class RotationPlot(Axes3D):
         x, y, z = transformed.xyz
         return x, y, z
 
-    def scatter(self, xs, fundamental_region=None, **kwargs):
-        x, y, z = self.transform(xs, fundamental_region=fundamental_region)
+    def scatter(self, xs, fundamental_zone=None, **kwargs):
+        x, y, z = self.transform(xs, fundamental_zone=fundamental_zone)
         return super().scatter(x, y, z, **kwargs)
 
     def plot(self, xs, **kwargs):
--- a/orix/plot/rotation_plot.py
+++ b/orix/plot/rotation_plot.py
@@ -30,8 +30,25 @@ class RotationPlot(Axes3D):
     name = None
     transformation_class = None
 
-    def transform(self, xs):
-        from orix.quaternion.rotation import Rotation
+    def transform(self, xs, fundamental_region=None):
+        from orix.quaternion import Rotation, Misorientation, OrientationRegion
+
+        # check whether xs is in fundamental region
+        if isinstance(xs, Misorientation):
+            if fundamental_region is None:
+                if isinstance(xs.symmetry, tuple):
+                    fundamental_region = OrientationRegion.from_symmetry(
+                        xs.symmetry[0], xs.symmetry[1]
+                    )
+                else:
+                    fundamental_region = OrientationRegion.from_symmetry(xs.symmetry)
+            # check fundamental_region is properly defined
+            if not isinstance(fundamental_region, OrientationRegion):
+                raise TypeError("Fundamental_region is not OrientationRegion.")
+            region_edge = fundamental_region.get_plot_data().angle.data.max()
+            # if xs is out of fundamental region, calculate symmetry reduction
+            if xs.angle.data.max() > region_edge:
+                xs = xs.map_into_symmetry_reduced_zone()
 
         if isinstance(xs, Rotation):
             transformed = self.transformation_class.from_rotation(xs.get_plot_data())
@@ -40,8 +57,8 @@ class RotationPlot(Axes3D):
         x, y, z = transformed.xyz
         return x, y, z
 
-    def scatter(self, xs, **kwargs):
-        x, y, z = self.transform(xs)
+    def scatter(self, xs, fundamental_region=None, **kwargs):
+        x, y, z = self.transform(xs, fundamental_region=fundamental_region)
         return super().scatter(x, y, z, **kwargs)
 
     def plot(self, xs, **kwargs):
--- a/orix/plot/rotation_plot.py
+++ b/orix/plot/rotation_plot.py
@@ -44,9 +44,8 @@ class RotationPlot(Axes3D):
             # check fundamental_zone is properly defined
             if not isinstance(fundamental_zone, OrientationRegion):
                 raise TypeError("Fundamental_zone is not an OrientationRegion object.")
-            region_edge = fundamental_zone.get_plot_data().angle.data.max()
-            # if xs is out of fundamental region, calculate symmetry reduction
-            if xs.angle.data.max() > region_edge:
+            # if any in xs is out of fundamental_zone, calculate symmetry reduction
+            if not (xs < fundamental_zone).all():
                 xs = xs.map_into_symmetry_reduced_zone()
 
         if isinstance(xs, Rotation):
--- a/orix/plot/rotation_plot.py
+++ b/orix/plot/rotation_plot.py
@@ -36,12 +36,11 @@ class RotationPlot(Axes3D):
         # Project rotations into fundamental zone if necessary
         if isinstance(xs, Misorientation):
             if fundamental_zone is None:
-                if isinstance(xs.symmetry, tuple):
-                    fundamental_zone = OrientationRegion.from_symmetry(
-                        xs.symmetry[0], xs.symmetry[1]
-                    )
-                else:
-                    fundamental_zone = OrientationRegion.from_symmetry(xs.symmetry)
+                symmetry = xs.symmetry
+                # Orientation.symmetry returns a Symmetry object not a tuple, so pack
+                if not isinstance(symmetry, tuple):
+                    symmetry = (symmetry,)
+                fundamental_zone = OrientationRegion.from_symmetry(*symmetry)
             # check fundamental_zone is properly defined
             if not isinstance(fundamental_zone, OrientationRegion):
                 raise TypeError("Fundamental_zone is not an OrientationRegion object.")
--- a/orix/plot/rotation_plot.py
+++ b/orix/plot/rotation_plot.py
@@ -44,7 +44,7 @@ class RotationPlot(Axes3D):
                     fundamental_zone = OrientationRegion.from_symmetry(xs.symmetry)
             # check fundamental_zone is properly defined
             if not isinstance(fundamental_zone, OrientationRegion):
-                raise TypeError("Fundamental_zone is not OrientationRegion.")
+                raise TypeError("Fundamental_zone is not an OrientationRegion object.")
             region_edge = fundamental_zone.get_plot_data().angle.data.max()
             # if xs is out of fundamental region, calculate symmetry reduction
             if xs.angle.data.max() > region_edge:
--- a/orix/plot/rotation_plot.py
+++ b/orix/plot/rotation_plot.py
@@ -44,7 +44,7 @@ class RotationPlot(Axes3D):
             # check fundamental_zone is properly defined
             if not isinstance(fundamental_zone, OrientationRegion):
                 raise TypeError("Fundamental_zone is not an OrientationRegion object.")
-            # if any in xs is out of fundamental_zone, calculate symmetry reduction
+            # if any in xs are out of fundamental_zone, calculate symmetry reduction
             if not (xs < fundamental_zone).all():
                 xs = xs.map_into_symmetry_reduced_zone()
 
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -234,7 +234,7 @@ class Misorientation(Rotation):
         return o_inside
 
     def distance(self, verbose=False, split_size=100):
-        """Symmetry reduced distance
+        """Symmetry reduced distance.
 
         Compute the shortest distance between all orientations
         considering symmetries.
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -382,7 +382,7 @@ class Misorientation(Rotation):
             to_plot = self.get_random_sample(size)
         else:
             to_plot = self
-        ax.scatter(to_plot, **kwargs)
+        ax.scatter(to_plot, fundamental_region=fundamental_region, **kwargs)
 
         if return_figure:
             return figure
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -49,7 +49,7 @@ from tqdm import tqdm
 
 from orix.quaternion.orientation_region import OrientationRegion
 from orix.quaternion.rotation import Rotation
-from orix.quaternion.symmetry import C1
+from orix.quaternion.symmetry import C1, Symmetry
 from orix.scalar import Scalar
 from orix._util import deprecated
 
@@ -127,6 +127,15 @@ class Misorientation(Rotation):
         """Tuple of :class:`~orix.quaternion.Symmetry`."""
         return self._symmetry
 
+    @symmetry.setter
+    def symmetry(self, values):
+        err_message = "Values must be a tuple of two Symmetry objects."
+        if not isinstance(values, (list, tuple)):
+            raise TypeError(err_message)
+        if len(values) != 2 or not all(isinstance(s, Symmetry) for s in values):
+            raise ValueError(err_message)
+        self._symmetry = values
+
     def __getitem__(self, key):
         m = super().__getitem__(key)
         m._symmetry = self._symmetry
@@ -178,6 +187,30 @@ class Misorientation(Rotation):
         [[-0.7071  0.7071  0.      0.    ]
         [ 0.      1.      0.      0.    ]]
         """
+        misori = self.__class__(self.data)
+        misori._symmetry = (Gl, Gr)
+        return misori.compute_symmetry_reduced_orientations()
+
+    def compute_symmetry_reduced_orientations(self, verbose=False):
+        """Computes equivalent transformations which have the smallest
+        angle of rotation and assigns these in-place.
+
+        Returns
+        -------
+        Misorientation
+            A new misorientation object with the assigned symmetry.
+
+        Examples
+        --------
+        >>> from orix.quaternion.symmetry import C4, C2
+        >>> data = np.array([[0.5, 0.5, 0.5, 0.5], [0, 1, 0, 0]])
+        >>> m = Misorientation(data).set_symmetry(C4, C2)
+        >>> m
+        Misorientation (2,) 4, 2
+        [[-0.7071  0.7071  0.      0.    ]
+        [ 0.      1.      0.      0.    ]]
+        """
+        Gl, Gr = self._symmetry
         symmetry_pairs = iproduct(Gl, Gr)
         if verbose:
             symmetry_pairs = tqdm(symmetry_pairs, total=Gl.size * Gr.size)
@@ -385,7 +418,8 @@ class Orientation(Misorientation):
         if isinstance(other, Orientation):
             # Call to Object3d.squeeze() doesn't carry over symmetry
             misorientation = Misorientation(self * ~other).squeeze()
-            return misorientation.set_symmetry(self.symmetry, other.symmetry)
+            misorientation.symmetry = (self.symmetry, other.symmetry)
+            return misorientation.compute_symmetry_reduced_orientations()
         return NotImplemented
 
     @classmethod
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -129,11 +129,10 @@ class Misorientation(Rotation):
 
     @symmetry.setter
     def symmetry(self, value):
-        err_message = "Value must be a tuple with two Symmetry instances."
         if not isinstance(value, (list, tuple)):
-            raise TypeError(err_message)
+            raise TypeError("Value must be a 2-tuple of Symmetry objects.")
         if len(value) != 2 or not all(isinstance(s, Symmetry) for s in value):
-            raise ValueError(err_message)
+            raise ValueError("Value must be a 2-tuple of Symmetry objects.")
         self._symmetry = tuple(value)
 
     def __getitem__(self, key):
@@ -382,7 +381,7 @@ class Misorientation(Rotation):
             to_plot = self.get_random_sample(size)
         else:
             to_plot = self
-        ax.scatter(to_plot, fundamental_region=fundamental_region, **kwargs)
+        ax.scatter(to_plot, fundamental_region, **kwargs)
 
         if return_figure:
             return figure
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -164,7 +164,7 @@ class Misorientation(Rotation):
 
     @deprecated(
         since="0.8",
-        alternative="orix.quaternion.Misorientation.assign_smallest_angle",
+        alternative="orix.quaternion.Misorientation.map_into_symmetry_reduced_zone",
         removal="0.9",
     )
     def set_symmetry(self, Gl, Gr, verbose=False):
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -619,7 +619,7 @@ class Orientation(Misorientation):
 
     @deprecated(
         since="0.8",
-        alternative="orix.quaternion.Orientation.assign_smallest_angle",
+        alternative="orix.quaternion.Orientation.map_into_symmetry_reduced_zone",
         removal="0.9",
     )
     def set_symmetry(self, symmetry):
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -134,7 +134,7 @@ class Misorientation(Rotation):
             raise TypeError(err_message)
         if len(values) != 2 or not all(isinstance(s, Symmetry) for s in values):
             raise ValueError(err_message)
-        self._symmetry = values
+        self._symmetry = tuple(values)
 
     def __getitem__(self, key):
         m = super().__getitem__(key)
@@ -162,6 +162,11 @@ class Misorientation(Rotation):
         equivalent = Gr.outer(orientations.outer(Gl))
         return self.__class__(equivalent).flatten()
 
+    @deprecated(
+        since="0.7",
+        alternative="orix.quaternion.Misorientation.compute_symmetry_reduced_orientations",
+        removal="0.8",
+    )
     def set_symmetry(self, Gl, Gr, verbose=False):
         """Assign symmetries to this misorientation.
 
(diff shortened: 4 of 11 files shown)


CHANGELOG.rst
```diff
@@ -21,6 +21,7 @@ Added
 - Sampling of vectors in UV mesh on a unit sphere (*S2*).
 - `ndim` attribute to Object3d and derived classes which returns number of navigation
   dimensions.
+- `symmetry` property for `Orientation` and `Misorientation` classes.
```
hakonanes, line 24: Both classes already have a symmetry ([see API](https://orix.readthedocs.io/en/stable/reference.html#orix.quaternion.Orientation.symmetry)).

```suggestion
- Setting the symmetry of a (Mis)Orientation via a `symmetry.setter`.
```

CHANGELOG.rst
```diff
@@ -32,6 +33,12 @@ Deprecated
 ----------
 - The `data_dim` attribute of Object3d and all derived classes is deprecated from 0.8
   and will be removed in 0.9. Use `ndim` instead.
+- `set_symmetry` will be deprecated from 0.8 in favour of a `symmetry` property for
+  `Orientation` and `Misorientation` classes. `set_symmetry` will be removed in 0.9.
+  The behaviour of `set_symmetry` can still be achieved by using the following snippet:
+  >>> o = Orientation(data)
+  >>> o.symmetry = symmetry
+  >>> o = o.compute_symmetry_reduced_orientations()
```
hakonanes, line 41: We should be brief in the changelog, ideally one, max. two lines per point.

```suggestion
- Setting (Mis)Orientation symmetry via `set_symmetry()` is deprecated in 0.8, in favour of
  setting it directly via a `symmetry.setter`, and will be removed in 0.9. Use
  `assign_smallest_angle()` instead.
```

orix/crystal_map/crystal_map.py
```diff
@@ -331,7 +331,9 @@ def orientations(self):
                 rotations = self.rotations[:, 0]
             else:
                 rotations = self.rotations
-            return Orientation(rotations).set_symmetry(phases[:].point_group)
+            o = Orientation(rotations)
+            o.symmetry = phases[:].point_group
+            return o.compute_symmetry_reduced_orientations()
```
hakonanes, line 336: I suggest changing the behaviour of `CrystalMap.orientations` and not returning the orientations with the smallest rotation angle, and leaving that up to the user to do. This change should be added to the changelog.

```suggestion
            orientations = Orientation(rotations)
            orientations.symmetry = phases[:].point_group
            return orientations
```

orix/quaternion/orientation.py
```diff
@@ -127,6 +127,15 @@ def symmetry(self):
...
         return self._symmetry
 
+    @symmetry.setter
+    def symmetry(self, values):
+        err_message = "Values must be a tuple of two Symmetry objects."
+        if not isinstance(values, (list, tuple)):
+            raise TypeError(err_message)
+        if len(values) != 2 or not all(isinstance(s, Symmetry) for s in values):
+            raise ValueError(err_message)
+        self._symmetry = tuple(values)
```
hakonanes, line 137:
```suggestion
    def symmetry(self, value):
        err_message = "Value must be a tuple with two Symmetry instances."
        if not isinstance(value, (list, tuple)):
            raise TypeError(err_message)
        if len(value) != 2 or not all(isinstance(s, Symmetry) for s in value):
            raise ValueError(err_message)
        self._symmetry = tuple(value)
```

orix/quaternion/orientation.py
```diff
@@ -153,6 +162,11 @@ def equivalent(self, grain_exchange=False):
         equivalent = Gr.outer(orientations.outer(Gl))
         return self.__class__(equivalent).flatten()
 
+    @deprecated(
+        since="0.7",
+        alternative="orix.quaternion.Misorientation.compute_symmetry_reduced_orientations",
+        removal="0.8",
+    )
```
hakonanes, line 169:
```suggestion
    @deprecated(
        since="0.8",
        alternative="orix.quaternion.Misorientation.assign_smallest_angle",
        removal="0.9",
    )
```

orix/quaternion/orientation.py
```diff
@@ -178,6 +192,31 @@ def set_symmetry(self, Gl, Gr, verbose=False):
         [[-0.7071  0.7071  0.      0.    ]
         [ 0.      1.      0.      0.    ]]
         """
+        misori = self.__class__(self.data)
+        misori.symmetry = (Gl, Gr)
+        return misori.compute_symmetry_reduced_orientations()
```
hakonanes, line 197:
```suggestion
        return misori.assign_smallest_angle()
```

orix/quaternion/orientation.py
```diff
@@ -178,6 +192,31 @@ def set_symmetry(self, Gl, Gr, verbose=False):
...
+        >>> from orix.quaternion.symmetry import C4, C2
+        >>> data = np.array([[0.5, 0.5, 0.5, 0.5], [0, 1, 0, 0]])
+        >>> m = Misorientation(data)
+        >>> m.symmetry = (C4, C2)
+        >>> m.compute_symmetry_reduced_orientations()
+        Misorientation (2,) 4, 2
+        [[-0.7071  0.7071  0.      0.    ]
+        [ 0.      1.      0.      0.    ]]
+        """
+        Gl, Gr = self._symmetry
```
hakonanes, line 219: Following the docstring, I find `assign_smallest_angle()` a more informative name, what do you think @harripj?

What do you think @pc494? You've touched this part of the code the most of us.

I've also suggested small changes to the docstring example, in anticipation of us testing these as part of CI (using `pytest --doctest`).

```suggestion
    def assign_smallest_angle(self, verbose=False):
        """Computes equivalent transformations which have the smallest
        angle of rotation and assigns these in-place.

        Returns
        -------
        Misorientation
            A new misorientation instance with the smallest angle of rotation.

        Examples
        --------
        >>> import numpy as np
        >>> from orix.quaternion import Misorientation, symmetry
        >>> data = np.array([[0.5, 0.5, 0.5, 0.5], [0, 1, 0, 0]])
        >>> m = Misorientation(data)
        >>> m.symmetry = (symmetry.C4, symmetry.C2)
        >>> m.assign_smallest_angle()
        Misorientation (2,) 4, 2
        [[-0.7071  0.7071  0.      0.    ]
        [ 0.      1.      0.      0.    ]]
        """
        Gl, Gr = self._symmetry
```

orix/quaternion/orientation.py
```diff
@@ -361,16 +402,30 @@ def symmetry(self):
         """Symmetry."""
         return self._symmetry[1]
 
+    @symmetry.setter
+    def symmetry(self, value):
+        if not isinstance(value, Symmetry):
+            raise TypeError(
+                "Value must be an instance of orix.quaternion.symmetry.Symmetry."
```
hakonanes, line 409: The `Symmetry` class is reachable from the `orix.quaternion` module, and should be imported from here.

```suggestion
                "Value must be an instance of orix.quaternion.Symmetry."
```

orix/quaternion/orientation.py
```diff
@@ -408,7 +464,7 @@ def from_euler(
         """
         o = super().from_euler(euler=euler, convention=convention, direction=direction)
         if symmetry:
-            o = o.set_symmetry(symmetry)
+            o.symmetry = symmetry
```
hakonanes, line 467: The fact that these methods don't return orientations with the smallest angle of rotation under symmetry should be added to the changelog!

orix/quaternion/orientation.py
```diff
@@ -563,6 +619,11 @@ def get_distance_matrix(self, lazy=False, chunk_size=20, progressbar=True):
 
         return Scalar(angles)
 
+    @deprecated(
+        since="0.7",
+        alternative="orix.quaternion.Orientation.compute_symmetry_reduced_orientations",
+        removal="0.8",
+    )
```
hakonanes, line 626:
```suggestion
    @deprecated(
        since="0.8",
        alternative="orix.quaternion.Orientation.assign_smallest_angle",
        removal="0.9",
    )
```

orix/tests/quaternion/test_orientation.py
```diff
@@ -57,7 +57,9 @@ def orientation(request):
     indirect=["orientation"],
 )
 def test_set_symmetry(orientation, symmetry, expected):
-    o = orientation.set_symmetry(symmetry)
+    o = Orientation(orientation.data)
+    o.symmetry = symmetry
```
hakonanes, line 61: We should allow passing the symmetry upon initialization, no? This should be documented in the `Orientation.__init__(self, data, symmetry=None)` docstring. *C1* should be default when `symmetry` isn't given, to give same result as before.

```suggestion
    o = Orientation(orientation.data, symmetry)
```

hakonanes: This should be done for `Misorientation.__init__(self, data, symmetry=None)` as well, with `symmetry` being a tuple with `Symmetry` instances.

orix/tests/quaternion/test_orientation.py
```diff
@@ -168,22 +177,89 @@ def test_transpose_3d(shape, expected_shape, axes):
...
+        # not Symmetry
+        o.symmetry = (1, 2)
+
+    with pytest.raises(ValueError, match="Values must be a tuple"):
+        # not Symmetry
+        o.symmetry = (C1, 2)
+
+    with pytest.raises(TypeError, match="Values must be a tuple"):
+        # not list or tuple
+        o.symmetry = 1
```
hakonanes, line 235: Shorten with `pytest.mark.parametrize`?

```suggestion
@pytest.mark.parametrize("error_type, value", [(ValueError, (1, 2)), (ValueError, (C1, 2)), (TypeError, 1)])
def test_symmetry_property_wrong_type_misorientation(error_type, value):
    mori = Misorientation.random((3, 2))
    with pytest.raises(error_type, match="Values must be a tuple"):
        mori.symmetry = value
```

orix/tests/test_miller.py
```diff
@@ -221,7 +221,8 @@ def test_unique(self):
 
     def test_multiply_orientation(self):
         o = Orientation.from_euler(np.deg2rad([45, 0, 0]))
-        o = o.set_symmetry(CUBIC_PHASE.point_group)
+        o.symmetry = CUBIC_PHASE.point_group
+        o = o.compute_symmetry_reduced_orientations()
```
hakonanes, line 225: Do we need to assign smallest rotation angle for this test to pass? If not, I'd skip it!

hakonanes (changes requested):
This reflects nicely what I had in mind after our discussion in #235! I have several suggestions for improvements, and also some suggestions for further changes, but we should leave these for other PRs.

**Before implementing the suggested changes though, I would like to get @pc494 broad impressions on this change, as he is most familiar with this part of the code among us.**

Also, we need to check the notebooks that they produce expected results. E.g. plotting of (mis)orientations inside fundamental zones. This was the motivation for always assigning (mis)orientations the smallest rotation angle, since they then fall inside the zones. I'll do this and suggest further changes where necessary.

orix/tests/test_miller.py
harripj: The test looks like it passes without assigning smallest angle! Updating this.

orix/crystal_map/crystal_map.py
harripj: Makes sense to me too, I will update the tests to reflect this.

orix/tests/quaternion/test_orientation.py
hakonanes: I suggest leaving this for another PR!

harripj: I agree.

orix/quaternion/orientation.py
harripj: The name of this function is certainly up for debate!

@hakonanes I think that the `smallest_angle` part is very informative, but to me `assign` indicates assign in-place rather than return a new instance (which is what the function is currently doing). I would prefer `compute` or `calculate` to this end. What do you think?

orix/tests/quaternion/test_orientation.py
harripj: Created issue #241 to discuss this.

orix/quaternion/orientation.py
hakonanes: I agree that by assign we mean in-place, which is what the docstring says, but actually doesn't do! Another place where this docstring could be improved, good catch indeed.

I think @pc494's suggestion of `map_into_symmetry_reduced_zone()` (or perhaps `map_into_fundamental_zone()`) is even better, what do you think?

harripj: Yes, well spotted! I will update the docstring here.

I think that both of these suggestions are informative and make sense. `map_into_symmetry_reduced_zone()` is easy to understand. `map_into_fundamental_zone()` I think we have to be clear that this is what is actually happening, and what would distinguish `zone` from `region` (as in `OrientationRegion`)?

hakonanes: Let's use map_to_symmetry_reduced_zone() then!

CHANGELOG.rst
```diff
@@ -21,18 +21,25 @@ Added
...
+- `CrystalMap.orientations` no longer returns smallest angle rotation.
 
 Deprecated
 ----------
 - The `data_dim` attribute of Object3d and all derived classes is deprecated from 0.8
   and will be removed in 0.9. Use `ndim` instead.
-
+- Setting (Mis)Orientation symmetry via `set_symmetry()` is deprecated in 0.8, in favour of
+  setting it directly via a `symmetry.setter`, and will be removed in 0.9. Use
+  `assign_smallest_angle()` instead.
```
hakonanes, line 41:
```suggestion
  `map_into_symmetry_reduced_zone()` instead.
```

orix/quaternion/orientation.py
```diff
@@ -153,6 +162,11 @@ def equivalent(self, grain_exchange=False):
         equivalent = Gr.outer(orientations.outer(Gl))
         return self.__class__(equivalent).flatten()
 
+    @deprecated(
+        since="0.8",
+        alternative="orix.quaternion.Misorientation.assign_smallest_angle",
```
hakonanes, line 167:
```suggestion
        alternative="orix.quaternion.Misorientation.map_into_symmetry_reduced_zone",
```

orix/quaternion/orientation.py
```diff
@@ -563,6 +617,11 @@ def get_distance_matrix(self, lazy=False, chunk_size=20, progressbar=True):
 
         return Scalar(angles)
 
+    @deprecated(
+        since="0.8",
+        alternative="orix.quaternion.Orientation.assign_smallest_angle",
```
hakonanes, line 622:
```suggestion
        alternative="orix.quaternion.Orientation.map_into_symmetry_reduced_zone",
```

orix/tests/quaternion/test_orientation.py
```diff
@@ -168,22 +177,85 @@ def test_transpose_3d(shape, expected_shape, axes):
...
+    o1.symmetry = Oh
+    o1 = o1.map_into_symmetry_reduced_zone()
     o2 = o1.transpose()
 
     assert o1.symmetry == o2.symmetry
 
 
+def test_symmetry_property_orientation():
+    o = Orientation.random((3, 2))
+
```
hakonanes, line 190: I would remove these blank lines, as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#blank-lines).

orix/tests/quaternion/test_orientation.py
```diff
@@ -292,3 +377,21 @@ def test_scatter(self, orientation, pure_misorientation):
...
 
         plt.close("all")
+
+
+def test_set_symmetry_deprecation_warning_orientation():
+    o = Orientation.random((3, 2))
+    with pytest.warns(
+        np.VisibleDeprecationWarning,
+        match="Function `set_symmetry()",
+    ):  # can only get this to work properly with one backtick...
```
hakonanes, line 387: Yeah, according to the [pytest documentation](https://docs.pytest.org/en/6.2.x/warnings.html#warns), `match` accepts text or a regex (with the string prepended by an `r`). I'd just leave this as it is, but remove the comment.

hakonanes (commented):
This looks good, and I approve the changes that are in this PR now.

Before I approve the changes for merging, I'd just like to test plotting of orientations within the symmetry reduced zone (axis/angle and Rodrigues), and seeing that this works as before. We might have to call `map_into_symmetry_reduced_zone()` on the (mis)orientations *within* the plotting methods to achieve the desired result. Will report back in this PR after I've checked.

hakonanes:
I might push fixes to the failing tests to your branch, if that's OK @harripj.

harripj:
@hakonanes Absolutely.

hakonanes:
As expected, we need to call `map_into_symmetry_reduced_zone()` when plotting (mis)orientations in symmetry reduced zones. This code

```python
ori = Orientation.random(1000)
ori.symmetry = symmetry.Oh
ori.scatter()
```

produces this plot

orientation_scatter_outside

hakonanes:
We need to add a call to `map_into_symmetry_reduced_zone()` in [`RotationPlot.transform()`](https://github.com/pyxem/orix/blob/master/orix/plot/rotation_plot.py#L33) method for (mis)orientations to retain previous behaviour:

```python
    def transform(self, xs):
        from orix.quaternion import Misorientation, Rotation

        if isinstance(xs, Misorientation):  # Also true for Orientation instances 
            xs = xs.map_into_symmetry_reduced_zone()
# As before below this
```

However, this means that `map_into_symmetry_reduced_zone()` is called with *every* call to `RotationPlot.transform()` with (mis)orientations, in contrast to only *once* in the previous implementation. This is a drawback when a workflow is mainly built out of many plots of (mis)orientations in fundamental zones.

The only solutions I can think of are:
1. Add a boolean parameter to `Misorientation.scatter()` which carry over to `RotationPlot.scatter()` and `RotationPlot.transform()` which allows us to bypass the call to `map_into_symmetry_reduced_zone()`. The user should only bypass this call when they themselves call `map_into_symmetry_reduced_zone()` before plotting.
2. Somehow check whether all (mis)orientation angles (`Orientation.angle.data.max()`) are within the fundamental zone, and only call `map_into_symmetry_reduced_zone()` when this is not the case. This requires adding something like the following to `RotationPlot.transform()`:

```python
from orix.quaternion import Misorientation, OrientationRegion, Rotation

if isinstance(xs, Misorientation):
    if isinstance(xs.symmetry, tuple):
        fr = OrientationRegion.from_symmetry(xs.symmetry[0], xs.symmetry[1])
    else:
        fr = OrientationRegion.from_symmetry(xs.symmetry)
    region_edge = fr.get_plot_data().angle.data.max()
    if xs.angle.data.max() > region_edge:
        xs = xs.map_into_symmetry_reduced_zone()
# The rest as before
```

The added computation of the orientation region and get_plot_data()  seems to be fairly cheap, 2.05 ms on my machine with `Oh` symmetry.

What do you think of these options, @harripj? And this would also be a good time to ask @pc494 what he thinks of these changes to fundamental zone plotting.

harripj:
@hakonanes well spotted. I think that 2ms is not a problem when calling a plot function, but on my machine I am getting ~50 ms:
```
%timeit fr = OrientationRegion.from_symmetry(*ori._symmetry)
47 ms ± 400 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit region_edge = fr.get_plot_data().angle.data.max()
53.3 ms ± 807 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

I would tend toward the first option though, providing a `bool` to pass to `scatter` etc. which allows for the user to use the functionality (by default) or not and this may make it easier to debug too.

A third option would be an internal flag within `Misorientation` classes, something like `_IS_IN_SYMMETRY_REDUCED_ZONE` which is `False` by default and is only set as `True` when the new instance is created within the `map_into_symmetry_reduced_zone` method. This would be very cheap computationally, but would limit the solution to `Misorientation` classes. Is there a need to extend this to 'Rotation' and 'Quaternion' classes too?

hakonanes:
> @hakonanes well spotted. I think that 2ms is not a problem when calling a plot function, but on my machine I am getting ~50 ms:

Hah, I believe I reported the uncertainty, not the mean. Sorry! However a 50 ms penalty isn't so bad either in my opinion.

> I would tend toward the first option though, providing a `bool` to pass to `scatter` etc. which allows for the user to use the functionality (by default) or not and this may make it easier to debug too.

Thanks, that's good to know. It's important to note that solution 1. assumes that the user has either read the API reference for the `scatter()` method or `RotationPlot` or a user guide (where we have to document this optional parameter). I prefer solution 2. since I cannot think of a state of (mis)orientations where this wouldn't work. We should still document the fact that if the user reduces (mis)orientations to the fundamental zone themselves, plotting will be faster (although we will *always* perform that check).

> A third option would be an internal flag within `Misorientation` classes, something like `_IS_IN_SYMMETRY_REDUCED_ZONE` which is `False` by default and is only set as `True` when the new instance is created within the `map_into_symmetry_reduced_zone` method.

Keeping track of such "state" parameters is difficult, and I would advice against it. For example, since all 3D objects in orix are mutable, we would have to update that parameter every time an element in an instance is changed. What we could do instead is to return whether the check I proposed in solution 2. is true or not as that parameter. Then we would never have to set it.

> This would be very cheap computationally, but would limit the solution to `Misorientation` classes. Is there a need to extend this to `Rotation` and `Quaternion` classes too?

Only `Misorientation` and `Orientation` classes have a notion of symmetry, so we don't have to worry about `Rotation` or `Quaternion` :)

harripj:
> Thanks, that's good to know. It's important to note that solution 1. assumes that the user has either read the API reference for the scatter() method or RotationPlot or a user guide (where we have to document this optional parameter). I prefer solution 2. since I cannot think of a state of (mis)orientations where this wouldn't work. We should still document the fact that if the user reduces (mis)orientations to the fundamental zone themselves, plotting will be faster (although we will always perform that check).

Good points and yes, case 2 will be more user friendly.

> Keeping track of such "state" parameters is difficult, and I would advice against it. For example, since all 3D objects in orix are mutable, we would have to update that parameter every time an element in an instance is changed. What we could do instead is to return whether the check I proposed in solution 2. is true or not as that parameter. Then we would never have to set it.

I hadn't thought of this at the time and it would certainly make it more difficult to implement. I like the idea of using your suggestion as a check. Perhaps `OrientationRegion` can be an optional argument to avoid its recalculation?

hakonanes:
> Perhaps OrientationRegion can be an optional argument to avoid its recalculation?

Yes, we could provide a function accepting an `OrientationRegion`, instead of providing an attribute where the region is created every time the attribute is accessed.

Ultimately, I cannot say which solution is best until @pc494 has reviewed that part of this PR. I know he's quite busy these days, so we just have to be patient :)

pc494:
I'll be on it this weekend! 

NB: turns out that github tagging doesn't give me an extra ping

hakonanes:
Great, thanks! Good to know, will send you an email next time.

pc494:
Apologies for the delay, unfortunately I'm going to be a dissenting voice on this.

If the user has an option to set a symmetry I think they would anticipate that the values of the quaternions left can be used to ID if Orientations/Misorientations are different. Otherwise, what exactly is the symmetry doing? This implementation would mean I could have two symmetry identical rotations, I could set the symmetry and I would they would still have different numerical values. If the objects represent the same physical transformation surely we want them to have the same numerical value. Also doing it at the start saves you a lot of fiddling around later (as the comments on this PR suggest).

I think the plots in https://github.com/pyxem/orix/issues/233#issuecomment-958779927 suggest 1 (or more) bugs. I'll have a quick look at the simplest case and feedback.

hakonanes:
> If the user has an option to set a symmetry I think they would anticipate that the values of the quaternions left can be used to ID if Orientations/Misorientations are different. Otherwise, what exactly is the symmetry doing?

Of course symmetry should be used in operations, like in class methods such as `angle_with()`. That is what I expect when using MTEX, that subsequent operations will respect symmetry; but I don't inspect the Euler angles in the `orientation` instance representation before and after setting the symmetry. Orientations don't change underlying quaternions in MTEX after setting the symmetry `ori.CS` (the representation, showing Euler angles, doesn't change either).

> Also doing it at the start saves you a lot of fiddling around later (as the comments on this PR suggest).

A problem is that `set_symmetry()` is a comparatively demanding computational task, and it is called many times internally when performing operations on (mis)orientations (just look at this PR), not just upon creation.

> I think the plots in [#233 (comment)](https://github.com/pyxem/orix/issues/233#issuecomment-958779927) suggest 1 (or more) bugs. I'll have a quick look at the simplest case and feedback.
> 
> EDIT: Ignore my last comment, I think the two plots in the linked comment describe different situations. The top/left case are figures for a no symmetry case while the bottom/right case are for an object that has symmetry.

It should form a complete track in the IPF with crystal symmetry, as the boundaries are symmetry axes, so it must be a bug, right? Either in `master` or in #235.

___

This PR would make setting orientation symmetries in orix more like how it's done in MTEX. This would make it simpler to compare results to MTEX, and also to draw inspiration from MTEX' code. It doesn't mean it is the correct way to do it, though.

pc494:
> > If the user has an option to set a symmetry I think they would anticipate that the values of the quaternions left can be used to ID if Orientations/Misorientations are different. Otherwise, what exactly is the symmetry doing?
> 
> Of course symmetry should be used in operations, like in class methods such as `angle_with()`. That is what I expect when using MTEX, that subsequent operations will respect symmetry; but I don't inspect the Euler angles in the `orientation` instance representation before and after setting the symmetry. Orientations don't change underlying quaternions in MTEX after setting the symmetry `ori.CS` (the representation, showing Euler angles, doesn't change either).
> 
> > Also doing it at the start saves you a lot of fiddling around later (as the comments on this PR suggest).
> 
> A problem is that `set_symmetry()` is a comparatively demanding computational task, and it is called many times internally when performing operations on (mis)orientations (just look at this PR), not just upon creation.

Okay, I think you've convinced me. Still some thing I'm unsure about but I think this is at least one question answered. The functionality set symmetry used to perform should be renamed to something more explicit.

harripj:
> It should form a complete track in the IPF with crystal symmetry, as the boundaries are symmetry axes, so it must be a bug, right? Either in master or in #235.

Yes there must be a bug somewhere, although I am not sure where at the moment! If we can find this bug then maybe this conversation becomes easier.

> The functionality set symmetry used to perform should be renamed to something more explicit.

Perhaps the naming is part of the problem. In a Pythonic way `set_symmetry` reads to me as an assignment, eg. `.symmetry =`, whereas it is currently doing more than this.

hakonanes:
> The functionality set symmetry used to perform should be renamed to something more explicit.

This is what this PR addresses by introducing `Misorientation.map_into_symmetry_reduced_zone()` inplace of `set_symmetry()`, and deprecating the latter.

As I see it, the only outstanding thing in this PR as it is now is to resolve the questions I raised in https://github.com/pyxem/orix/pull/240#issuecomment-970143921.

harripj:
> This is what this PR addresses by introducing Misorientation.map_into_symmetry_reduced_zone() inplace of set_symmetry(), and deprecating the latter.

Absolutely.

@hakonanes I have implemented path 2 in https://github.com/pyxem/orix/pull/240#issuecomment-970143921 and allowed the `fundamental_region` to be passed into `RotationPlot.transform` to avoid duplication in the case of `Misorientation.scatter`.

pc494:
Am I correct in thinking this will be good for review fairly soon? :D

harripj:
@pc494 yep, I think it's good to go!

CHANGELOG.rst
```diff
@@ -21,18 +21,25 @@ Added
...
 - `ndim` attribute to Object3d and derived classes which returns number of navigation
   dimensions.
+- Setting the symmetry of a (Mis)Orientation via a `symmetry.setter`.
 
 Changed
 -------
 - `StereographicPlot` doesn't use Matplotlib's `transforms` framework anymore, and
   (X, Y) replaces (azimuth, polar) as internal coordinates.
 - Renamed `Symmetry` method `fundamental_sector()` to `fundamental_zone()`.
+- `Orientation` class methods `from_euler`, `from_matrix`, and `from_neo_euler` no longer
```
pc494, line 31: Would like to adjust the wording here

orix/quaternion/orientation.py
```diff
@@ -127,6 +127,15 @@ def symmetry(self):
         """Tuple of :class:`~orix.quaternion.Symmetry`."""
         return self._symmetry
 
+    @symmetry.setter
+    def symmetry(self, value):
+        err_message = "Value must be a tuple with two Symmetry instances."
```
pc494, line 132: I see why you've used this pattern but I would prefer to have the two error messages made explicitly where they are called.

orix/quaternion/orientation.py
```diff
@@ -361,16 +402,28 @@ def symmetry(self):
...
         """Unit orientations."""
-        return super().unit.set_symmetry(self.symmetry)
+        o = super().unit
+        o.symmetry = self.symmetry
+        return o
 
     def __invert__(self):
-        return super().__invert__().set_symmetry(self.symmetry)
+        o = super().__invert__()
+        o.symmetry = self.symmetry
```
pc494, line 420: eg. this (comment 3)

pc494 (commented):
I'm basically happy with this, I have two small changes (that I'm happy to put in myself) which I've made as comments, if you (@harripj) could just agree to them that would be great.

My one question is how did you check that you had got all calls to .`set_symmetry` in when you changed it in the package (example in 'comment 3')?

harripj:
> My one question is how did you check that you had got all calls to .set_symmetry in when you changed it in the package (example in 'comment 3')?

I searched `.set_symmetry` globally in the repository and changed each one to the equivalent calculation (which is now `o.symmetry = ..., o.map_into_symmetry_reduced_zone()`). This is apart from `CrystalMap` as discussed here https://github.com/pyxem/orix/pull/240#discussion_r746697369, and the class methods and both are documented in the Changelog here https://github.com/pyxem/orix/pull/240/commits/5062b738aff8e5040eb6fba3c2f073ac44d27969 and here https://github.com/pyxem/orix/pull/240/commits/5347279b7c27d6caeeca7367a705e42a0fcad9db.

A global search in the repository now only shows 4 instances of `set_symmetry` in the working code, specifically in `orientation.py`, all of which are wrapped within the deprecation warnings.

pc494:
> > My one question is how did you check that you had got all calls to .set_symmetry in when you changed it in the package (example in 'comment 3')?
> 
> I searched `.set_symmetry` globally in the repository and changed each one to the equivalent calculation (which is now `o.symmetry = ..., o.map_into_symmetry_reduced_zone()`). This is apart from `CrystalMap` as discussed here [#240 (comment)](https://github.com/pyxem/orix/pull/240#discussion_r746697369), and the class methods and both are documented in the Changelog here [5062b73](https://github.com/pyxem/orix/commit/5062b738aff8e5040eb6fba3c2f073ac44d27969) and here [5347279](https://github.com/pyxem/orix/commit/5347279b7c27d6caeeca7367a705e42a0fcad9db).
> 
> A global search in the repository now only shows 4 instances of `set_symmetry` in the working code, specifically in `orientation.py`, all of which are wrapped within the deprecation warnings.

That seems robust.

orix/tests/quaternion/test_orientation.py
```diff
@@ -163,27 +171,82 @@ def test_transpose_2d():
...
+def test_symmetry_property_orientation_data():
+    """Test that data remains unchanged after setting symmetry property."""
+    o = Orientation.random((3, 2))
+    d1 = o.data.copy()
+    o.symmetry = Oh
+    assert np.allclose(o.data, d1)
+
+
+def test_symmetry_property_misorientation():
+    o = Misorientation.random((3, 2))
```
hakonanes, line 202: Name the variable mori or m?

orix/tests/quaternion/test_orientation.py
```diff
@@ -163,27 +171,82 @@ def test_transpose_2d():
...
+    mori = Misorientation.random((3, 2))
+    with pytest.raises(error_type, match="Value must be a tuple"):
+        mori.symmetry = value
+
+
+@pytest.mark.parametrize(
+    "error_type, value",
+    [
+        (ValueError, (C1,)),
+        (ValueError, (C1, C2, C1)),
```
hakonanes, line 227: Make list one line by dropping the final comma in this line.

orix/tests/test_crystal_map.py
```diff
@@ -655,13 +658,15 @@ def test_get_orientations_array(self, crystal_map_input, phase_list):
         o1 = xmap1.get_map_data("orientations")
         o2 = xmap2.get_map_data("orientations")
 
-        expected_o1 = xmap1.orientations.to_euler()
+        expected_o1 = xmap1.orientations
```
hakonanes, line 661: Why two lines? I would revert to one line, also in the change below.

hakonanes (changes requested):
I've suggested some doc changes and some minor simplifications of code, otherwise I think this change is robust and the right way forward.

orix/tests/quaternion/test_orientation.py
```diff
@@ -69,7 +71,9 @@ def test_set_symmetry(orientation, symmetry, expected):
 def test_orientation_persistence(symmetry, vector):
     v = symmetry.outer(vector).flatten()
     o = Orientation.random()
-    oc = o.set_symmetry(symmetry)
+    oc = Orientation(o.data)
```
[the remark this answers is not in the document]
harripj: This would be a nice improvement.

orix/tests/quaternion/test_orientation.py
harripj: Good spot! Updated.

orix/tests/test_crystal_map.py
harripj: Not sure why this happened... updated!

orix/plot/rotation_plot.py
```diff
@@ -30,8 +30,25 @@ class RotationPlot(Axes3D):
     name = None
     transformation_class = None
 
-    def transform(self, xs):
-        from orix.quaternion.rotation import Rotation
+    def transform(self, xs, fundamental_region=None):
```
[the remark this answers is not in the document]
harripj: No problem with this. `_get_region_extent` and `_correct_aspect_ratio` also within `plot/rotation_plot.py` use `fundamental_region`, perhaps they should all be consistent? I am not sure whether there is a difference between region and zone in orix, would you be be able to shed some light on this if possible?

orix/plot/rotation_plot.py
```diff
@@ -30,8 +30,25 @@ class RotationPlot(Axes3D):
...
+
+        # check whether xs is in fundamental region
+        if isinstance(xs, Misorientation):
+            if fundamental_region is None:
+                if isinstance(xs.symmetry, tuple):
+                    fundamental_region = OrientationRegion.from_symmetry(
+                        xs.symmetry[0], xs.symmetry[1]
+                    )
+                else:
+                    fundamental_region = OrientationRegion.from_symmetry(xs.symmetry)
```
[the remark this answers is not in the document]
harripj: I think for `Orientation` this won't work as `xs.symmetry` returns just a single `Symmetry` which is iterable and then gets unpacked into the tuple. `xs._symmetry` should return the correct tuple though, eg. `(C1, Th)`? This will, however, be inverted for the default arguments for `OrientationRegion.from_symmetry(s1, s2=C1)`.

orix/quaternion/orientation.py
```diff
@@ -178,6 +192,31 @@ def set_symmetry(self, Gl, Gr, verbose=False):
...
+        misori.symmetry = (Gl, Gr)
+        return misori.compute_symmetry_reduced_orientations()
+
+    def compute_symmetry_reduced_orientations(self, verbose=False):
+        """Computes equivalent transformations which have the smallest
+        angle of rotation and return these as a new Misorientation object.
+
+        Returns
+        -------
+        Misorientation
```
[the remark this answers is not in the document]
hakonanes: `Orientation.map_into_symmetry_reduced_zone()` must return an `Orientation` instance.

We have a separate docstring for `Orientation.set_symmetry()` because we *have* to override `Misorientation.set_symmetry()` anyway due to the different number of parameters it accepts. With `map_into_symmetry_reduced_zone()` we don't have to do this, and I think we should keep it this way to not increase the number of docstrings to maintain. The docstring is inconsistent with `Orientation`, but I think it should be clear *enough* for the user. What do you think?

orix/quaternion/orientation.py
hakonanes: I believe that is what @pc494 means, yes.

orix/plot/rotation_plot.py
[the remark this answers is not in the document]
harripj: I have just noticed that whilst `(Mis)Orientation` DS use fundamental zone, within the function body they use `fundamental_region`. This leaves is us with the case in `RotationPlot.scatter` where `fundamental_zone=fundamental_region`, which is perhaps not ideal. I can remove the kwarg for now, but perhaps this should be cleared up in the future?

orix/plot/rotation_plot.py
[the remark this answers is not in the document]
hakonanes: I've tested that this works. It works because `tuple(Orientation.symmetry)` becomes a 1-tuple which is passed to `s1`.

orix/plot/rotation_plot.py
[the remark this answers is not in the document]
hakonanes: Ouch, my bad... I'd prefer to rename everything to `zone` in `RotationPlot`. This does not break any API, since the methods you point to are private.

hakonanes:
One of the user guide notebooks have cell output stored and so must be re-uploaded using the new API. The other notebooks using `set_symmetry()` will after this PR show the deprecation warning, but still work as expected. I'll deal with this in a separete PR #246.

orix/quaternion/orientation.py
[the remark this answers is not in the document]
harripj: Sounds okay to me. I think it should be okay for the user.

orix/plot/rotation_plot.py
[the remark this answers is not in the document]
harripj: No sweat, and leave as `fundamental_region` in `Orientation.scatter` or change to `zone` as well?

orix/quaternion/orientation.py
harripj: Okay, I have assigned the errors in place and they read "Value must be a 2-tuple of Symmetry objects."

orix/plot/rotation_plot.py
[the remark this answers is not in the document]
hakonanes: The variables in the code can be renamed to `fundamental_zone` as well!

orix/plot/rotation_plot.py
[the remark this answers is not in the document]
harripj: This works for `C1` for me, but not other `Symmetry` with more than one transformation:
```python
from orix.quaternion.symmetry import Th, C1

o1 = Orientation.random()
o1.symmetry = C1
tuple(o1.symmetry)
>>> (Symmetry (1,)
    [[1. 0. 0. 0.]],)  # this is okay

o2 = Orientation.random()
o2.symmetry = Th
tuple(o2.symmetry)
>>> (Symmetry (1,)
    [[1. 0. 0. 0.]], Symmetry (1,)
    [[0. 0. 0. 1.]], Symmetry (1,)
    [[0.5 0.5 0.5 0.5]], Symmetry (1,)
    [[-0.5 -0.5  0.5  0.5]], Symmetry (1,)
    [[-0.5  0.5 -0.5  0.5]], Symmetry (1,)...  # I don't think this will work

tuple(o2._symmetry)
>>> (Symmetry (1,) 1
    [[1. 0. 0. 0.]], Symmetry (24,) m-3
    [[ 1.   0.   0.   0. ]
     [ 0.   0.   0.   1. ]
     [ 0.5  0.5  0.5  0.5]
     [-0.5 -0.5  0.5  0.5]...)  # this works but C1 is now tuple[0]
```

The last case works but `C1` is now tuple[0], which is the wrong argument order for `OrientationRegion.from_symmetry(s1, s2=C1)`. Do you foresee this being a problem?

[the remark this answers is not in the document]
hakonanes: You're right, I was too quick, sorry. This should work for all symmetry sizes, right?

```python
                symmetry = xs.symmetry
                if not isinstance(symmetry, tuple):
                    symmetry = tuple((symmetry,))
                fundamental_region = OrientationRegion.from_symmetry(*symmetry)
```

orix/plot/rotation_plot.py
```diff
@@ -30,8 +30,25 @@ class RotationPlot(Axes3D):
...
+                    )
+                else:
+                    fundamental_zone = OrientationRegion.from_symmetry(xs.symmetry)
+            # check fundamental_zone is properly defined
+            if not isinstance(fundamental_zone, OrientationRegion):
+                raise TypeError("Fundamental_zone is not an OrientationRegion object.")
+            region_edge = fundamental_zone.get_plot_data().angle.data.max()
+            # if xs is out of fundamental region, calculate symmetry reduction
+            if xs.angle.data.max() > region_edge:
+                xs = xs.map_into_symmetry_reduced_zone()
```
hakonanes, line 51: I have to correct myself here, this check is naive and won't work for all zones. We already have [a check for this](https://github.com/pyxem/orix/blob/73a8127b67a707bba5a423d9f18853e403055d06/orix/quaternion/orientation_region.py#L172-L182) implementing `<` for (Mis)Orientation (left) and OrientationRegion (right). Therefore the check must be

```python
import numpy as np
[...]
            # if xs is out of fundamental region, calculate symmetry reduction
            if not np.all(xs < fundamental_region):
                xs = xs.map_into_symmetry_reduced_zone()
```

orix/plot/rotation_plot.py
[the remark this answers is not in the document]
harripj: Yes this should work no problem. I will update the code!

orix/plot/rotation_plot.py
hakonanes: This check performs a `Quaternion(self).dot_outer(other.data)`, which shouldn't be too time consuming, since `self.data` contains 14 rotations at most for the point groups.

harripj: Makes sense. Do you want to commit this? Happy to do it also. Could also avoid the import with `(xs < fundamental_region).all()`.

orix/plot/rotation_plot.py
```diff
@@ -30,8 +30,23 @@ class RotationPlot(Axes3D):
...
+        if isinstance(xs, Misorientation):
+            if fundamental_zone is None:
+                symmetry = xs.symmetry
+                # Orientation.symmetry returns a Symmetry object not a tuple, so pack
+                if not isinstance(symmetry, tuple):
+                    symmetry = (symmetry,)
+                fundamental_zone = OrientationRegion.from_symmetry(*symmetry)
+            # check fundamental_zone is properly defined
+            if not isinstance(fundamental_zone, OrientationRegion):
+                raise TypeError("Fundamental_zone is not an OrientationRegion object.")
```
hakonanes, line 46:
```suggestion
                raise TypeError("fundamental_zone is not an OrientationRegion object.")
```

orix/plot/rotation_plot.py
```diff
@@ -30,8 +30,23 @@ class RotationPlot(Axes3D):
...
+            # if any in xs are out of fundamental_zone, calculate symmetry reduction
+            if not (xs < fundamental_zone).all():
+                xs = xs.map_into_symmetry_reduced_zone()
```
hakonanes, line 49: This snippet should test this bit

```python
>>> import numpy as np
>>> from orix.quaternion import Orientation, OrientationRegion, symmetry
>>> ori = Orientation((1, 0, 0, 0), (0.5, 0.5, 0.5, 0.5))
>>> ori.symmetry = symmetry.D6
>>> fz = OrientationRegion.from_symmetry(ori.symmetry)
>>> assert np.allclose(ori < fz, (True, False))
>>> ori.scatter()
```

I cannot get the positions of the scatter points from the returned figure (I tried looking around in `figure.axes[0].collections[1]` which is a `Path3DCollection` with the two scatter points), but couldn't find the 3D coordinates to test whether the last point is actually plotted inside the fundamental zone. But it is at least on my machine...

hakonanes (changes requested):
I've got two final comments, after that I'll approve and wait for Phillip to approve and merge!

orix/plot/rotation_plot.py
harripj: We can get them from the plot as:
```python
fig = ori.scatter(return_figure=True)
# pathCollection
pc = fig.axes[0].collections[1]
x, y, z = pc._offsets3d
vec = Vector3d(np.column_stack((x, y, z))
```
if that helps! Not sure how to go about testing whether vectors are within `OrientationRegion` though!

hakonanes: There you go, great! Then I'd write the test as

```python
>>> import numpy as np
>>> from orix.quaternion import Orientation, OrientationRegion, symmetry
>>> ori = Orientation((1, 0, 0, 0), (0.5, 0.5, 0.5, 0.5))
>>> ori.symmetry = symmetry.D6
>>> fz = OrientationRegion.from_symmetry(ori.symmetry)
>>> assert np.allclose(ori < fz, (True, False))
>>> fig = ori.scatter(return_figure=True)
>>> xyz_symmetry = fig.axes[0].collections[1]._offsets3d
>>> ori.symmetry = symmetry.C1
>>> fig2 = ori.scatter(return_figure=True)
>>> xyz = fig.axes[0].collections[1]._offsets3d
>>> assert not np.allclose(xyz_symmetry, xyz)
```

Here we've established that there's at least a change in where the points are plotted in the 3d axes.

pc494:
Approved - Merging.

Issue #241, opened 2021-11-10
Add symmetry argument to Orientation-based classes __init__


harripj:
As mentioned by @hakonanes in https://github.com/pyxem/orix/pull/240#discussion_r746712531, it would be useful to have a symmetry argument in the `__init__` methods of `Orientation` and `Misorientation` along the lines of `Misorientation.__init__(self, data, symmetry=None)`.

To be considered in conjunction with/after #240.

pc494:
+1 for this.

PR #251, opened 2021-12-14
Replace use of set_symmetry() with map_into_symmetry_reduced_zone() in user guide

### LINKED ISSUE #246: Update user guide notebooks with new API following #240
Update user guide notebooks with new API following #240

hakonanes:
* Use workflow introduced in #240 to set orientation symmetry in relevant user guide notebooks.
* Closes #246.
* Update paper reference on index page to the journal one (with DOI link), but keep arXiv link.

Changes in changed notebooks:
* Crystal map
* Clustering across fundamental region boundaries: Add a few more lines, but not many.
* Clustering orientations:
  * Add comment before call to `Orientation.from_euler()`: "Create an orientation instance from the Euler angles, defined in the MTEX convention as transformations from the crystal to the sample reference frame".
  * Add a few more lines here as well, but also removes one unnecessary call to `set_symmetry()` which was there previously.
  * Add a *622* inverse pole figure with the rotation axes of the orientation cluster means before re-centering. Is this interpretable? It seems like all axes are aligned along a fibre? 
* Uniform sampling orientation space:
  * Remove calls to `set_symmetry()`, not replaced, since `Orientation.scatter("rodrigues")` (or "axangle") performs mapping into fundamental zone if necessary upon plotting.
  * Replace rotation of x and z vector by orientations, showed in stereographic projection, with coverage of the x, y, and z inverse pole figures. Use the IPFs of the cubochoric sample as the notebook gallery image instead of the fundamental zones.

#### Progress of the PR

#### For reviewers


hakonanes:
Documentation to compare:
* Crystal map: [old](https://orix.readthedocs.io/en/stable/crystal_map.html#Orientation-data), [new](https://orix--251.org.readthedocs.build/en/251/crystal_map.html#Orientation-data)
* Clustering across fundamental region boundaries: [old](https://orix.readthedocs.io/en/stable/clustering_across_fundamental_region_boundaries.html#Generate-artificial-data), [new](https://orix--251.org.readthedocs.build/en/251/clustering_across_fundamental_region_boundaries.html#Generate-artificial-data)
* Clustering orientations: [old](https://orix.readthedocs.io/en/stable/clustering_orientations.html#Import-data), [new](https://orix--251.org.readthedocs.build/en/251/clustering_orientations.html#Import-data)
* Uniform sampling of orientation space: [old](https://orix.readthedocs.io/en/stable/uniform_sampling_of_orientation_space.html), [new](https://orix--251.org.readthedocs.build/en/251/uniform_sampling_of_orientation_space.html)

doc/clustering_orientations.ipynb
```diff
@@ -154,7 +158,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "ori = ori.set_symmetry(D6)"
+    "ori.symmetry = D6\n",
```
harripj: I don't think this line is needed as it is set in `from_euler`.

harripj (changes requested):
I like the switch between light and dark mode between old and new- makes it easy to compare- a nice touch 😄.

I like the addition of showing the sampling points in IPF too, nice work.

This PR has also convinced me, though, that we should have a symmetry argument for all `Orientation` class methods, as discussed in #241.

doc/clustering_orientations.ipynb
hakonanes: Unfortunately not, neither `reshape()` nor `__getitem__()` are overwritten in the `Orientation` class, so no other attributes than `data` are carried over. This is something we *should* do.

Instead, I'll remove passing `symmetry` to `from_euler()`, because it is lost before used.

harripj: Yes you are right, I agree that this should be done as actually it is quite unintuitive that the `symmetry` is lost just by reshaping or transposing the data, for example.

In the mean time I agree with your workaround- I think it will be less confusing that setting symmetry twice on the same variable.

In fact I think I will make a PR that includes setting `symmetry` in `Misorientation.__init__` and tries to keep the attribute through restructuring the various data where it makes sense.

hakonanes: Such a PR would be welcome, but we must be careful how we do it, since there are many methods and we should try to limit the necessary "duplicate" code lines and docstrings.

hakonanes:
> I like the switch between light and dark mode between old and new- makes it easy to compare- a nice touch smile.

This is one of the great features of the [Furo theme](https://pradyunsg.me/furo/): the light/dark theme depends on the OS choice, if available to the browser, otherwise light by default. It can further be toggled via the "sun/moon" icon in the upper right corner.

Thanks to you mentioning this, I added a comment to the contributing guide stating that it's important to make notebooks readable in both themes, and updated the "Clustering orientations" accordingly :)

> I like the addition of showing the sampling points in IPF too, nice work.

Yes, this shows the coverage of crystal directions sampled better than the stereographic projections without the crystal axes.

hakonanes:
Added a point to the OP under changes to the "Clustering orientations" notebook.

doc/clustering_orientations.ipynb
hakonanes: Resolve this conversation whenever you're happy :)

harripj (approved):

pc494:
Thanks @hakonanes for the PR, and @harripj for the review, merging.

hakonanes:
Thanks @pc494! FYI, as stated in the OP, I added an IPF plot of the un-centered rotation axes in the visualization part of the "Clustering orientations" notebook.

PR #252, opened 2021-12-15
Add symmetry arg to Orientation __init__ and override methods to maintain symmetry

harripj:
As mentioned in #241 we should introduce a `symmetry` argument to `Orientation.__init__` which would allow the user to set the symmetry of the instance at instantiation. This is already the case for the class methods, eg. `from_euler` etc., so this PR would just be bringing the syntax together.

A secondary goal of this PR is to maintain the `symmetry` of a `(Mis)Orientation` instance when operating on it, see  https://github.com/pyxem/orix/pull/251#discussion_r768914214 . This is done by overriding the methods in either class and docstrings will be inherited. Currently the following methods are overridden and maintain symmetry:
`Misorientation`:
- `__getitem__`
- `transpose`

`Orientation`:
- `__invert__`
- `__neg__`

This PR proposes that the following are added:
`Misorientation`:

- `reshape`
~~- `transpose`~~
- `flatten`
- `squeeze`

I also think that potentially `unique` should be overridden to maintain symmetry, and in fact possibly should account for symmetry, but would like to know what other people think about this.

#### Progress of the PR
- [x] Add symmetry arg to `Misorientation.__init__`
- [x] Override `Misorientation` `reshape` ~~`transpose`~~ `flatten` and `squeeze` methods to maintain symmetry
- [x] Update dead link in `contributing.rst`
- [x] Update changelog
- [x] Change notebooks to demonstrate this syntax

```python
>>> from orix.quaternion.symmetry import C2
>>> o = Orientation(data, symmetry=C2)
>>> o.symmetry
Symmetry (2,) 2
[[1. 0. 0. 0.]
 [0. 0. 0. 1.]]

>>> o = Orientation.random((5, 4))
>>> o.symmetry = C2
>>> o1 = o.reshape(4, 5)
>>> o1.symmetry
Symmetry (2,) 2
[[1. 0. 0. 0.]
 [0. 0. 0. 1.]]


```

#### For reviewers


hakonanes:
Sorely needed update, since the loss of symmetry has surprised myself many times before, and surely some other users!

Will review once you mark it as ready for review. I hope we can get it in for v0.8, that is merged within the week end?

> I also think that potentially unique should be overridden to maintain symmetry, and in fact possibly should account for symmetry, but would like to know what other people think about this.

This should be as simple as (1) mapping them into the fundamental zone, (2) finding the indices into the initial orientation array of the unique orientations within the zone using `Rotation.unique(return_index=True)`, and (3) returning the initial orientations in the unique indices, right? I suggest to leave this for its own PR.

harripj:
@pc494 no it's okay. Once this is merged we can make some small changes to the docs in #251, but this can be done after the fact in a new small PR!

harripj:
@pc494 I will do some small changes to the docs in #251 in this PR- it's all good!

harripj:
I have just updated the notebooks from #251 to use the `symmetry` arg in `Orientation.__init__()`. I would appreciate it if @hakonanes you could have a quick look over it to make sure everything still runs smoothly on your machine! (Everything looks fine on my machine but I don't have the data in "clustering_orientations.ipynb" to hand to check for sure).

@pc494 this means that it is good for review when you get a chance!

orix/quaternion/orientation.py
```diff
@@ -140,8 +145,28 @@ def __getitem__(self, key):
...
+
+    def squeeze(self):
+        m = super().squeeze()
+        m._symmetry = self._symmetry
+        return m
+
+    def transpose(self, *axes):
+        m = super().transpose(*axes)
+        m._symmetry = self._symmetry
+        return m
```
pc494, line 166: Each of these should have one line docstrings.

pc494 (changes requested):
A set of docstrings to add and then this is good to merge.

orix/quaternion/orientation.py
pc494: I had assumed overwriting methods meant you didn't inherit docstrings... Will double check and get back to you.

hakonanes: One way to find out is to add these to the list of `Orientation` methods in `reference.rst`, build the documentation locally, and inspect the `Orientation` section in the API reference in the local HTML files in the browser (as per the [contributing guide](https://orix.readthedocs.io/en/stable/contributing.html#build-and-write-documentation)).

pc494: My minimum working examples suggests overwritting methods loses the docstrings (I also checked the call to super didn't help, although that's not in the image).

do_docstrings_get_overwritten

hakonanes: Huh, this surprised me, as I thought base (parent) docstrings carried over when a new one wasn't written. Thank you for checking this, @pc494!

Then I think we should add docstrings for all methods, so they look nice in the API reference (and via IPython, Jupyter, etc.). We must be careful in the future to also update any "downstream" (read Orientation) docstrings when we update "upstream" (read Rotation) docstrings!

pc494: In general the best option is to overwrite as few methods as possible, but in this case the symmetry property does make that kind of difficult.

hakonanes: Yes, that's the motivation behind parts of this PR: to maintain any properties the user would assume are maintained, like the symmetry, when transposing, reshaping, flattening, etc.

hakonanes:
If it's OK with you @harripj, I'll push the updated "Clustering orientations" notebook with your addition to this PR.

hakonanes:
The other two affected notebooks "Crystal map" and "Clustering across fundamental region boundaries" look good in the [documentation](https://orix--252.org.readthedocs.build/en/252/index.html) built from this PR.

hakonanes:
Other methods which could (not saying *should* here) be overwritten:
* `Orientation.mean()`
* `Orientation.stack()` (when the instances to stack are orientations with the same symmetry)
* `Orientation.__mul__()` perhaps? When two orientations of the same symmetry are multiplied, the symmetry can be kept?

hakonanes:
I'll review this tomorrow.

orix/quaternion/orientation.py
harripj: So in the current state of the repo `Object3d.squeeze` has a docstring and `Misorientation.squeeze` does not. After building the docs the docstring for `Misorientation.squeeze` is inherited from `Object3d.squeeze`:
image

@pc494 I have tested your notebook and experience the same behaviour: `__doc__` remains `None` in the inherited method, but IPython correctly inherits the docstring:
image

(in fact in this case the docstring for `Orientation.flatten` is inherited from `Rotation.flatten` rather than `Object3d.flatten` and is why they are different, but the point is the same that IPython will inherit the docstrings. In fact if we decide to stick with inherited docstrings then we should remove the one for `Rotation.flatten`.)
So for the docs this isn't a problem, but the `__doc__` attribute isn't set for the inherited methods, and it is whether we mind this behaviour or not as it means less maintenance. From my POV I am always working with IPython and use the `?` operator regularly, whereas I never use `__doc__`, so I don't mind this but appreciate others may have different viewpoints. What do you all think?

harripj:
> If it's OK with you @harripj, I'll push the updated "Clustering orientations" notebook with your addition to this PR.

Yes of course, go ahead.

> Other methods which could (not saying should here) be overwritten:
> Orientation.mean()
> Orientation.stack() (when the instances to stack are orientations with the same symmetry)
> Orientation.__mul__() perhaps? When two orientations of the same symmetry are multiplied, the symmetry can be kept?

I was thinking about `stack` for sure, it makes sense to me, but only if _all_ of the inputs have the same symmetry. `mean` also makes sense, but `__mul__` seems to be more complex and I would have to do some more thinking/reading on it.

pc494:
I think the approach in #254 would resolve most of these issues (docstring would be written once, and would be correct, code duplication would be kept to a minimum etc.) - I think it's probably easier to merge this and #253 first though and then I'll do the refactor separately.

orix/quaternion/orientation.py
hakonanes: @harripj, good catch. Seems like the Python language with `__doc__` doesn't do any inspection, while IPython/Jupyter does some inspection and finds the docstring both with `?` and with `Shift` + `Tab` with the cursor within function parantheses.

pc494:
The Ipython support is good enough for now, we can deal with the tidy ups elsewhere. I will resolve the conflicts (from merging #253) and merge this.

CHANGELOG.rst
```diff
@@ -27,6 +27,7 @@ Added
   sample direction.
 - Inverse pole figure colouring of orientations given a Laue symmetry and sample
   direction.
+- `Orientation`-based classes now accept a `symmetry` argument for `__init__()`.
```
hakonanes, line 30:
```suggestion
- `Orientation` based classes now accept a `symmetry` argument upon initialisation.
```

CHANGELOG.rst
```diff
@@ -36,6 +37,9 @@ Changed
 - `Orientation` class methods `from_euler`, `from_matrix`, and `from_neo_euler` no longer 
   return the smallest angle orientation when a `symmetry` is given.
 - `CrystalMap.orientations` no longer returns smallest angle orientation.
+- The methods `flatten`, `reshape`, and `squeeze` have been overridden in
+  `Misorientation`-based classes to maintain the initial symmetry of the returned
+  object.
```
hakonanes, line 42:
```suggestion
  `Misorientation` based classes to maintain the initial symmetry of the returned
  instance.
```

orix/tests/quaternion/test_orientation.py
```diff
@@ -115,6 +113,74 @@ def test_getitem(orientation, symmetry):
     assert orientation[0].symmetry._tuples == symmetry._tuples
 
 
+@pytest.mark.parametrize("symmetry", ([C2, C3], [Oh, C2], [O, D3]))
+def test_reshape_maintains_symmetry_misorientation(symmetry):
+    m = Misorientation.random((4, 5))
```
hakonanes, line 118: Why not set the symmetry in this line?

orix/tests/quaternion/test_orientation.py
```diff
@@ -115,6 +113,74 @@ def test_getitem(orientation, symmetry):
...
+    m.symmetry = symmetry
+    m1 = m.reshape(5, 4)
+    for s1, s2 in zip(m1.symmetry, symmetry):
```
hakonanes, line 121: I'd use `np.allclose(m1.symmetry.data, symmetry.data)` here instead.

hakonanes (approved):
Looks great. I have a few small comments which aren't required but would be nice to change.

pc494:
Sorry, I had forgotten that @hakonanes had a review pending. This is good to go as far as I am concerned, so I will leave @hakonanes to merge.

hakonanes:
If it's OK with everyone, I'll apply the changes suggested in the review, fix the conflicts and merge.

orix/tests/quaternion/test_orientation.py
hakonanes: It's not overwritten...

orix/tests/quaternion/test_orientation.py
hakonanes: Nevermind, this makes sense with the symmetry tuple.

hakonanes:
Will merge when checks pass.

orix/quaternion/orientation.py
hakonanes: See https://github.com/pyxem/orix/pull/252#issuecomment-996082199

orix/tests/quaternion/test_orientation.py
harripj: Sorry for the delay, I wasn't very available today. But yes, perhaps we should add a `symmetry` argument to `random` in the future.

Issue #248, opened 2021-12-06
Restructure user guide into topics


hakonanes:
To simplify adding and maintaining notebooks in the future, we should split some of the existing ones and structure them into topics like [MTEX](https://mtex-toolbox.github.io/Documentation.html) does.

A nice setup is with an `reStructuredTex` (rst) file per topic, with each topic file listing a notebook gallery (`nbgallery`) (like is done for the "Examples" part in the [kikuchipy documentation](https://kikuchipy.org/en/stable/examples.html)). We should try to make a notebook gallery pointing to each topic. Not sure how to do that.

The topics and notebooks I can think of from the existing ones are:
- Vectors
  - Definition
  - Operations
  - Sampling
  - Stereographic projection
- Crystal Geometry
  - Crystal symmetry
  - Miller indices
  - Operations
  - Fundamental sectors
- Orientations
  - Definition
  - Clustering
  - Sampling
- Misorientations
  - Definition
  - Clustering
- Crystal Map
  - Create, import and export
  - Phases
  - Selecting data
  - Plotting

hakonanes:
I'd like to get your opinion @pc494 and @harripj on this restructuring?

I often find that I'd like to add an example or something small, but don't want to expand the already long user guides, and so leave it be. The MTEX documentation is great in that they have longer examples, but also shorter more focused ones, so I don't have to scroll long to find what I was after.

I can set up this structure with an aim for a patch release, but releasing this restructure isn't urgent.

hakonanes:
This was done in #322.

PR #253, opened 2021-12-15
Add new classmethods from_axes_angles() to Rotation and Orientation classes

### LINKED ISSUE #250: Support Rotation.from_axes_angles(axes, angles)
We should support `Rotation.from_axes_angles(axes, angles)` as a complement to the current `Rotation.from_neo_euler()`. The reason is that at the moment, initializing a rotation from an axis/angle pair is too much work.

"Minimal" initialization of a rotation from a Euler angle triplet

```python
>>> import numpy as np
>>> rot = Rotation.from_euler(np.radians((45, 0, 0)))
>>> rot
Rotation (1,)
[[ 0.9239 -0.     -0.     -0.3827]]
```

From an axis/angle pair (notice the extra two lines: import and axangle initialization)

```python
>>> import numpy as np
>>> from orix.vector import AxAngle
>>> axangle = AxAngle.from_axes_angles((0, 0, -1), np.radians(45))
>>> rot = Rotation.from_neo_euler(axangle)
>>> rot
Rotation (1,)
[[ 0.9239  0.      0.     -0.3827]]
```

What we should support

```python
>>> import numpy as np
>>> rot = Rotation.from_axes_angles((0, 0, -1), np.radians(45))
>>> rot
Rotation (1,)
[[ 0.9239  0.      0.     -0.3827]]
```

Should we then support `Rotation.from_rodrigues()` and `Rotation.from_homochoric()` as well? I don't see a reason why. We can also still support `Rotation.from_neo_euler()`.

The proposed solution is perhaps less elegant code wise, but *much* simpler to use.

I added this to the v0.8 milestone since it is a very simple addition.

hakonanes:
* Add new classmethods `from_axes_angles()` to `Rotation` and `Orientation`, accepting two arguments `axes` and `angles`. `Orientation` also accepts `symmetry=None`. This method is a shortcut for instatiating an `AxAngle` instance and passing this to `from_neo_euler()`.

Closes #250.

I haven't updated any user guides to use these methods. I anticipate to do that once we have the current PRs with user guide changes in so that we can restructure the guide into topics, as suggested in #248.

#### Progress of the PR

```python
>>> import numpy as np
>>> axis = Vector3d((1, 1, 0))
>>> angles = np.linspace(0, np.radians(54.74), 100)
>>> ori = Orientation.from_axes_angles(-axis, angles, symmetry.Oh)
>>> directions = Vector3d(((1, 0, 0), (0, 1, 0), (0, 0, 1)))
>>> ori.scatter("ipf", c=angles, direction=directions)  # IPF
>>> ori.scatter("rodrigues", c=angles)  # Rodrigues Fundamental Zone
>>> fig = (~ori * directions[2]).scatter(c=angles, grid=True, axes_labels=["x", "y", None], return_figure=True)  # [001] PF
>>> fig.axes[0].draw_circle(axis.perpendicular)
```

ipfs

rfz

pf

#### For reviewers

### DIFF
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -51,6 +51,7 @@ from orix.quaternion.orientation_region import OrientationRegion
 from orix.quaternion.rotation import Rotation
 from orix.quaternion.symmetry import C1, Symmetry
 from orix.scalar import Scalar
+from orix.vector import AxAngle
 from orix._util import deprecated
 
 
@@ -254,11 +255,10 @@ class Misorientation(Rotation):
         Examples
         --------
         >>> import numpy as np
-        >>> from orix.quaternion.symmetry import C4, C2
-        >>> from orix.quaternion.orientation import Misorientation
+        >>> from orix.quaternion import Misorientation, symmetry
         >>> data = np.array([[0.5, 0.5, 0.5, 0.5], [0, 1, 0, 0]])
         >>> m = Misorientation(data)
-        >>> m.symmetry = (C4, C2)
+        >>> m.symmetry = (symmetry.C4, symmetry.C2)
         >>> m = m.map_into_symmetry_reduced_zone()
         >>> m.distance()
         array([[3.14159265, 1.57079633],
@@ -271,12 +271,12 @@ class Misorientation(Rotation):
         """Returns a new Misorientation containing the same data
         transposed.
 
-        If ndim is originally 2, then order may be undefined. In this
+        If `ndim` is originally 2, then order may be undefined. In this
         case the first two dimensions will be transposed.
 
         Parameters
         ----------
-        axes: int, optional
+        axes : int, optional
             The transposed axes order. Only navigation axes need to be
             defined. May be undefined if self only contains two
             navigation dimensions.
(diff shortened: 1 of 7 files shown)


pc494:
As this is most likely to play nicely with #251, #234 and #201 I will review this one first.

orix/quaternion/rotation.py
```diff
@@ -210,8 +210,8 @@ def angle_with(self, other):
         return angles
 
     def outer(self, other, **kwargs):
-        """Compute the outer product of this rotation and the other object."""
-        r = super().outer(other, **kwargs)
+        """Compute the outer product of this rotation and the other."""
+        r = super().outer(other)
```
pc494, line 214: what's going on here?

orix/quaternion/rotation.py
```diff
@@ -261,6 +261,37 @@ def from_neo_euler(cls, neo_euler):
         r = cls(np.stack([a, b, c, d], axis=-1))
         return r
 
+    @classmethod
+    def from_axes_angles(cls, axes, angles):
+        """Creates rotation(s) from axis-angle pair(s).
+
```
pc494, line 267: Might be me being thick, but isn't this just overwriting the method defined in `Quaternion`? Is there a reason for this? Am I missing something?

pc494 (commented):
Two short comments. Core functionality looks good.

orix/quaternion/rotation.py
hakonanes: I started to allow keyword arguments to `outer()` without needing it in #196... My bad. I removed it here, but not entirely. I'll remove it fully. Thanks for asking!

orix/quaternion/rotation.py
hakonanes: Which method in `Quaternion` do you mean? Feel free to suggest an alternative implementation. I selected this approach because it seemed the simplest to write.

pc494: Sorry got this the wrong way round (it's a duplication) - see next comment

orix/quaternion/orientation.py
```diff
@@ -504,8 +504,42 @@ def from_neo_euler(cls, neo_euler, symmetry=None):
             o.symmetry = symmetry
         return o
 
+    @classmethod
```
pc494, line 507: This method isn't needed as `Orientation` inherits it from `Rotation`

pc494 (commented):
Think this just needs a single block delete.

orix/quaternion/orientation.py
hakonanes: It's needed because the symmetry is passed.

orix/quaternion/rotation.py
hakonanes: No problem!

orix/quaternion/orientation.py
pc494: I feel like we should be able to refactor around this, but equally it's the same problem as #252 so maybe a more general solution is needed. I'll have a think.

hakonanes: I agree, I think we could create some decorator that wraps the methods and sets the symmetry afterwards? Since I wrote the `@deprecated` decorator (using what Matplotlib and scikit-image had done, of course), I can look at this, if you think this is a good solution.

hakonanes: But such a decorator would decorate the method here that you don't want... Hm. Not entirely sure how to do it.

pc494:
Okay, I'm going to raise a separate issue to deal with the symmetry problems. I think all this now needs is the removal of the leftover `kwargs`

pc494:
Will treat the problems discussed in #254 somewhere else, merging this.

PR #255, opened 2021-12-17
Ensure Rotation.to_euler() doesn't divide by zero to suppress warning

### LINKED ISSUE #231: Should get rid of RuntimeWarning in Rotation.to_euler()
These warnings are annoying, when either q03 or q12 are zero in [Rotation.to_euler()](https://github.com/pyxem/orix/blob/master/orix/quaternion/rotation.py#L265)

```python
/home/hakon/miniconda3/envs/kp-dev/lib/python3.9/site-packages/orix/quaternion/rotation.py:312: RuntimeWarning: invalid value encountered in true_divide
  np.divide(b * d - a * c, chi), np.divide(-a * b - c * d, chi)
/home/hakon/miniconda3/envs/kp-dev/lib/python3.9/site-packages/orix/quaternion/rotation.py:316: RuntimeWarning: invalid value encountered in true_divide
  np.divide(a * c + b * d, chi), np.divide(c * d - a * b, chi)
```

Let's try to get rid of them. If we release a v0.7.1 patch release, we should add this.

hakonanes:

* Ensure Rotation.to_euler() doesn't divide by zero to suppress warning
* Add test to ensure this

Closes #231.

#### Progress of the PR

#### For reviewers

- [n/a] New features, API changes, and deprecations are mentioned in the

### DIFF
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -338,20 +338,28 @@ class Rotation(Quaternion):
             e[..., 2] = np.where(cond, 0, e[..., 2])
 
         if np.sum(chi != 0) > 0:
-            not_zero = chi != 0
+            not_zero = ~np.isclose(chi, 0)
             alpha = np.arctan2(
-                np.divide(b * d - a * c, chi, where=not_zero, out=np.zeros_like(chi)),
-                np.divide(-a * b - c * d, chi, where=not_zero, out=np.zeros_like(chi)),
+                np.divide(
+                    b * d - a * c, chi, where=not_zero, out=np.full_like(chi, np.inf)
+                ),
+                np.divide(
+                    -a * b - c * d, chi, where=not_zero, out=np.full_like(chi, np.inf)
+                ),
             )
             beta = np.arctan2(2 * chi, q_zero_three - q_one_two)
             gamma = np.arctan2(
-                np.divide(a * c + b * d, chi, where=not_zero, out=np.zeros_like(chi)),
-                np.divide(c * d - a * b, chi, where=not_zero, out=np.zeros_like(chi)),
+                np.divide(
+                    a * c + b * d, chi, where=not_zero, out=np.full_like(chi, np.inf)
+                ),
+                np.divide(
+                    c * d - a * b, chi, where=not_zero, out=np.full_like(chi, np.inf)
+                ),
             )
 
-            e[..., 0] = np.where(chi != 0, alpha, e[..., 0])
-            e[..., 1] = np.where(chi != 0, beta, e[..., 1])
-            e[..., 2] = np.where(chi != 0, gamma, e[..., 2])
+            e[..., 0] = np.where(not_zero, alpha, e[..., 0])
+            e[..., 1] = np.where(not_zero, beta, e[..., 1])
+            e[..., 2] = np.where(not_zero, gamma, e[..., 2])
 
         return e
 
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -338,12 +338,15 @@ class Rotation(Quaternion):
             e[..., 2] = np.where(cond, 0, e[..., 2])
 
         if np.sum(chi != 0) > 0:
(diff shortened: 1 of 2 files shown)


pc494:
is this actually correct?

`np.divide(3,0)` 

return `inf` not zero so I am little lost.

orix/quaternion/rotation.py
```diff
@@ -338,12 +338,15 @@ def to_euler(self, convention="bunge"):
             e[..., 2] = np.where(cond, 0, e[..., 2])
 
         if np.sum(chi != 0) > 0:
+            not_zero = chi != 0
```
harripj, line 341: If  `chi` contains floats (it looks like it is defined by a `sqrt` fn), then perhaps it would be better to use `~np.isclose(chi, 0)` here.

hakonanes:
> is this actually correct?

The warning is raised when the denominator is zero. We don't want inf, do we? Or do we want that? As in an array of inf should be passed to out?

@harripj I think the warning is given when chi is exactly zero, and that's the case I'd like to catch, not close to zero. Or do you think the warning is given when it is close to zero? If so, then your suggestion should be adopted.

harripj:
> @harripj I think the warning is given when chi is exactly zero, and that's the case I'd like to catch, not close to zero. Or do you think the warning is given when it is close to zero? If so, then your suggestion should be adopted.

It was my understanding that `islcose` should always be used with floating points, however at least on my machine `np.all(np.sqrt(np.zeros(50, dtype=float)) == 0)` is `True`, so it looks like the code above is safe.

The following also doesn't throw a warning: `np.divide(1, np.finfo(np.float64).min)`, so it seems to be good to go!

harripj:
I will let @pc494 pass an eye over this, but I am happy if this is merged.

hakonanes:
> It was my understanding that `islcose` should always be used with floating points rather than checking for exact equality, 

I didn't know this, do you know where you have it from? The documentation for [numpy.isclose()](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html) doesn't say anything about this.

Anyway, I changed the equality check to use `isclose()`.

> is this actually correct?
> np.divide(3,0)
> return inf not zero so I am little lost.

You're right, I've changed the output from `divide()` from 0 to `inf` when `chi` is zero.

Thank you both for good feedback!

pc494:
I think we do want to passed array to be `inf`, otherwise we're not just removing a warning but actively changing the behaviour of the code (the `inf` gets turned into a finite number immediately by the `arctan2` so never makes it back to the user)

harripj:
> I didn't know this, do you know where you have it from? The documentation for numpy.isclose() doesn't say anything about this. Anyway, I changed the equality check to use isclose()

Sorry if I was confusing here, I think the previous logic was fine as the warning only arises for exact zeros and testing for exact zeros seems to be okay. 

> The following also doesn't throw a warning: np.divide(1, np.finfo(np.float64).min), so it seems to be good to go!

This should also have been written differently. The only case I can get force an error is division by precisely `0` or `1 / np.spacing(0)`, where `np.spacing(0) = 5e-324` on my machine.

hakonanes:
> Sorry if I was confusing here, I think the previous logic was fine as the warning only arises for exact zeros and testing for exact zeros seems to be okay.

You weren't, I decided to trust your instinct and go with a broader definition of when chi is zero.

pc494:
> > It was my understanding that `islcose` should always be used with floating points rather than checking for exact equality,
> 
> I didn't know this, do you know where you have it from? The documentation for [numpy.isclose()](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html) doesn't say anything about this.
> 
> Anyway, I changed the equality check to use `isclose()`.
> 
> > is this actually correct?
> > np.divide(3,0)
> > return inf not zero so I am little lost.
> 
> You're right, I've changed the output from `divide()` from 0 to `inf` when `chi` is zero.
> 
> Thank you both for good feedback!

I'm merging now, but for what it's worth it's a shame the testing couldn't catch this...

PR #261, opened 2021-12-21
Packaging commits (Changelog and release)

pc494:
We used to do this as manual single commits but this is slightly nicer. Will still self merge though.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,9 @@ this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.htm
 Unreleased
 ==========
 
+2021-12-21 - version 0.8.0
+==========================
+
 Added
 -----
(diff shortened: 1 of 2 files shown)


hakonanes (approved):

hakonanes:
To half the number of PR checks, I recommend making PRs from branches in forks...

pc494:
> To half the number of PR checks, I recommend making PRs from branches in forks...

Honestly, I had forgotten it was two commits and not one.

pc494:
will stop doing it like that then

hakonanes:
We could remove the `push` part, but then checks wouldn't be run on `master` after a PR is merged:

https://github.com/pyxem/orix/blob/9889817132eed74d7229ca7f6c05ab88cdffd16c/.github/workflows/build.yml#L1-L3

> will stop doing it like that then

To save GitHub some energy for computing power, and by extension the Earth!

PR #277, opened 2022-02-09
Fix misorientation calculation in Orientation methods angle_with(), dot() and dot_outer()

### LINKED ISSUE #276: Inconsistency with Orientation.angle_with
I have encountered a potential bug with certain angles between two given orientations.  The orientations start as euler angles and are then converted to orix Orientations.   
 Example:
```python
oritest1 = Orientation.from_euler(np.radians([120.        ,  54.26282923,  45.        ]), symmetry=symmetry.Oh)
oritest2 = Orientation.from_euler(np.radians([ 0.        , 54.73561032, 45.        ]), symmetry=symmetry.Oh)
```
Then if I compare the results of 
```python
np.degrees((oritest1 - oritest2).angle.data)
np.degrees(oritest1.angle_with(oritest2).data)
```
I would expect the angle difference to be very small for both but only the first way gives the expected answer of 0.4727 while angle_with returns 30.0036.  It seems to not take into account the 120degree symmetry in the first euler angle.

Further euler angle examples that give similarly strange results are: 
`
[302.   ,      -52.35178357 , -46.0036662 ] to [ 0.     ,    53.59948955 , 47.5       ]
`
`
[308.5 ,       -51.76150996 , -52.        ]       to [ 0.     ,    51.76150996 , 52.        ]
`
`
[328.5 ,      -47.2657904 , -67.5      ]           to [ 0.    ,    47.2657904 , 67.5      ]
`
These all have large differences when using angle_with and small, more expected, differences when using (Orientation - Orientation).angle.
However (Orientation - Orientation).angle also seems to have some flaws example:
`
[178.5    ,     36.97748349  , 63.61592335] to [ 0.      ,   36.48968436 , 61.39659559]
`
Here Orientation.angle_with gives the expected result of 1.441 while (Orientation - Orientation).angle returns 39.268; seemingly not taking into account the 180degree symmetry in the first angle.

Hopefully this is enough to understand the issue, but if not I can provide more.

hakonanes:
Orientation disorientation angles and dot products returned from `angle_with()` and `dot()` and `dot_outer()`, now calculates the misorientation as other * ~self instead of (~self) * other.

With this change the two routes to obtain a misorientation angle, notably `(o1 - o2).angle` and `o1.angle_with(o2)` (which calls `dot()` and `dot_outer()` internally to get the disorientation angle [symmetry reduced misorientation angle]) return the same results, which was the concern of @soupmongoose in #276. Will close #276.

All tests passed without modifications after making this change. I've added a test to ensure that getting the misorientation from two orientations and and doing the inverse operation results in the identity orientation. Also removed a now unnecessary call to `o1.map_into_symmetry_reduced_zone()`.

Thanks to @harripj for spotting this bug and @soupmongoose for reporting the issue that led to this fix!

#### Progress of the PR

See https://github.com/pyxem/orix/issues/276#issuecomment-1033700249.

#### For reviewers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,7 +16,7 @@ Added
 
 Fixed
 -----
-- Orientation disorientation angles and dot products returned from `angle_with()` and
+- `Orientation` disorientation angles and dot products returned from `angle_with()` and
   `dot()` and `dot_outer()`, which now calculates the misorientation as other * ~self.
 
 2021-12-21 - version 0.8.0
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Added
 -----
 - Python 3.10 support.
 
+Fixed
+-----
+- Orientation disorientation angles and dot products returned from `angle_with()` and
+  `dot()` and `dot_outer()`, which now calculates the misorientation as other * ~self.
+
 2021-12-21 - version 0.8.0
 ==========================
 
@@ -78,7 +83,6 @@ Fixed
 - `CrystalMap.get_map_data()` can return an array of shape (3,) if there are that many
   points in the map.
 - Reading of point groups with "-" sign, like -43m, from EMsoft h5ebsd files.
-
 
 2021-09-07 - version 0.7.0
 ==========================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,7 +17,8 @@ Added
 Fixed
 -----
 - `Orientation` disorientation angles and dot products returned from `angle_with()` and
-  `dot()` and `dot_outer()`, which now calculates the misorientation as other * ~self.
+  `dot()` and `dot_outer()`, which now calculates the misorientation as `other * ~self`.
+  Disorientation angles `(o2 - o1).angle` and `o1.angle_with(o2)` are now the same.
 
 2021-12-21 - version 0.8.0
 ==========================
--- a/orix/quaternion/orientation.py
+++ b/orix/quaternion/orientation.py
@@ -562,7 +562,7 @@ class Orientation(Misorientation):
         dot_outer
         """
         symmetry = self.symmetry.outer(other.symmetry).unique()
-        misorientation = (~self) * other
+        misorientation = other * ~self
         all_dot_products = Rotation(misorientation).dot_outer(symmetry).data
         highest_dot_product = np.max(all_dot_products, axis=-1)
         return Scalar(highest_dot_product)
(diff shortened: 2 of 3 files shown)


harripj:
I haven't used the dask implementation much, but I suppose this would also need changing there?

https://github.com/pyxem/orix/blob/0fa14d11070e1b86eaf905d5d0f538743a91af93/orix/quaternion/orientation.py#L906-L913

hakonanes:
Thanks @harripj, well spotted.

harripj:
Just another comment on the topic, there is a test which confirms that the output from `Quaternion.outer(other)` and `Quaternion._outer_dask(other)` are the same?

hakonanes:
> Just another comment on the topic, there is a test which confirms that the output from `Quaternion.outer(other)` and `Quaternion._outer_dask(other)` are the same?

Again, good that you're thinking of these things. Luckily, I added a test for this when writing the alternative implementation (:

CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Added
 -----
 - Python 3.10 support.
 
+Fixed
+-----
+- Orientation disorientation angles and dot products returned from `angle_with()` and
+  `dot()` and `dot_outer()`, which now calculates the misorientation as other * ~self.
```
harripj, line 20: Could you format `Orientation`, and add a small comment to say the three calculations are now consistent?

harripj (approved):
Looks good to me, I would be happy to merge the fixes if @pc494 can spend a quick minute to look it over.

CHANGELOG.rst
harripj: That they produce the same output, as they didn't before.

hakonanes: By "they" you mean the two routes to obtaining the misorientation angle?

harripj: Sorry, yes, I should be more clear! Yes the two different routes `angle_with` and `dot`, and consistent with each other.

hakonanes: Hehe, no problem! But I still don't understand you... `angle_with()` uses `dot()`, and `dot()` uses `dot_outer()`. These were consistent with each other before, although wrong. What's now consistent which weren't before are `(o2 - o1).angle` and `o1.angle_with(o2)`: do you mean we should state that the latter is consistent in the changelog?

harripj: Yes, you are right and I have been confusing both myself and you the whole time! Yes I think we should state that `(o2 - o1).angle` and `o1.angle_with(o2)` have been made consistent, as it may be useful information to others who come back to look at their code in the future.

hakonanes: Done.

pc494 (approved):
LGTM, merging

PR #278, opened 2022-02-11
Improve handling of property arrays in .ang writer

### LINKED ISSUE #189: .ang writer cannot write property arrays (image quality etc.) over varying shape
Say I have a crystal map with one image quality value per point, but 20 best matching score values per point. Passing `index=0` to the .ang writer is necessary to write the scores to e.g. the pattern fit column in the .ang file, but then the image quality values cannot be written as well. Not passing `index` throws an error when trying to write the best matching score, but *can* write the image quality array. Should not be difficult to fix this.

Also, the column names in the .ang file header doesn't correspond to the actual data column.

I hope to fix this within the week, but might have to be pushed for a v0.6.1 patch.

hakonanes:
Fix #189.

#### Progress of the PR

#### For reviewers

### DIFF
--- a/orix/io/plugins/ang.py
+++ b/orix/io/plugins/ang.py
@@ -401,27 +401,30 @@ def file_writer(
     image_quality_prop : str, optional
         Which map property to use as the image quality. If None
         (default), "iq" or "imagequality", if present, is used, or just
-        zeros. If the property has more than one value per point, only
-        the first value is used.
+        zeros. If the property has more than one value per point and
+        `index` is not given, only the first value is used.
     confidence_index_prop : str, optional
         Which map property to use as the confidence index. If None
-        (default), "ci" or "confidenceindex", if present, is used, or
-        just zeros. If the property has more than one value per point,
-        only the first value is used.
+        (default), "ci", "confidenceindex", "scores", or "correlation",
+        if present, is used, or just zeros. If the property has more
+        than one value per point and `index` is not given, only the
+        first value is used.
     detector_signal_prop : str, optional
         Which map property to use as the detector signal. If None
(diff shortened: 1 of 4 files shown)


pc494 (approved):
LGTM

hakonanes:
Thanks! I'll create release notes tagged off of master for people to review after this is merged.

orix/io/plugins/ang.py
```diff
@@ -636,29 +638,29 @@ def _get_prop_arrays(xmap, prop_names, desired_prop_names, map_size, index, deci
...
     # in the loop
     all_expected_prop_names = [
-        ["iq", "image_quality", "imagequality"],
-        ["ci", "confidence_index", "confidenceindex"],
-        ["ss", "sem_signal", "semsignal"],
-        ["fit", "pattern_fit", "patternfit"],
+        ["iq", "imagequality"],
+        ["ci", "confidenceindex", "scores", "correlation"],
+        ["ss", "semsignal", "detectorsignal"],
+        ["fit", "patternfit"],
```
harripj, line 652: I haven't used this functionality before, so just an off-comment. `detector_signal` and `confidence_index` here have underscores, but in line 508-9 they do not. Is this expected or should they be the same?

hakonanes: The column names in the .ang file header are just for readability, and aren't used in any software as far as I know. In the shown lines we assume underscores and spaces are stripped away in order to find them among `CrystalMap.prop`.

harripj: Thanks for clarifying!

harripj:
Merging.

hakonanes:
Thanks!

Issue #267, opened 2022-01-03
Use of numpy quaternion


din14970:
I recently had a very good experience using the numpy-quaternion package for dealing with rotations/orientations. It may be an idea to use this package as a dependency in orix instead of maintaining a custom quaternion class; perhaps even a thin wrapper could be used.

pc494:
I agree in principle, but given our current shortage of developers this might have to go in the "would be nice to have" column.

hakonanes:
The package looks nice and easy to work with, and you bringing it up here confirms that it was good to list it among among the projects users of orix might find useful [in our documentation](https://orix.readthedocs.io/en/stable/related_projects.html)! Can you give some examples where using numpy-quaternion was better than orix? That might motivate someone to work on this.

Here are my thoughts regarding wrapping numpy-quaternion:

Pros:
* We get access to a lot of numpy.ndarray class methods for free.
* Quaternion multiplication is about 50x faster than in orix!

Cons:
* Not really a "con", but I would say maintaining our Quaternion class is easy, since it hasn't changed much since I started contributing in 2019. It should be noted here that some of the functionality in the numpy-quaternion class we have in our Rotation class.
* Vector multiplication is about 4.3x *slower* than in orix.

___

orix uses slightly more memory in quaternion and vector multiplication. I tested timings with

```python
import numpy as np
import quaternion
from orix.quaternion import Rotation
from orix.vector import Vector3d

v_size = 10
q_size = 1000
v = Vector3d(np.random.random(v_size * 3).reshape((v_size, 3)))
q_orix = Rotation.random((q_size,))  # Not available in Quaternion
q_numpy = quaternion.as_quat_array(q_orix.data)

%timeit q_orix.outer(v)
# 321 µs ± 7.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit quaternion.rotate_vectors(q_numpy, v.data)
# 1.39 ms ± 8.83 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

np.allclose(
    q_orix1.outer(v1).data,
    quaternion.rotate_vectors(q_numpy1, v1.data)
)
# True

%timeit q_orix * q_orix
# 149 µs ± 1.21 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit q_numpy * q_numpy
# 3.01 µs ± 35.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

np.allclose(
[...]

Issue #264, opened 2021-12-22
Scoping Issue for new features (0.9.0)


pc494:
This is mainly an issue to get the lay of the land in terms of new features people either (a) want or (b) are hoping to implement in a short/medium term timeline (6 months give or take). After #201 and #235 I think @hakonanes had almost everything he had hoped for? @harripj are you working on anything with that kind of time frame?

NB: Work that is not ready to be seen in public can be sent to the pyxem email and will be treated as confidential.

hakonanes:
I have a couple of things I'd like to get in at some point.

Entirely new:
- #151: I have a `MapRegions` class locally which is a nice tool to create spatial masks from enclosed regions based on some property, say particles of a certain size. These masks can be used to for example get one averaged diffraction pattern (EBSD in my case) per region. I hope to set up a PR with this class once I have a manuscript ready which uses it within... February? Not because the class is top secret, but because I'll only see which parts of it are useful after the manuscript is ready.

Better use of existing functionality:
- Pole figure plotting should be included. Currently, we have (~orientations * single_crystal_direction).scatter() working, however we need to support (~orientations * symmetrised_crystal_direction).scatter(), which throws an error because of shape mismatch. This should not be much work, and I can do this. I'll raise an issue at a later point.
- #171: This would be very nice when deciding crystal axis alignment in #249.
- #186: This is in kikuchipy, only needs to be moved here.
- #51: This is in kikuchipy, only needs to be moved here.

Apart from this I'm quite happy with the functionality in orix for the moment.

pc494:
Okay, this might need an actual meeting, but I feel that this release (ie. the next major/minor after `0.8.x`) should be the `1.0.0` - I think this an honest reflection of the state of the code, and it communicates how substantial some of the rewrites we're now considering (#267, #270). I've retitled this issue accordingly and am keen for thoughts: @hakonanes, @harripj, @din14970 

NB: apologies for the ping late on a Friday, my orix life just do be like that now.

din14970:
As I have not directly contributed to orix in a while and have not followed all the awesome new developments I think I my opinion in this discussion is not so relevant. As for #267 and #270, they are merely some suggestions that I thought were interesting but I'm not sure if I would get around to properly implementing them. #267 should in theory be easier, #270 may require new features in ASE to be feasible, unless we support both ASE and diffpy as "backends".

What I would still like to contribute to is on improving and adding some features to diffsims. I think some refactoring there is also in order to make use of some of the new orix functionality.

[...]

PR #281, opened 2022-02-14
Implement quaternion arithmetic from numpy-quaternion

harripj:
As discussed in #267 and #264, I thought it best to start to implement quaternion arithmetic from `numpy-quaternion` into `orix.quaternion.Quaternion` given the performance advantage it has shown, amongst other benefits. The purpose of this PR is to lay some groundwork and discuss before making any concrete changes. To this end I have created a `QuaternionNumpy` class which inherits from `Quaternion` to help with testing.

Currently quaternion multiplication (`Quaternion` and `Vector3d`), conjugation, and outer products have been overloaded to use `numpy-quaternion` and give the same results. A possible extension would be to subclass `orix.quaternion.Quaternion` from `numpy-quaternion` but this would complicate things further and I don't think this is the right direction to go in. Just by using the `numpy-quaternion` arithmetic we get a computational speed advantage rather easily. Memory profiling to be done later.

One (potential) downside is that `numpy-quaternion` will be a further dependency.

@din14970 I achieved the current `Quaternion * Vector3d` functionality using your logic as written in https://github.com/moble/quaternion/pull/191. Orix currently doesn't perform the outer product when calculating `Quaternion * Vector3d` so this works well, and `quaternion.rotate_vectors` is implemented only in `Quaternion.outer`.

#### Progress of the PR

```python
from orix.quaternion import Quaternion, Rotation, QuaternionNumpy
import numpy as np
import time

def test_orix_numpy_quaternion_rotate_vector_shapes():
    q_shape = [(1,), (1,), (4,), (4, 5), (4, 5), (1,), (6,), (4, 5), (4, 2, 5), (4, 2, 5)]
    v_shape = [(1,), (4,), (1,), (1,), (6,), (4, 5), (4, 5), (5,), (2, 5), (3, 6)]

    print('Q shape\t\t V shape\tOrix\t\t Numpy\t\tEqual\t\tOuter V eq\t\tOuter Q eq')

    for qs, vs in zip(q_shape, v_shape):
        q = Rotation.random(qs)
        qn = QuaternionNumpy(q.data)
        assert q.shape == qn.shape

        v = np.random.randn(*vs, 3)
        v3d = Vector3d(v)

        try:
            vo = q * v3d
            passes_orix = True
        except ValueError:
            passes_orix = False

        try:
            vn = qn * v3d
            passes_numpy = True
        except ValueError:
            passes_numpy = False

        assert passes_numpy == passes_orix
        if all([passes_numpy, passes_orix]):
            ac = np.allclose(vo.data, vn.data)
        else:
            ac = 'n/a'
        
        oov = q.outer(v3d)
        onv = qn.outer(v3d)

        q2 = Rotation.random(vs)
        ooq = q.outer(q2)
        onq = qn.outer(q2)
        print(f'{qs}\t\t{vs}\t\t{passes_orix}\t\t{passes_numpy}\t\t{ac}\t\t{np.allclose(oov.data, onv.data)}\t\t\t{np.allclose(ooq.data, onq.data)}')

test_orix_numpy_quaternion_rotate_vector_shapes()
```
```
Q shape		 V shape	Orix		 Numpy		Equal		Outer V eq		Outer Q eq
(1,)		(1,)		True		True		True		True			True
(1,)		(4,)		True		True		True		True			True
(4,)		(1,)		True		True		True		True			True
(4, 5)		(1,)		True		True		True		True			True
(4, 5)		(6,)		False		False		n/a		True			True
(1,)		(4, 5)		True		True		True		True			True
(6,)		(4, 5)		False		False		n/a		True			True
(4, 5)		(5,)		True		True		True		True			True
(4, 2, 5)	(2, 5)		True		True		True		True			True
(4, 2, 5)	(3, 6)		False		False		n/a		True			True
```

Some timings...

Vector operations
------------------
```python
qs, vs = (3,), (1,)
q = Quaternion(Rotation.random(qs))
qn = QuaternionNumpy(q)

v = Vector3d(np.random.randn(*vs, 3))
%timeit q * v
%timeit qn * v

71.8 µs ± 5.31 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
37.3 µs ± 1.49 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

```python
qs, vs = (120, 12), (1,)
q = Quaternion(Rotation.random(qs))
qn = QuaternionNumpy(q)

v = Vector3d(np.random.randn(*vs, 3))
%timeit q * v
%timeit qn * v

136 µs ± 8.5 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
66.9 µs ± 3.46 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

```python
qs, vs = (4, 2, 5), (2, 5,)
q = Quaternion(Rotation.random(qs))
qn = QuaternionNumpy(q)

v = Vector3d(np.random.randn(*vs, 3))
%timeit q * v
%timeit qn * v

82.5 µs ± 3.68 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
41.3 µs ± 863 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

`Quaternion.outer(Vector3d)` seems to be on par for small calculations or quicker for larger ones (using `quaternion.rotate_vectors`):
```python
qs, vs = (4, 2, 5), (2, 5,)
q = Quaternion(Rotation.random(qs))
qn = QuaternionNumpy(q)

v = Vector3d(np.random.randn(*vs, 3))
%timeit q.outer(v)
%timeit qn.outer(v)

112 µs ± 14.3 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
117 µs ± 4.97 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

```python
qs, vs = (140, 21, 5), (21, 5,)
q = Quaternion(Rotation.random(qs))
qn = QuaternionNumpy(q)

v = Vector3d(np.random.randn(*vs, 3))
%timeit q.outer(v)
%timeit qn.outer(v)

66.1 ms ± 6.97 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
17.9 ms ± 1.78 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Quaternion operations
----------------------
```python
qs, vs = (140, 21, 5), (21, 5,)
q1 = Quaternion(Rotation.random(qs))
q1n = QuaternionNumpy(q1)

q2 = Quaternion(Rotation.random(vs))
q2n = QuaternionNumpy(q2)
%timeit q * q2
%timeit q1n * q2n

574 µs ± 34 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
201 µs ± 11.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

```python
qs, vs = (3, 2, 5), (1, 5,)
q1 = Quaternion(Rotation.random(qs))
q1n = QuaternionNumpy(q1)

q2 = Quaternion(Rotation.random(vs))
q2n = QuaternionNumpy(q2)
%timeit q * q2
%timeit q1n * q2n

725 µs ± 31 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
17.5 µs ± 1.15 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

```python
qs, vs = (3, 2, 5), (1, 5,)
q1 = Quaternion(Rotation.random(qs))
q1n = QuaternionNumpy(q1)

q2 = Quaternion(Rotation.random(vs))
q2n = QuaternionNumpy(q2)
%timeit q.outer(q2)
%timeit q1n.outer(q2n)

3.19 ms ± 78.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
23.1 µs ± 1.7 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

```python
qs, vs = (3, 2, 5), (100, 5,)
q1 = Quaternion(Rotation.random(qs))
q1n = QuaternionNumpy(q1)

q2 = Quaternion(Rotation.random(vs))
q2n = QuaternionNumpy(q2)
%timeit q.outer(q2)
%timeit q1n.outer(q2n)

558 ms ± 18.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
169 µs ± 22.6 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

#### For reviewers

### DIFF
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -90,39 +90,25 @@ class Quaternion(Object3d):
 
     @property
     def conj(self):
-        a = self.a.data
-        b, c, d = -self.b.data, -self.c.data, -self.d.data
-        q = np.stack((a, b, c, d), axis=-1)
-        return Quaternion(q)
+        q = quaternion.from_float_array(self.data).conj()
+        return Quaternion(quaternion.as_float_array(q))
 
     def __invert__(self):
         return self.__class__(self.conj.data / (self.norm.data**2)[..., np.newaxis])
 
     def __mul__(self, other):
         if isinstance(other, Quaternion):
-            sa, oa = self.a.data, other.a.data
-            sb, ob = self.b.data, other.b.data
-            sc, oc = self.c.data, other.c.data
-            sd, od = self.d.data, other.d.data
-            a = sa * oa - sb * ob - sc * oc - sd * od
-            b = sb * oa + sa * ob - sd * oc + sc * od
-            c = sc * oa + sd * ob + sa * oc - sb * od
-            d = sd * oa - sc * ob + sb * oc + sa * od
-            q = np.stack((a, b, c, d), axis=-1)
-            return other.__class__(q)
+            q1 = quaternion.from_float_array(self.data)
+            q2 = quaternion.from_float_array(other.data)
+            return other.__class__(quaternion.as_float_array(q1 * q2))
         elif isinstance(other, Vector3d):
-            a, b, c, d = self.a.data, self.b.data, self.c.data, self.d.data
-            x, y, z = other.x.data, other.y.data, other.z.data
-            x_new = (a**2 + b**2 - c**2 - d**2) * x + 2 * (
-                (a * c + b * d) * z + (b * c - a * d) * y
+            # check broadcast shape is correct before calculation, as
+            # quaternion.rotat_vectors will perform outer product
+            # this keeps current __mul__ broadcast behaviour
+            q1 = quaternion.from_float_array(self.data)
+            v = quaternion.as_vector_part(
+                (q1 * quaternion.from_vector_part(other.data)) * ~q1
             )
-            y_new = (a**2 - b**2 + c**2 - d**2) * y + 2 * (
-                (a * d + b * c) * x + (c * d - a * b) * z
-            )
-            z_new = (a**2 - b**2 - c**2 + d**2) * z + 2 * (
-                (a * b + c * d) * y + (b * d - a * c) * x
-            )
-            v = np.stack((x_new, y_new, z_new), axis=-1)
             if isinstance(other, Miller):
                 m = other.__class__(xyz=v, phase=other.phase)
                 m.coordinate_format = other.coordinate_format
@@ -225,33 +211,15 @@ class Quaternion(Object3d):
         orix.quaternion.Quaternion or orix.vector.Vector3d
         """
 
-        def e(x, y):
-            return np.multiply.outer(x, y)
-
         if isinstance(other, Quaternion):
-            q = np.zeros(self.shape + other.shape + (4,), dtype=float)
-            sa, oa = self.data[..., 0], other.data[..., 0]
-            sb, ob = self.data[..., 1], other.data[..., 1]
-            sc, oc = self.data[..., 2], other.data[..., 2]
-            sd, od = self.data[..., 3], other.data[..., 3]
-            q[..., 0] = e(sa, oa) - e(sb, ob) - e(sc, oc) - e(sd, od)
-            q[..., 1] = e(sb, oa) + e(sa, ob) - e(sd, oc) + e(sc, od)
-            q[..., 2] = e(sc, oa) + e(sd, ob) + e(sa, oc) - e(sb, od)
-            q[..., 3] = e(sd, oa) - e(sc, ob) + e(sb, oc) + e(sa, od)
-            return other.__class__(q)
+            q1 = quaternion.from_float_array(self.data)
+            q2 = quaternion.from_float_array(other.data)
+            # np.outer works with flattened array
+            q = np.outer(q1, q2).reshape(q1.shape + q2.shape)
+            return other.__class__(quaternion.as_float_array(q))
         elif isinstance(other, Vector3d):
-            a, b, c, d = self.a.data, self.b.data, self.c.data, self.d.data
-            x, y, z = other.x.data, other.y.data, other.z.data
-            x_new = e(a**2 + b**2 - c**2 - d**2, x) + 2 * (
-                e(a * c + b * d, z) + e(b * c - a * d, y)
-            )
-            y_new = e(a**2 - b**2 + c**2 - d**2, y) + 2 * (
-                e(a * d + b * c, x) + e(c * d - a * b, z)
-            )
-            z_new = e(a**2 - b**2 - c**2 + d**2, z) + 2 * (
-                e(a * b + c * d, y) + e(b * d - a * c, x)
-            )
-            v = np.stack((x_new, y_new, z_new), axis=-1)
+            q = quaternion.from_float_array(self.data)
+            v = quaternion.rotate_vectors(q, other.data)
             if isinstance(other, Miller):
                 m = other.__class__(xyz=v, phase=other.phase)
                 m.coordinate_format = other.coordinate_format
@@ -340,83 +308,3 @@ class Quaternion(Object3d):
 
         new_chunks = tuple(chunks1[:-1]) + tuple(chunks2[:-1]) + (-1,)
         return da.stack((a, b, c, d), axis=-1).rechunk(new_chunks)
-
-
-class QuaternionNumpy(Quaternion):
-    r"""Basic quaternion object.
-
-    Quaternions support the following mathematical operations:
-
-    - Unary negation.
-    - Inversion.
-    - Multiplication with other quaternions and vectors.
-
-    Attributes
-    ----------
-    data : numpy.ndarray
-        The numpy array containing the quaternion data.
-    a, b, c, d : Scalar
-        The individual elements of each vector.
-    conj : Quaternion
-        The conjugate of this quaternion :math:`q^* = a - bi - cj - dk`.
-    """
-
-    @property
-    def conj(self):
-        q = quaternion.from_float_array(self.data).conj()
-        return Quaternion(quaternion.as_float_array(q))
-
-    def __mul__(self, other):
-        if isinstance(other, Quaternion):
-            q1 = quaternion.from_float_array(self.data)
-            q2 = quaternion.from_float_array(other.data)
-            return other.__class__(quaternion.as_float_array(q1 * q2))
-        elif isinstance(other, Vector3d):
-            # check broadcast shape is correct before calculation, as
-            # quaternion.rotat_vectors will perform outer product
-            # this keeps current __mul__ broadcast behaviour
-            q1 = quaternion.from_float_array(self.data)
-            v = quaternion.as_vector_part(
-                (q1 * quaternion.from_vector_part(other.data)) * ~q1
-            )
-            if isinstance(other, Miller):
-                m = other.__class__(xyz=v, phase=other.phase)
-                m.coordinate_format = other.coordinate_format
-                return m
-            else:
-                return other.__class__(v)
-        return NotImplemented
-
-    def outer(self, other):
-        """Compute the outer product of this quaternion and the other
-        quaternion or vector.
-
-        Parameters
-        ----------
-        other : orix.quaternion.Quaternion or orix.vector.Vector3d
-
-        Returns
-        -------
-        orix.quaternion.Quaternion or orix.vector.Vector3d
-        """
-
-        if isinstance(other, Quaternion):
-            q1 = quaternion.from_float_array(self.data)
-            q2 = quaternion.from_float_array(other.data)
-            # np.outer works with flattened array
-            q = np.outer(q1, q2).reshape(q1.shape + q2.shape)
-            return other.__class__(quaternion.as_float_array(q))
-        elif isinstance(other, Vector3d):
-            q = quaternion.from_float_array(self.data)
-            v = quaternion.rotate_vectors(q, other.data)
-            if isinstance(other, Miller):
-                m = other.__class__(xyz=v, phase=other.phase)
-                m.coordinate_format = other.coordinate_format
-                return m
-            else:
-                return other.__class__(v)
-        else:
-            raise NotImplementedError(
-                "This operation is currently not avaliable in orix, please use outer "
-                "with `other` of type `Quaternion` or `Vector3d`"
-            )
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -20,6 +20,7 @@ import warnings
 
 import dask.array as da
 import numpy as np
+import quaternion
 
 from orix.base import check, Object3d
 from orix.scalar import Scalar
@@ -339,3 +340,83 @@ class Quaternion(Object3d):
 
         new_chunks = tuple(chunks1[:-1]) + tuple(chunks2[:-1]) + (-1,)
         return da.stack((a, b, c, d), axis=-1).rechunk(new_chunks)
+
+
+class QuaternionNumpy(Quaternion):
+    r"""Basic quaternion object.
+
+    Quaternions support the following mathematical operations:
+
+    - Unary negation.
+    - Inversion.
+    - Multiplication with other quaternions and vectors.
+
+    Attributes
+    ----------
+    data : numpy.ndarray
+        The numpy array containing the quaternion data.
+    a, b, c, d : Scalar
+        The individual elements of each vector.
+    conj : Quaternion
+        The conjugate of this quaternion :math:`q^* = a - bi - cj - dk`.
+    """
+
+    @property
+    def conj(self):
+        q = quaternion.from_float_array(self.data).conj()
+        return Quaternion(quaternion.as_float_array(q))
+
+    def __mul__(self, other):
+        if isinstance(other, Quaternion):
+            q1 = quaternion.from_float_array(self.data)
+            q2 = quaternion.from_float_array(other.data)
+            return other.__class__(quaternion.as_float_array(q1 * q2))
+        elif isinstance(other, Vector3d):
+            # check broadcast shape is correct before calculation, as
+            # quaternion.rotat_vectors will perform outer product
+            # this keeps current __mul__ broadcast behaviour
+            q1 = quaternion.from_float_array(self.data)
+            v = quaternion.as_vector_part(
+                (q1 * quaternion.from_vector_part(other.data)) * ~q1
+            )
+            if isinstance(other, Miller):
+                m = other.__class__(xyz=v, phase=other.phase)
+                m.coordinate_format = other.coordinate_format
+                return m
+            else:
+                return other.__class__(v)
+        return NotImplemented
+
+    def outer(self, other):
+        """Compute the outer product of this quaternion and the other
+        quaternion or vector.
+
+        Parameters
+        ----------
+        other : orix.quaternion.Quaternion or orix.vector.Vector3d
+
+        Returns
+        -------
+        orix.quaternion.Quaternion or orix.vector.Vector3d
+        """
+
+        if isinstance(other, Quaternion):
+            q1 = quaternion.from_float_array(self.data)
+            q2 = quaternion.from_float_array(other.data)
+            # np.outer works with flattened array
+            q = np.outer(q1, q2).reshape(q1.shape + q2.shape)
+            return other.__class__(quaternion.as_float_array(q))
+        elif isinstance(other, Vector3d):
+            q = quaternion.from_float_array(self.data)
+            v = quaternion.rotate_vectors(q, other.data)
+            if isinstance(other, Miller):
+                m = other.__class__(xyz=v, phase=other.phase)
+                m.coordinate_format = other.coordinate_format
+                return m
+            else:
+                return other.__class__(v)
+        else:
+            raise NotImplementedError(
+                "This operation is currently not avaliable in orix, please use outer "
+                "with `other` of type `Quaternion` or `Vector3d`"
+            )
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -40,6 +40,31 @@ class Quaternion(Object3d):
     - Inversion.
     - Multiplication with other quaternions and vectors.
 
+    Quaternion-quaternion multiplication for two quaternions
+    :math:`q_1 = (a_1, b_1, c_1, d_1)`
+    and :math:`q_2 = (a_2, b_2, c_2, d_2)`
+    with :math:`q_3 = (a_3, b_3, c_3, d_3) = q_1 * q_2` follows as:
+
+    .. math::
+       a_3 = (a_1 * a_2 - b_1 * b_2 - c_1 * c_2 - d_1 * d_2)
+
+       b_3 = (a_1 * b_2 + b_1 * a_2 + c_1 * d_2 - d_1 * c_2)
+
+       c_3 = (a_1 * c_2 - b_1 * d_2 + c_1 * a_2 + d_1 * b_2)
+
+       d_3 = (a_1 * d_2 + b_1 * c_2 - c_1 * b_2 + d_1 * a_2)
+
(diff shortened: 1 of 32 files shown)


hakonanes:
This PR is great! Smart to write it alongside the current implementation for benchmarking.

I'm already convinced we should replace our own quaternion multiplication and vector rotation with numpy-quaternion. What I'd like to see in this PR is that change, *and* that we support the missing shapes in the table. This is necessary for pole figure plotting.

> One (potential) downside is that numpy-quaternion will be a further dependency.

[Their dependencies](https://github.com/moble/quaternion/blob/main/requirements.txt) are SciPy, NumPy and Numba, which we already depend on, so this is isn't a downside in my opinion.

harripj:
> and that we support the missing shapes in the table. This is necessary for pole figure plotting.

Do you mean in the testing shape results shown above or something else?

hakonanes:
> I would favour keeping the behaviour in orix, as it means not changing the current behaviour (and so potentially breaking others' code).

I see you've given this more thought than I have, as I've only encountered the undesirable `ValueError` when multiplying 2D quaternion and vector instances with the thought that "this should be allowed". I cannot actually think back to the use case where I encountered this. I guess I could have used `outer()` in that case... I think you're right in that we should keep current behaviour.

harripj:
I will leave this PR as is for a couple days for discussion if anyone would be interested to pull this PR and check the functionality (`Quaternion` vs `QuaternionNumpy`). After that I will tidy it up and finish it off.

harripj:
A note on memory usage between the two implementations, here is the testing script I wrote, run it with `python -m memory_profiler script.py`:

Code
-----
```python
from memory_profiler import profile
import numpy as np

from orix.quaternion import Quaternion, QuaternionNumpy, Rotation

S1 = (2200, 1001)
S2 = S1[-1]
S3 = (27,)

# conjugation
@profile
def test_q_conj_orix():
    q = Quaternion(Rotation.random(S1))
    qc = q.conj
    qinv = ~q


@profile
def test_q_conj_numpy():
    q = QuaternionNumpy(Rotation.random(S1))
    qc = q.conj
    qinv = ~q


# straight multiplication
@profile
def test_q_multiplication_orix():
    q1 = Quaternion(Rotation.random(S1))
    q2 = Quaternion(Rotation.random(S2))
    q = q1 * q2
    assert q.shape == q1.shape


@profile
def test_q_multiplication_numpy():
    q1 = QuaternionNumpy(Rotation.random(S1))
    q2 = QuaternionNumpy(Rotation.random(S2))
    q = q1 * q2
    assert q.shape == q1.shape


@profile
def test_v_mutliplication_orix():
    q1 = Quaternion(Rotation.random(S1))
    vdata = np.random.randn(q1.shape[-1], 3)
    v1 = Vector3d(vdata)
    v = q1 * v1
    assert v.shape == q1.shape


@profile
def test_v_multiplication_numpy():
    q1 = QuaternionNumpy(Rotation.random(S1))
    vdata = np.random.randn(q1.shape[-1], 3)
    v1 = Vector3d(vdata)
    v = q1 * v1
    assert v.shape == q1.shape


# outer product
@profile
def test_q_outer_orix():
    q1 = Quaternion(Rotation.random(S1))
    q2 = Quaternion(Rotation.random(S3))
    q = q1.outer(q2)
    assert q.shape == q1.shape + q2.shape


@profile
def test_q_outer_numpy():
    q1 = QuaternionNumpy(Rotation.random(S1))
    q2 = QuaternionNumpy(Rotation.random(S3))
    q = q1.outer(q2)
    assert q.shape == q1.shape + q2.shape


@profile
def test_v_outer_orix():
    q1 = Quaternion(Rotation.random(S1))
    vdata = np.random.randn(*S3, 3)
    v1 = Vector3d(vdata)
    v = q1.outer(v1)
    assert v.shape == q1.shape + v1.shape


@profile
def test_v_outer_numpy():
    q1 = QuaternionNumpy(Rotation.random(S1))
    vdata = np.random.randn(*S3, 3)
    v1 = Vector3d(vdata)
    v = q1.outer(v1)
    assert v.shape == q1.shape + v1.shape


if __name__ == "__main__":
    # test_q_conj_orix()
    # test_q_conj_numpy()
    test_q_multiplication_orix()
    # test_q_multiplication_numpy()
    # test_v_mutliplication_orix()
    # test_v_multiplication_numpy()
    # test_q_outer_orix()
    # test_q_outer_numpy()
    # test_v_outer_orix()
    # test_v_outer_numpy()
```
NB. run the tests above one by one to avoid intermediate garbage collection noise.

Results
-------
NB. results are in the order above as shown in `__main__`, ie. same code with `orix` then `numpy-quaternion`, `orix` then `numpy-quaternion` etc. for direct comparison. 
```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    12    124.8 MiB    124.8 MiB           1   @profile
    13                                         def test_q_conj_orix():
    14    479.2 MiB    354.4 MiB           1       q = Quaternion(Rotation.random(S1))
    15    579.5 MiB    100.3 MiB           1       qc = q.conj
    16    714.1 MiB    134.7 MiB           1       qinv = ~q

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    19    124.6 MiB    124.6 MiB           1   @profile
    20                                         def test_q_conj_numpy():
    21    479.0 MiB    354.4 MiB           1       q = QuaternionNumpy(Rotation.random(S1))
    22    613.2 MiB    134.2 MiB           1       qc = q.conj
    23    697.5 MiB     84.3 MiB           1       qinv = ~q

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    27    125.3 MiB    125.3 MiB           1   @profile
    28                                         def test_q_multiplication_orix():
    29    479.7 MiB    354.4 MiB           1       q1 = Quaternion(Rotation.random(S1))
    30    479.4 MiB     -0.3 MiB           1       q2 = Quaternion(Rotation.random(S2))
    31    613.6 MiB    134.1 MiB           1       q = q1 * q2
    32    613.6 MiB      0.0 MiB           1       assert q.shape == q1.shape

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    35    124.7 MiB    124.7 MiB           1   @profile
    36                                         def test_q_multiplication_numpy():
    37    479.1 MiB    354.4 MiB           1       q1 = QuaternionNumpy(Rotation.random(S1))
    38    478.9 MiB     -0.3 MiB           1       q2 = QuaternionNumpy(Rotation.random(S2))
    39    613.1 MiB    134.2 MiB           1       q = q1 * q2
    40    613.1 MiB      0.0 MiB           1       assert q.shape == q1.shape

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    43    124.7 MiB    124.7 MiB           1   @profile
    44                                         def test_v_mutliplication_orix():
    45    479.1 MiB    354.4 MiB           1       q1 = Quaternion(Rotation.random(S1))
    46    479.1 MiB      0.0 MiB           1       vdata = np.random.randn(q1.shape[-1], 3)
    47    479.1 MiB      0.0 MiB           1       v1 = Vector3d(vdata)
    48    562.7 MiB     83.5 MiB           1       v = q1 * v1
    49    562.7 MiB      0.0 MiB           1       assert v.shape == q1.shape

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    52    124.7 MiB    124.7 MiB           1   @profile
    53                                         def test_v_multiplication_numpy():
    54    479.2 MiB    354.5 MiB           1       q1 = QuaternionNumpy(Rotation.random(S1))
    55    479.2 MiB      0.0 MiB           1       vdata = np.random.randn(q1.shape[-1], 3)
    56    479.2 MiB      0.0 MiB           1       v1 = Vector3d(vdata)
    57    747.3 MiB    268.1 MiB           1       v = q1 * v1
    58    747.3 MiB      0.0 MiB           1       assert v.shape == q1.shape

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    62    125.0 MiB    125.0 MiB           1   @profile
    63                                         def test_q_outer_orix():
    64    479.3 MiB    354.3 MiB           1       q1 = Quaternion(Rotation.random(S1))
    65    479.3 MiB      0.0 MiB           1       q2 = Quaternion(Rotation.random(S3))
    66   2294.2 MiB   1814.9 MiB           1       q = q1.outer(q2)
    67   2294.2 MiB      0.0 MiB           1       assert q.shape == q1.shape + q2.shape

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    70    124.6 MiB    124.6 MiB           1   @profile
    71                                         def test_q_outer_numpy():
    72    478.9 MiB    354.3 MiB           1       q1 = QuaternionNumpy(Rotation.random(S1))
    73    478.9 MiB      0.0 MiB           1       q2 = QuaternionNumpy(Rotation.random(S3))
    74   2360.4 MiB   1881.5 MiB           1       q = q1.outer(q2)
    75   2360.4 MiB      0.0 MiB           1       assert q.shape == q1.shape + q2.shape

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    78    125.0 MiB    125.0 MiB           1   @profile
    79                                         def test_v_outer_orix():
    80    479.1 MiB    354.1 MiB           1       q1 = Quaternion(Rotation.random(S1))
    81    479.1 MiB      0.0 MiB           1       vdata = np.random.randn(*S3, 3)
    82    479.1 MiB      0.0 MiB           1       v1 = Vector3d(vdata)
    83   1856.9 MiB   1377.8 MiB           1       v = q1.outer(v1)
    84   1856.9 MiB      0.0 MiB           1       assert v.shape == q1.shape + v1.shape

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    87    124.7 MiB    124.7 MiB           1   @profile
    88                                         def test_v_outer_numpy():
    89    479.2 MiB    354.5 MiB           1       q1 = QuaternionNumpy(Rotation.random(S1))
    90    479.2 MiB      0.0 MiB           1       vdata = np.random.randn(*S3, 3)
    91    479.2 MiB      0.0 MiB           1       v1 = Vector3d(vdata)
    92   1945.6 MiB   1466.4 MiB           1       v = q1.outer(v1)
    93   1945.6 MiB      0.0 MiB           1       assert v.shape == q1.shape + v1.shape
```

It appears that there is a small memory penalty for using `numpy-quaternion` which is most apparent when rotating vectors.

Further testing...
```python
@profile
def test_v_multiplication_further():
    q1 = QuaternionNumpy(Rotation.random(S1))
    vdata = np.random.randn(q1.shape[-1], 3)
    v1 = Vector3d(vdata)
    # from source code
    q1 = quaternion.from_float_array(q1.data)
    v_intermediate = quaternion.from_vector_part(v1.data)
    mult1 = q1 * v_intermediate
    q1_inv = ~q1
    mult2 = mult1 * q1_inv
    v = quaternion.as_vector_part(mult2)
```
```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    97    124.2 MiB    124.2 MiB           1   @profile
    98                                         def test_v_multiplication_further():
    99    479.3 MiB    355.1 MiB           1       q1 = QuaternionNumpy(Rotation.random(S1))
   100    479.4 MiB      0.0 MiB           1       vdata = np.random.randn(q1.shape[-1], 3)
   101    479.4 MiB      0.0 MiB           1       v1 = Vector3d(vdata)
   102                                             # from source code
   103    546.1 MiB     66.7 MiB           1       q1 = quaternion.from_float_array(q1.data)
   104    546.1 MiB      0.0 MiB           1       v_intermediate = quaternion.from_vector_part(v1.data)
   105    546.1 MiB      0.0 MiB           1       mult1 = q1 * v_intermediate
   106    613.3 MiB     67.2 MiB           1       q1_inv = ~q1
   107    680.5 MiB     67.2 MiB           1       mult2 = mult1 * q1_inv
   108    680.5 MiB      0.0 MiB           1       v = quaternion.as_vector_part(mult2)
```

The `q1_inv` array is intermediate so I suppose would soon be garbage collected, bringing the overall memory footprint down, but peak memory usage is still increased.

Overall I think memory footprint is outweighed by the performance (CPU) gains and it would be worth including `numpy-quaternion` in orix.

harripj:
So a note on the failing test:

https://github.com/pyxem/orix/blob/ff6e541cab81cdb99d742ec907d257ef96dcdb75/orix/tests/test_miller.py#L210-L220

https://github.com/pyxem/orix/blob/ff6e541cab81cdb99d742ec907d257ef96dcdb75/orix/tests/test_miller.py#L216
This line says the test should have 450 unique vectors, however after using the `numpy-quaternion` implementation, this is not the case and we get 285 vectors.

On further testing this appears to be possibly related to rounding errors (NB I think line 673 should be removed from this function):
https://github.com/pyxem/orix/blob/ff6e541cab81cdb99d742ec907d257ef96dcdb75/orix/vector/miller.py#L645-L689

Using the original implementation without `numpy-quaternion` if the data is rounded to 14 decimal places in line 675 then we also get only 285 unique vectors. So the question is do we have a ground truth here? @hakonanes git blame informs me you wrote the test, what do you think?

hakonanes:
14 decimals sounds more than enough. I'll have a look the results with rounding if they are as expected.

harripj:
> 14 decimals sounds more than enough.

Yes, that's what I thought too. So then it follows as to whether we actually expect all 450 unique vectors, or whether some are the same within this margin.

hakonanes:
Tested the return from `Miller.unique(use_symmetry=True)` with the following example from the [user guide](https://orix.readthedocs.io/en/stable/crystal_geometry.html#A-diamond-[111]-pole-figure):

```python


ver = "stable"  # 0.8.1
#ver = "paddy"  # harripj:numpy-quaternion

diamond = Phase(space_group=227)

# Figure 1 (both hemispheres in all cases)
h = Miller.from_highest_indices(phase=diamond, uvw=[10, 10, 10])  # stable: 9260; paddy: 9260
fig = h.scatter(hemisphere="both", return_figure=True, figure_kwargs=dict(figsize=(20, 10)))
fig.tight_layout()
fig.savefig(os.path.join(dir_test, f"diamond_pf_{ver}.png"), **savefig_kwargs)

# Figure 2
h2 = h.unique(use_symmetry=True)  # stable: 450; paddy: 285
fig2 = h2.scatter(hemisphere="both", return_figure=True, figure_kwargs=dict(figsize=(20, 10)))
fig2.tight_layout()
fig2.savefig(os.path.join(dir_test, f"diamond_unique_pf_{ver}.png"), **savefig_kwargs)

# Figure 3
h3 = h[h <= diamond.point_group.fundamental_sector]  # stable: 285; paddy: 285
fig3 = h3.scatter(hemisphere="both", return_figure=True, figure_kwargs=dict(figsize=(20, 10)))
fig3.tight_layout()
fig3.savefig(os.path.join(dir_test, f"diamond_sector_pf_{ver}.png"), **savefig_kwargs)
```

stable | paddy
:--:|:---:
diamond_pf_stable|diamond_pf_paddy
diamond_unique_pf_stable|diamond_unique_pf_paddy
diamond_sector_pf_stable|diamond_sector_pf_paddy

With this PR, the number of Miller indices obtained by considering only unique indices under symmetry, 285, is equal to the number of indices within the fundamental sector, 285 (comparing row 2 and 3). This is not the case for the current release, which returns 450 indices in the first instance, and 285 in the second. If the fundamental sector of m-3m is defined correctly (which I believe it is), the numbers should be the same. Thus, I conclude that there is an error in the current release which this PR solves.

What do you think, @harripj?

hakonanes:
By changing this line `Miller.unique()`

https://github.com/pyxem/orix/blob/ff6e541cab81cdb99d742ec907d257ef96dcdb75/orix/vector/miller.py#L675

to `data = v2.data.round(16)`, we get the same 450 indices unique under symmetry from our instance of 9260 indices; using `15` we get 449, and using `14` we get 285. Changing the line to `data = v2.data.astype(int)` (Miller indices aren't restricted to integers, so don't do this!) also returns 285 in this case.

You've fixed a bug here, and I'm very grateful :)

hakonanes:
> (NB I think line 673 should be removed from this function):

Yes please!

harripj:
This makes sense to me, in fact it would be great if `unique` chose the same fundamental sector on the unit sphere as `point_group.fundamental_sector`.

> to data = v2.data.round(16), we get the same 450 indices unique under symmetry from our instance of 9260 indices; using 15 we get 449, and using 14 we get 285. Changing the line to data = v2.data.astype(int) (Miller indices aren't restricted to integers, so don't do this!) also returns 285 in this case.

This test is very informative! So it looks like the correct test value should in fact be 285. On how we got there though, I am not sure that simply implementing `numpy-quaternion` properly fixes the bug. I will open a new issue to discuss.

hakonanes:
> I will open a new issue to discuss.

I didn't mean to say that you shouldn't open the issue, btw. Sorry if that's what it read like.

hakonanes:
> it would be great if unique chose the same fundamental sector on the unit sphere as point_group.fundamental_sector

Actually, I find the assumption that the user wants the vectors within the fundamental sector not strong enough to be used in `Miller.unique()`. We should instead fix `Miller.in_fundamental_sector()` (inherited from `Vector3d`) so that it works, so that the unique ones within the sector can be found by projection followed by rejection.

If you want the direct or reciprocal lattice vectors within the fundamental sector in one call, we should make a convenience method `Miller.from_fundamental_sector()` calling these different methods. Will that meet your needs @harripj?

orix/quaternion/quaternion.py
```diff
@@ -89,39 +90,25 @@ def antipodal(self):
...
-        b, c, d = -self.b.data, -self.c.data, -self.d.data
-        q = np.stack((a, b, c, d), axis=-1)
-        return Quaternion(q)
+        q = quaternion.from_float_array(self.data).conj()
+        return Quaternion(quaternion.as_float_array(q))
 
     def __invert__(self):
         return self.__class__(self.conj.data / (self.norm.data**2)[..., np.newaxis])
 
     def __mul__(self, other):
```
hakonanes, line 99: I suggest to explain the two multiplications performed here in LaTeX in the docstring, so that the understanding of these operations aren't lost when the explicit code is replaced.

orix/quaternion/quaternion.py
```diff
@@ -224,33 +211,15 @@ def outer(self, other):
...
-            sa, oa = self.data[..., 0], other.data[..., 0]
-            sb, ob = self.data[..., 1], other.data[..., 1]
-            sc, oc = self.data[..., 2], other.data[..., 2]
-            sd, od = self.data[..., 3], other.data[..., 3]
-            q[..., 0] = e(sa, oa) - e(sb, ob) - e(sc, oc) - e(sd, od)
-            q[..., 1] = e(sb, oa) + e(sa, ob) - e(sd, oc) + e(sc, od)
-            q[..., 2] = e(sc, oa) + e(sd, ob) + e(sa, oc) - e(sb, od)
-            q[..., 3] = e(sd, oa) - e(sc, ob) + e(sb, oc) + e(sa, od)
-            return other.__class__(q)
+            q1 = quaternion.from_float_array(self.data)
```
hakonanes, line 215: The same goes for the docstring here as well, to write out the equations.

orix/quaternion/quaternion.py
harripj: Good idea, will get on them.

harripj: It looks like conjugation is already defined in the `Quaternion` docstring, if this is what you are thinking?

https://github.com/pyxem/orix/blob/ff6e541cab81cdb99d742ec907d257ef96dcdb75/orix/quaternion/quaternion.py#L46-L51

harripj:
@hakonanes I have just added the equations for quaternion-quaternion and quaternion-vector multiplication to the `Quaternion` doctsring. Let me know if this is what you had in mind?

harripj:
Update: this is waiting on #285, after this the tests should pass.

PR #285, opened 2022-02-17
Implement rounding consistently to 12 decimal places in orix, eg. in `unique()` functions

harripj:
A discussed in #281 there appears to be a bug in `Miller.unique` which subsequently causes a failing test in that PR. As As @hakonanes has confirmed in that PR, this appears to be a floating point bug, and rounding, even to a large precision (14 dp), corrects the bug.

So whilst implementing #281 has shown the bug and corrects it, I am not sure whether we should rely completely on this or whether we should manually implement some rounding in `unique` to be sure we always get the expected behaviour.

Some thoughts on this would be great.

This PR also corrects the test and removes an unneeded line as discussed in #281.

Update:
Rounding, for example in `unique` functions, has been consistently fixed to 12 dp for the repository.

#### Progress of the PR

### Example code
From @hakonanes in #281
https://github.com/pyxem/orix/pull/281#issuecomment-1043086981

#### For reviewers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Changed
 Fixed
 -----
 - `Miller.in_fundamental_sector()` doesn't raise errors.
-- `Miller.unique` now correctly returns unique vectors due to implemented rounding.
+- `Miller.unique()` now correctly returns unique vectors due to implemented rounding.
 
 2022-02-14 - version 0.8.1
 ==========================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,15 @@ this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.htm
 Unreleased
 ==========
 
+Changed
+-------
+- Rounding in functions is now set consistently at 12 dp, eg. `Object3d.unique` and 
+  `Rotation.unique`.
+
+Fixed
+-----
+- `Miller.unique` now correctly returns unique vectors due to implemented rounding.
+
 2022-02-14 - version 0.8.1
 ==========================
 
--- a/orix/base/__init__.py
+++ b/orix/base/__init__.py
@@ -182,11 +182,9 @@ class Object3d:
             The indices of the (flattened) data in the unique array.
 
         """
-        data = self.flatten()._data.round(9)
+        data = self.flatten()._data.round(12)
         data = data[~np.all(np.isclose(data, 0), axis=1)]  # Remove zeros
-        _, idx, inv = np.unique(
-            data.round(4), axis=0, return_index=True, return_inverse=True
-        )
+        _, idx, inv = np.unique(data, axis=0, return_index=True, return_inverse=True)
         obj = self.__class__(data[np.sort(idx), : self.dim])
         obj._data = data[np.sort(idx)]
         if return_index and return_inverse:
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -156,7 +156,7 @@ class Rotation(Quaternion):
                     rotation.improper,
                 ],
                 axis=-1,
-            ).round(6)
+            ).round(12)
         _, idx, inv = np.unique(abcd, axis=0, return_index=True, return_inverse=True)
         idx_argsort = np.argsort(idx)
         idx_sort = idx[idx_argsort]
@@ -196,7 +196,7 @@ class Rotation(Quaternion):
                 i,
             ),
             axis=-1,
-        ).round(6)
+        ).round(12)
         return abcd
 
     def angle_with(self, other):
--- a/orix/sampling/S2_sampling.py
+++ b/orix/sampling/S2_sampling.py
@@ -80,13 +80,14 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
 
     res = np.radians(resolution)
 
+    grid_type = grid_type.lower()
     grid_types = ["normalized", "spherified_edge", "spherified_corner"]
     if grid_type == grid_types[0]:
         grid_length = np.tan(res)
         steps = np.ceil(max_distance / grid_length)
         i = np.arange(-steps, steps) / steps
     elif grid_type == grid_types[1]:
-        steps = np.ceil(max_angle / res)
+        steps = np.ceil((max_angle / res).round(1))
         k = np.arange(-steps, steps)
         theta = np.arctan(max_distance) / steps
         i = np.tan(k * theta)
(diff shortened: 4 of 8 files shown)


hakonanes:
> manually implement some rounding in unique to be sure we always get the expected behaviour

`np.finfo(float).eps` returns 2.220446049250313e-16 on my machine. Rounding to 14 decimals fixed the bug for the example in #281. I'm OK with rounding to 14 decimals. But Miller indices in orix aren't restricted to integers, and can be cartesian coordinates xyz. So we shouldn't round too much. `np.finfo(np.float32).eps` is 1.1920929e-07, this might be a bit too much.

harripj:
From my testing the reason between the two versions ('stable' and 'paddy') in https://github.com/pyxem/orix/pull/281#issuecomment-1043086981 is that at this line in `Miller.unique` https://github.com/pyxem/orix/blob/ff6e541cab81cdb99d742ec907d257ef96dcdb75/orix/vector/miller.py#L675
the `min` and `max` differences between the two arrays are `-1.7763568394002505e-15` and `1.7763568394002505e-15`, which then flows through the rest of the function.

harripj:
The behaviour is the same with the next test https://github.com/pyxem/orix/blob/ff6e541cab81cdb99d742ec907d257ef96dcdb75/orix/tests/test_miller.py#L218

The output size is `205` and stable rounding the data from 0 dp all the way down to 14 dp.

harripj:
From @hakonanes #281:

> np.finfo(float).eps returns 2.220446049250313e-16 on my machine. Rounding to 14 decimals fixed the bug for the example in https://github.com/pyxem/orix/pull/281. I'm OK with rounding to 14 decimals. But Miller indices in orix aren't restricted to integers, and can be cartesian coordinates xyz. So we shouldn't round too much. np.finfo(np.float32).eps is 1.1920929e-07, this might be a bit too much.

I agree, if we decide to round in `unique` functions (we already do in `Rotation.unique`) then I suggest we keep the same value. We could offer this as precision as a function argument, but I suggest we keep it the same for all functions. In `Rotation.unique` it is currently `6`, but as @hakonanes suggested maybe something in the range `12-14` would be more appropriate.

harripj:
The tests for `Rotation.unique` also pass at 12 and 14 dp precision.

hakonanes:
This issue reminds me of https://github.com/mtex-toolbox/mtex/pull/1261, where rounding to 12 decimals of vectors fixed an issue in TSL IPF color keys for some Laue classes.

We shouldn't round too much, but not too little so that we have to fix this bug again. I suggest 12 decimals for arrays of 64 bit floats, in all rounding cases as you say. If users don't set array data types themselves, they should always be 64 bit floats.

> We could offer this as precision as a function argument

We could, but I don't see a use case for it?

harripj:
> I suggest 12 decimals for arrays of 64 bit floats, in all rounding cases as you say.

I agree.

There was a nasty rounding case in the tests. In `sample_S2_cube_mesh`:
https://github.com/pyxem/orix/blob/ff6e541cab81cdb99d742ec907d257ef96dcdb75/orix/sampling/S2_sampling.py#L88-L92

With 5 degree resolution, depending on the rounding the returned value from `ceil` changed by a whole value, ie. for 9dp the value into `ceil` was 9+delta (9.000000003023132) -> 10, but with 12dp rounding it was 9-delta -> 9. I implemented a round to 2 dp before ceil to correct for this case, but another case where we should be careful with rounding. The test was also corrected for this.

It comes down to whether everyone is okay with `ceil(9.000000003023132)` being `9` rather than `10` due to the 2dp rounding.

Perhaps this will also have changed the notebooks? I will check.

hakonanes:
Sorry, should made this https://github.com/pyxem/orix/pull/281#issuecomment-1044200937 here instead.

harripj:
> Actually, I find the assumption that the user wants the vectors within the fundamental sector not strong enough to be used in Miller.unique(). We should instead fix Miller.in_fundamental_sector() (inherited from Vector3d) so that it works, so that the unique ones within the sector can be found by projection followed by rejection.

Am I correct in thinking this can now be done as of #288?
And I think you are right about `unique`, if I am following your thought path then for a cubic material `(-1, 0, 1)` should be considered unique from `(1, 0, 1)` in `unique`, but if we projected to the fundamental sector they would both be considered the same? 

> If you want the direct or reciprocal lattice vectors within the fundamental sector in one call, we should make a convenience method Miller.from_fundamental_sector() calling these different methods. Will that meet your needs @harripj?

It's not necessarily on my cards at the moment, so happy to hold off. I think this can be done explicitly using a combination of `in_fundamental_sector()` followed by `unique` (as per your suggestion)? Maybe best to let the user do that explicitly if that is what they want to do?

hakonanes:
> Am I correct in thinking this can now be done as of https://github.com/pyxem/orix/pull/288

Yes, with that PR and the rounding in `Miller.unique()` from this one

```python
>>> from orix.vector import Miller
>>> diamond = Phase(space_group=227)
>>> h = Miller.from_highest_indices(phase=diamond, uvw=[10, 10, 10])
>>> h2 = h.in_fundamental_sector()
>>> h3 = h2.unique(use_symmetry=False)
>>> h4 = h.unique(use_symmetry=True)
>>> print(h.size, h2.size, h3.size, h4.size)
9260 9260 285 285
>>> fig = h3.scatter(return_figure=True, s=5)
>>> h4.scatter(figure=fig, c="g", s=5)
>>> fig.axes[0].plot(diamond.point_group.fundamental_sector.edges, color="r", zorder=0)
```

with_in_fundamental_sector

CHANGELOG.rst
```diff
@@ -10,6 +10,15 @@ this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.htm
 Unreleased
 ==========
 
+Changed
+-------
+- Rounding in functions is now set consistently at 12 dp, eg. `Object3d.unique` and
```
hakonanes, line 15: I think this is the main change in this PR, I suggest to change the title accordingly.

CHANGELOG.rst
```diff
@@ -10,6 +10,15 @@ this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.htm
...
 ==========
 
+Changed
+-------
+- Rounding in functions is now set consistently at 12 dp, eg. `Object3d.unique` and 
+  `Rotation.unique`.
+
+Fixed
+-----
+- `Miller.unique` now correctly returns unique vectors due to implemented rounding.
```
hakonanes, line 20:
```suggestion
- `Miller.unique()` now correctly returns unique vectors due to implemented rounding.
```

hakonanes (approved):
Good catch, important fix! Two minor suggestions.

pc494:
I'm happy with this, for normal rounding 12/14 decimals place seems fine, ceil bits can be annoying. In general, given how annoying machine specific bugs are I would tend to round too much (and get an error because we rounded - explicit) than too little (and get an error because of machine precision - implicit).

Approve, will leave @hakonanes to merge.

pc494 (approved):

pc494:
Turns out this was needed for the next thing I had to look at so I'm merging myself.

PR #292, opened 2022-02-18
Reduce data precision before creating RGB grid with scipy in IPF

### LINKED ISSUE #290: Rounding lead to extremely slow scipy.interpolate.griddata when plotting IPF color keys
Three latest Read the Docs builds (https://readthedocs.org/projects/orix/builds/16129716/) failed due to timeout (> 900 s) while waiting for nbsphinx to run the `inverse_pole_figures.ipynb` user guide notebook. Turns out the new rounding leads to these arrays

https://github.com/pyxem/orix/blob/25b69ba6eff09a179929f0c6eade10565adbf78a/orix/plot/direction_color_keys/direction_color_key_tsl.py#L111-L113

having shapes (1001, 1000) instead of (1000, 1000) as before. This lead `scipy.interpolate.griddata()` to never finish (within my patience, like 60 s, since it previously took 1-2 s). I'm looking to fix this now, and incorporate in the 0.8.2 patch release.

EDIT: This is not the issue, the issue is the cartesian coordinates of the vectors within the sector:

https://github.com/pyxem/orix/blob/25b69ba6eff09a179929f0c6eade10565adbf78a/orix/plot/direction_color_keys/direction_color_key_tsl.py#L104

I traced it back to this change

https://github.com/pyxem/orix/pull/285/files#diff-dccf6ae132bb72e27f84d29f467e89463c36d42bff2cf3bdf832e29e3d4c79b3L185

Thus, the issue is fixed if we round the resulting cartesian coordinates to a precision lower than 12, for example the previous 9.

hakonanes:

Fix #290. ~Also save some time by using linear instead of cubic interpolation, which gave the same results by visual inspection.~ Cubic was better.

#### Progress of the PR

```python
>>> from orix.plot import IPFColorKeyTSL
>>> from orix.quaternion import symmetry
>>> IPFColorKeyTSL(symmetry.D3d).plot()  # Doesn't take forever
```

#### For reviewers

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,7 +4,7 @@ include LICENSE
 include MANIFEST.in
 include README.rst
 include RELEASE.rst
-include readthedocs.yml
+include readthedocs.yaml
 include setup.cfg
 include setup.py
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,7 +4,7 @@ include LICENSE
 include MANIFEST.in
 include README.rst
 include RELEASE.rst
-include readthedocs.yaml
+include readthedocs.yml
 include setup.cfg
 include setup.py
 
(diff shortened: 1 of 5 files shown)


hakonanes:
Good, this change brought the RTD HTML build back to ~210 s from 445 s with the less strict rounding.

pc494 (approved):
LGTM

PR #293, opened 2022-02-21
Post 0.8.2 release updates

hakonanes:
Typical post-release actions:
* Revert version from 0.8.2 to 0.9.dev0.
* Add URL pointing to Read the Docs documentation in setup.py and remove reference to non-existing *.css files in MANIFEST.in file. These two changes were recommended by the [build check](https://github.com/pyxem/orix/runs/5272913196?check_suite_focus=true#step:5:1) before the package was uploaded to PyPI.
* Add "Unreleased" to changelog

Updates to the PyPI publish action:
Use [PyPA action](https://github.com/marketplace/actions/pypi-publish) and [PyPA `build`](https://pypa-build.readthedocs.io/en/stable/index.html) package. Most importantly, these changes replace the PyPI username/password secrets with a repo specific PyPI API token secret (already generated by me and added to the orix GitHub repo secrets). I've added the possibility to upload to TestPyPI, but this can be removed if we find we don't need it. From the PyPA docs:

> it’ll publish any push to TestPyPI which is useful for providing test builds to your alpha users as well as making sure that your release pipeline remains healthy!

I've added a TestPyPI API token secret specific to my TestPyPI account which we can replace only after orix is uploaded to TestPyPI for the first.

#### Progress of the PR

#### For reviewers

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,7 +1,15 @@
-# This workflows will upload a Python Package using Twine when a release is created
-# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+# This workflows runs when a tagged release is created or it is triggered manually.
+# For more information see:
+# - Python docs: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+# - GitHub action: https://github.com/marketplace/actions/pypi-publish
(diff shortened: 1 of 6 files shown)


pc494 (approved):
LGTM, merging (I don't think we will use TestPyPi much, but also it can't hurt us)

hakonanes:
> I don't think we will use TestPyPi much

No, I only used it when I first uploaded packages to PyPI. I guess release candidates should be released on PyPI, not TestPyPI, so I cannot think of a specific use case for it yet. We can remove it at some later point if unused.

pc494:
I would send release candidates to pypi personally (eg. https://pypi.org/project/numpy/#history)

Issue #249, opened 2021-12-08
Conventions in orix. Mega-Thread


pc494:
Following on from: https://github.com/pyxem/orix/pull/234#issuecomment-988742635

There are lots of conventions needed to work with orientations. We (the developers) are starting to learn that anything other than extremely well documented conventions will lead to problems. This thread aims to spell these conventions out, agree upon them, and also design a strategy for documenting them. I have decided to design it into "Phases" so as not to lead to information overload. Each new phase will be added to the top line to make life easier.

Phase 0 - Setting Terms
--------------------------------
- Where possible we will follow a mixture of the three manuscripts on which orix was built, for compatibility. These are:
[1] D Rowenhorst et al 2015 Modelling Simul. Mater. Sci. Eng. 23 083501
[2] Johnstone et al. J. Appl. Cryst. (2020). 53, 1293–1298
[3] Krakow R et al.. 2017 On three-dimensional misorientation spaces. Proc. R. Soc. A 473: 20170274.

And the computer package MTEX (which we will refer to as [4]).

Phase 1
-----------

- The vast majority of our use cases will involve a sample (specimen) which is a macroscopic object and crystal(s) which are microscopic arrangements of atoms. I suggest the following.

1. The sample z direction goes perpendicular to the array of our 2D sample (obtained through some scanning mechanism)
2. The sample basis (and all such basis) should be right handed [1]

Two options jump out:
a1. Follow [4], x east, y south, z into page.
a2. Stick with what we have, x east, y north, z out of page.

Having decided this we should then decided how the crystal axes for each crystal system 'align' with this basis when no rotation has been applied. This basically involves completing the following table, ideally from a single authoritative source (perhaps noting disagreements with other source as and when needed). Appendix B of [3] suggests this might be a bit tricky. Adjacently, I think this (I think) is the cause of https://github.com/pyxem/orix/issues/236

| Crystal System  |   | x  | y  | z |
|---|---|---|---|---|
| Cubic  |   |   |   |   |
| Hexagonal  |   |   |   |   |
| Tetragonal  |   |   |   |   |
| Trigonal  |   |   |   |   |
| Orthorhombic  |   |   |   |   |
|  Monoclinic |   |   |   |   |
| Trigonal  |   |   |   |   |

So questions. 
q1) a1 or a2?
q2) Do we have an authoritative source to complete this table, if not, which ones do we think are obvious and can be agreed upon (eg. cubic is a bit trivial)?

-------

harripj:
[...]

PR #294, opened 2022-02-23
Directionality in `Euler` methods

harripj:
As discussed in #249 the default `direction` kwarg in `from_euler` methods is currently `crystal2lab`, but should be `lab2crystal`. This is the Bunge convention as described in ref. [1] of #249 and opposite of the convention used by [MTEX](https://mtex-toolbox.github.io/MTEXvsBungeConvention.html). Previously a `convention` parameter was also optional, this has been removed as it is potentially confusing.

This PR aims to make a more consistent and easy-to-follow definition of `Rotation` from Euler angles by deprecating the `convention` parameter and relying only on `direction`, as the two have overlap. `'mtex'` is offered as a supported direction for those users coming from MTEX. `object_type="argument"` has been created `deprecated` and currently removes the argument.

#### Progress of the PR

```python
import numpy as np
e = np.deg2rad((45, 30, 90))
        
r_mtex = Rotation.from_euler(e, direction="mtex")
r_c2l = Rotation.from_euler(e, direction="crystal2lab")
r_l2c = Rotation.from_euler(e, direction="lab2crystal")
r_default = Rotation.from_euler(e)

assert np.allclose(r_default.data, r_l2c.data)
True

assert np.allclose(r_mtex.data, r_c2l.data)
True

assert np.allclose((r_l2c * r_c2l).angle.data, 0)
True

Rotation.from_euler(e, convention='bunge')
VisibleDeprecationWarning: Argument `convention` is deprecated and will be removed in version 1.0. Use `direction` instead.
```

#### For reviewers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,10 +16,15 @@ Changed
   convention by default, ie. `direction="lab2crystal"`. The returned `Rotation` from
   this function may be inverted from prior releases and users are advised to check their
   code.
-- The convention parameter in `from_euler()` methods has been removed in favour of
-  `direction`. The `direction` parameter, in addition to `"lab2crystal"` (now default)
-  and `"crystal2lab"`, now also accepts a convenience argument `"mtex"` which is
-  consistent with the default `"crystal2lab"` direction in `MTEX <https://mtex-toolbox.github.io/MTEXvsBungeConvention.html>`_.
+- The `direction` parameter in `from_euler()` methods, in addition to `"lab2crystal"`
+  (now default) and `"crystal2lab"`, now also accepts a convenience argument `"mtex"
+  which is consistent with the default `"crystal2lab"` direction in
+  `MTEX <https://mtex-toolbox.github.io/MTEXvsBungeConvention.html>`_.
+
+Deprecated
+----------
+- The `convention` parameter in `from_euler()` methods has been deprecated in favour of
+  `direction`. This parameter will be removed in release 1.0.
 
 2022-02-21 - version 0.8.2
 ==========================


hakonanes (changes requested):
I like the solution adopted in this PR. Neat extension of the `@deprecated` decorator for arguments.

Note that this PR changes the API for users who've set the `direction` themselves in `from_euler()`, that is we won't warn those users in a release before making the change, so they can adapt. The default behaviour stays the same (when not setting `direction`) so the API is not changed for users who don't set the `direction`.

I'm OK with releasing this in a 0.8.3 patch to get it out as quickly as possible, so users can adapt before 1.0. Another solution is to leave the meaning of the two `direction` options, but warn that they will swap in 1.0, and release this in 0.8.3.

The PR is WIP, but since I'm looking at it, I'll just mention:
* Should update the changelog with an entry under "Changed" stating that the meaning of the two (in addition to "mtex") options in `direction` has changed
* `removal` should be set to 1.0, since we decided we won't release a 0.9
* Two lines in the tests aren't covered (and can thus be removed?): https://coveralls.io/builds/46820525

harripj:
Yes this is currently a breaking change for those who have specified `direction`. In my mind we could either leave the codebase as is and add a deprecation warning, as you mentioned, to change in 1.0, or make the change now and add a one-time warning that 'direction` has swapped for anyone using `from_euler()` which is removed in 1.0.

I would be happy with either, but favour the breaking patch and warning.

hakonanes:
> I have thought about it more and this goes against semantic versioning. I now thing we should deprecate, warn of the change and make the breaking change in 1.0.

One shouldn't break semver, but I'm OK with doing it if not doing it if the break is "mild" and (1) blocks development or (2) takes unproportional maintainer effort. In this case, I'd say (1, 2) isn't met, so I'm fine with doing the change first in 1.0.

harripj:
I am still a bit confused whether you mean release this in a minor 0.8.3 patch or deprecate now and change for 1.0!

So I have made the breaking change here. I think the deprecation should be done in another PR, and this PR can be left until the 1.0 release.

harripj:
> we should release the warning and/or break in a 0.8.3 patch.

Sounds like a plan! 👍

pc494:
just as an addition on semvar:

"Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable."

[source here](https://semver.org/#spec-item-4).

pc494:
will review this tomorrow to check that I'm 100% on the convention change uses discussion, but am fairly certainly it'll be an easy merge

pc494 (changes requested):
This looks good, just needs a CHANGELOG entry, ideally pointing out that this version will now give different (correct) results.

harripj:
For what it is worth I think that this would be best suited to a `0.9` release that we have been considering https://github.com/pyxem/orix/pull/295#issuecomment-1050296527.

pc494 (approved):
Great :D, will let @hakonanes merge

orix/quaternion/orientation.py
```diff
@@ -463,12 +461,11 @@ def from_euler(
         symmetry : Symmetry, optional
             Symmetry of orientation(s). If None (default), no symmetry
             is set.
-        convention : str
-            Only 'bunge' is currently supported for new data
         direction : str
-            'lab2crystal' or 'crystal2lab'
+            "lab2crystal" or "crystal2lab". If "MTEX" is provided then
```
hakonanes, line 465: Is it useful to specify that "lab2crystal" is the Bunge convention?

Should add a `"lab2crystal" (default)`.

orix/quaternion/rotation.py
```diff
@@ -48,6 +48,7 @@
 from orix.quaternion import Quaternion
 from orix.vector import AxAngle, Vector3d
 from orix.scalar import Scalar
+from orix._util import deprecated
```
hakonanes, line 51: This is now unused.

It seems like you @harripj and @pc494 are in favor of not deprecating the `convention` argument. I've never used the argument myself (the `direction` argument I've used a couple of times), but there might be users who do. They will get a `TypeError` with 0.9. I prefer to emit a warning in 0.9 instead, stating that it will be removed in 1.0. This takes little effort from our end (since you've already done the work, @harripj), we just have to remember to remove the warning in 1.0.

I'm sorry for being vague on what I thought was the best solution here.

I'd appreciate it if you could just quickly state here whether you still want to remove it.

orix/tests/quaternion/test_orientation.py
```diff
@@ -531,6 +531,11 @@ def test_set_symmetry_deprecation_warning_misorientation():
         _ = o.set_symmetry(C2, C2)
 
 
+def test_from_euler_convention_raises():
```
hakonanes, line 534: If we go for deprecation, a comment like "# TODO: Remove in 1.0" would be good.

orix/tests/quaternion/test_rotation.py
```diff
@@ -224,23 +224,36 @@ def test_to_from_euler(self, e):
...
+        r_l2c = Rotation.from_euler(e, direction="lab2crystal")
+        r_default = Rotation.from_euler(e)
+        assert np.allclose(r_default.data, r_l2c.data)
+        assert np.allclose(r_mtex.data, r_c2l.data)
+        assert np.allclose((r_l2c * r_c2l).angle.data, 0)
 
     def test_direction_kwarg(self, e):
         _ = Rotation.from_euler(e, direction="lab2crystal")
 
+    # def test_convention_arg
```
hakonanes, line 241: I'd remove this?

orix/tests/quaternion/test_rotation.py
```diff
@@ -224,23 +224,36 @@ def test_to_from_euler(self, e):
...
         r = Rotation.from_euler(e)
-        with pytest.raises(ValueError, match="The chosen convention is not supported,"):
-            _ = r.to_euler(convention="unsupported")
+        with pytest.raises(TypeError, match=r"to_euler\(\) got an unexpected keyword "):
+            _ = r.to_euler(convention="bunge")
 
-    def test_unsupported_conv_from(self, e):
-        with pytest.raises(ValueError, match="The chosen convention is not one of "):
-            _ = Rotation.from_euler(e, convention="unsupported")
+    def test_conv_from_raises(self, e):
```
hakonanes, line 252: If we don't go for deprecation, I'd remove any test ensuring that passing the `convention` argument raises an error.

orix/tests/test_util.py
```diff
@@ -55,6 +55,52 @@ def bar_func2(self, n):
                 """
                 return n + 2
 
+            @deprecated(
```
hakonanes, line 58: If we don't go for deprecation, these tests test internal utilities that aren't in use. Will they be in the future? If so I think we should keep the functionality. If not, we should remove it. What do you think?

hakonanes (changes requested):
This improves the API and makes it clearer.

I'm sorry for being vague about what I think we should do regarding deprecation. See my comments on this.

Otherwise, I've got a few docstring suggestions and such.

orix/_util.py
```diff
@@ -59,23 +67,42 @@ def __init__(self, since, alternative=None, removal=None, object_type="function"
...
             object_type = "Property"
             parentheses = ""
-        else:
+            name = func.__name__
+        elif self.object_type.lower() == "argument":
+            object_type = "Argument"
+            parentheses = ""
+            if self.argument is None:
+                raise TypeError("argument must be defined for object_type 'argument'.")
+            name = self.argument
```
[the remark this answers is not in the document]
harripj: No you are right, these are only useful for development. In fact this causes a bit of a headache when it comes to testing. As it should only be used for development, perhaps we could just remove the `if` statement, in which case `argument` defaults to `None`, and assume that if used the developer will read the deprecation warning and correct this `None`?

harripj:
@hakonanes @pc494 from my end I am happy to deprecate the `convention` argument in this PR but correct the `lab2crystal` error. That way any use of `convention` will not throw an error. I think that the `direction`s should be corrected in this PR (and not deprecated and warned to change in a future release).

From my end I don't use the other `pyxem` libraries, so this is not much of a headache for me, but I think that you both, as developers for dependent libraries, are better suited to decide whether to deprecate this now and remove later or remove now.

orix/_util.py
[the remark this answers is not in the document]
hakonanes: Yes, I think this is safe to do!

hakonanes:
> @hakonanes @pc494 I am happy to deprecate the `convention` argument in this PR but correct the `lab2crystal` error.

Yes, that is exactly what I meant, sorry again for not being 100% clear on which of the two changes to deprecate.

> From my end I don't use the other `pyxem` libraries, so this is not much of a headache for me, but I think that you both, as developers for dependent libraries, are better suited to decide whether to deprecate this now and remove later or remove now.

Thankfully, we only pass Euler angles to `Rotation.from_euler()` in kikuchipy, so as long as the default behaviour doesn't change, this PR has no effect on that package. It's with regards to end users of orix I'm (mildly) concerned: basically, we haven't broken public API arguments without warning previously (as far as I can tell), and it would be good to keep it that way.

orix/tests/quaternion/test_orientation.py
harripj: Good idea!

harripj:
@pc494 @hakonanes I have updated the PR to deprecate the `convention` argument in `0.9` for removal in `1.0`, as discussed to not break others' code, so want to check this is okay with you both.

pc494:
Broadly (pre-1.0) I lean more towards breaking stuff so I'm absolutely fine with this as implemented and am merging.

Issue #198, opened 2021-06-04
Consider the usage of orix.scalar.Scalar


pc494:
> Actually, I avoid using `orix.scalar.Scalar` wherever I can, because I don't like writing and reading code where the `.data` is 
> needed to get the NumPy array. I haven't looked at the reason why we need `Scalar`, it might a good one, perhaps it enables 
> us to not check for instance shapes or something similar, but I would be interested in seeing if we can remove completely.

_Originally posted by @hakonanes in https://github.com/pyxem/orix/issues/197#issuecomment-853942078_

I think this is a broadly fair assessment and worth looking into. The Scalar class can be a nuisance.

harripj:
As @hakonanes mentioned this again in #264, it may be worth having a discussion here. From having looked at the code I don't know what benefit we get from the `Scalar` class rather than being an extra (unnecessary) abstraction over a NumPy array and extra `.data` calls. Personally I would be in favour of trying to remove it.

PR #295, opened 2022-02-24
Remove Scalar class

harripj:
As discussed in #198, the `Scalar` class is nothing more than a glorified NumPy array and I think it should be removed.
This PR aims to remove the `Scalar` class and protect users from writing `.data`.

To do:
- [x] I have done a repo wide search for `Scalar` and removed it
- [ ] Next in line is to fix `.data` which is now a memory view

#### Progress of the PR

```python
>>> from orix import vector
>>> v = vector.Vector3d([1, 1, 1])
>>> v.data
array([[1, 1, 1]])  # not Scalar
```

#### For reviewers


pc494:
For what it's worth (and despite what I said in the #294) I think this change should bump the minor version (to 0.9.0) or wait for 1.0.0 - my preference is for the former.

orix/plot/crystal_map_plot.py
```diff
@@ -184,7 +183,7 @@ def plot_map(
             for _, p in crystal_map.phases_in_data:
                 patches.append(mpatches.Patch(color=p.color_rgb, label=p.name))
         else:  # Create masked array of correct shape
-            if isinstance(value, (Scalar, Vector3d)):
+            if isinstance(value, Vector3d):
```
harripj, line 186: The failing test is here. In fact it was only checked because `.angle` used to return a `Scalar`. A `Vector3d` was never tested. Is a `Vector3d` a reasonable thing to pass in as `value` to `plot_map`?

harripj:
> For what it's worth (and despite what I said in the https://github.com/pyxem/orix/pull/294) I think this change should bump the minor version (to 0.9.0) or wait for 1.0.0 - my preference is for the former.

Okay with me. The change in this PR in principle changes nothing, although as-written code may have to change to remove `.data` calls, which may be inconvenient to others.

hakonanes (changes requested):
This change reduces amount of necessary code in orix *and* in scripts without loosing functionality, readability or clarity: great!

kikuchipy is the "largest" downstream user of orix as far as I know, and this change makes 22 tests fail in a straightforward way (.data is now a numpy array memoryview). I suggest to release this and a warning from #294 in a 0.9.0rc1 release candidate on PyPI, so that kikuchipy can be updated to work with 0.9.0 upon its release.

diffsims and pyxem seams to be unaffected, as none of their tests failed.

hakonanes:
Could you also remove acceses of `.data` in the [crystal geometry](https://orix--295.org.readthedocs.build/en/295/crystal_geometry.html) and [clustering across FZ](https://orix--295.org.readthedocs.build/en/295/clustering_across_fundamental_region_boundaries.html) notebooks? It runs fine but outputs a memoryview some places. I'll re-run the clustering orientations notebook with this PR and push the updated notebook with cell output.

orix/vector/vector3d.py
```diff
@@ -169,57 +165,53 @@ def polar(self):
...
 
     def __radd__(self, other):
-        if isinstance(other, Scalar):
-            return self.__class__(other.data[..., np.newaxis] + self.data)
-        elif isinstance(other, (int, float)):
+        if isinstance(other, (int, float)):
             return self.__class__(other + self.data)
         elif isinstance(other, (list, tuple)):
             other = np.array(other)
+
```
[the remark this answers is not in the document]
harripj: Yes, in fact I got lost somewhere here in this code until I realised that there were separate `if` blocks. I have kept them separate and also separated the `return` for clarity

hakonanes:
None of the notebooks except for "Clustering orientations" are stored with the cell outputs. If cell outputs are stored, they won't be run by Read the Docs every time the documentation is built. These should be removed.

harripj:
> None of the notebooks except for "Clustering orientations" are stored with the cell outputs. If cell outputs are stored, they won't be run by nbsphinx Read the Docs every time the documentation is built. These should be removed.

Whoops my bad, removed in latest commit.

hakonanes:
This fact should be stated in the contributing guide, so this is on me. I'll add a note about storing cell outputs and push a commit to your branch.

hakonanes:
I've updated the notebook, and the contributing guide explaining the use of cell outputs in our user guide notebooks. I've also moved the contributing guide to the to directory to make it more visible (it's still included in the docs via a stub file with a link).

pc494 (approved):
LGTM, will leave @hakonanes to could confirm and merge :)

doc/crystal_geometry.ipynb
```diff
@@ -737,7 +737,8 @@
    "outputs": [],
    "source": [
     "o = Orientation.from_euler(np.deg2rad([45, 0, 0]))\n",
-    "o = o.set_symmetry(cubic.point_group)\n",
+    "o.symmetry = cubic.point_group\n",
+    "o = o.map_into_symmetry_reduced_zone()\n",
```
hakonanes, line 741: I suggest changing these to just one call (it's irrelevant whether the orientation angle is the smallest) `o = Orientation.from_euler(np.deg2rad([45, 0, 0]), symmetry=cubic.point_group)`

doc/crystal_geometry.ipynb
```diff
@@ -800,7 +801,8 @@
    "outputs": [],
    "source": [
     "o2 = Orientation.from_euler(np.deg2rad([10, 20, 30]))\n",
-    "o2 = o2.set_symmetry(trigonal.point_group)\n",
+    "o2.symmetry = trigonal.point_group\n",
+    "o2 = o2.map_into_symmetry_reduced_zone()\n",
```
hakonanes, line 805: Same here: `o2 = Orientation.from_euler(np.deg2rad([10, 20, 30]), symmetry=trigonal.point_group)`

orix/quaternion/orientation.py
```diff
@@ -553,41 +552,47 @@ def angle_with(self, other):
...
+        all_dot_products = Rotation(misorientation).dot_outer(symmetry)
         highest_dot_product = np.max(all_dot_products, axis=-1)
-        return Scalar(highest_dot_product)
+        return highest_dot_product
 
     def dot_outer(self, other):
         """Symmetry reduced dot product of every orientation in this
-        instance to every orientation in another instance, returned as
-        :class:`~orix.scalar.Scalar`.
+        instance to every orientation in another instance.
```
hakonanes, line 581: Add a `Returns` block here as well?

orix/vector/vector3d.py
```diff
@@ -274,22 +270,20 @@ def dot(self, other):
 
         Returns
         -------
-        Scalar
+        nnumpyp.ndarray
```
hakonanes, line 273: I'd fix this one

hakonanes (changes requested):
Thank you for taking the time to make this PR. I like it a lot, as I've said before. I've got a couple of very minor requests.

orix/quaternion/orientation.py
harripj: I think there is already one below!

PR #301, opened 2022-03-01
Remove `orix.scalar` and `Scalar` class

harripj:
I made a fairly annoying mistake where I deleted my fork instead of deleting my branch after the last PR. I am sorry for this as it has made a bit of a mess. On the plus side I had a local copy of the code so reinstated my fork but now need to make new PR.

This PR is a direct follow on from #295.

#### Progress of the PR

#### For reviewers


CHANGELOG.rst
```diff
@@ -12,6 +12,8 @@ Unreleased
 
 Changed
 -------
+- `orix.scalar.Scalar` class has been removed and the data held by `Scalar` is now
```
hakonanes, line 15: This should be moved to a "Removed" section as per [Keep a Changelog](https://keepachangelog.com/en/1.1.0/#how).

orix/tests/test_misc.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
```
hakonanes, line 1: Why is this test needed?

hakonanes (changes requested):
Fantastic PR. Another benefit of removing `Scalar` is that the code is easier to read and write.

Have to two suggestions only, one request and one question.

No problem on my end from you deleting your fork. The git diff is still available in the [old PR](https://github.com/pyxem/orix/pull/295/files), even though it isn't possible to re-open it, so if you didn't have a local copy, you would still be able to copy/paste the contents of the files in the old PR to make a new one.

CHANGELOG.rst
harripj: Good spot!

orix/tests/test_misc.py
harripj: Sorry, didn't realize that this was there. VS Code must have auto-added the file encoding when I made the file.

hakonanes: I mean the test, why is it needed?

harripj: From a clean build I don't think it is, however I ran into an error where the scalar.py class had been removed but was still cached, so it would actually try to import. You are right though, we shouldn't need this for the build. I will remove it.

harripj: It was cached in `__pycache__` in the `orix.scalar` directory, but I have completely removed this module in the PR. I haven't come across the problem since deleting the directory (and so `__pycache__`).

hakonanes: OK, thanks for checking, let's this doesn't become an issue for others.

hakonanes (approved):
I'm happy with this removal. Since I've contributed to this PR, I'll leave it up to @pc494 to merge.

Issue #254, opened 2021-12-16
Accomodating symmettry (which applies to Orientations but not Rotations) in the inheritance structure


pc494:
So in #252 and #253 we've run into some issues with passing symmetry through `Orientations` when inherited methods are involved. My strong feeling is that we shouldn't rely on duplicating code in the medium-long run. My proposed solution is to add a symmetry property to the `Rotation` class as follows:

```python
@property
def symmetry(self):
   raise warnings.warn('Rotation objects do not have symmetries but you have access one.')
   return None


@symmetry.setter
def symmetry(self, value):
        if value is not None:
            raise warnings.warn('Rotation objects do not have symmetries but you have attempted to assign one.')
        else:
            pass
```

Then all of the methods in `Rotation` can be refactored to allow symmetry with an `if sym_value is not None` clause, for example:

```python
def flatten(self,symmetry=None):
        """A new object with the same data in a single column."""
        r = super().flatten()
        r.improper = self.improper.T.flatten().T
        if symmetry is not None:
              r._symmetry = self._symmetry
        return r
```

Feedback welcome.

hakonanes:
> My strong feeling is that we shouldn't rely on duplicating code in the medium-long run.

I agree, more code means more room for error.

> My proposed solution is to add a symmetry property to the `Rotation` class

Could this be done simply by adding a `_symmetry = None` to the class, similar to `Quaternion`'s `dim = 4`, since this is the attribute (not `symmetry` without underscore) we're setting? The underscore shows it's not meant to be used.

> Then all of the methods in `Rotation` can be refactored to allow symmetry with an `if sym_value is not None` clause

This looks like a good solution for methods which don't need a `symmetry` *parameter* if it were added to `Orientation`, like the ones in #252. However, this does not apply to the methods in #253, since `Rotation.from_axes_angles()` and `Orientation.from_axes_angles()` should have unique method signatures (only the latter allows `symmetry`) and docstrings.

What do you think @pc494?

harripj:
@pc494 sounds like a good idea. I agree with what @hakonanes has suggested here. A private `_symmetry = None` class attribute for `Rotation` seems okay but I don't think an outward facing property is needed (and could be confusing).

After a quick test I think that a small adaptation of the OP would work for both `Rotation` and `Orientation` classes if `Rotation` has a `_symmetry = None` class attribute (I think @hakonanes this is what you were also thinking?):
```python
def flatten(self,symmetry=None):
[...]

PR #302, opened 2022-03-01
Add `Orientation.angle_with_outer()` which is based on `get_distance_matrix()`

harripj:
As mentioned in #301 I made the mistake of deleting my fork. As a result a new PR must be made from the unmerged branch.

This PR is a direct follow on from #298 .

#### Progress of the PR

#### For reviewers


orix/quaternion/orientation.py
```diff
@@ -561,47 +561,10 @@ def angle_with(self, other):
...
-    )
-    def distance(self, verbose=False, split_size=100):
-        return super().distance(verbose=verbose, split_size=split_size)
-
-    def get_distance_matrix(self, lazy=False, chunk_size=20, progressbar=True):
+    def angle_with_outer(self, other, lazy=False, chunk_size=20, progressbar=True):
         r"""The symmetry reduced smallest angle of rotation transforming
-        each orientation in this instance to every other orientation.
+        every orientation in this instance to every orientation in
+        another instance.
```
hakonanes, line 567: Could you add

```python
See also
--------
angle_with_outer
```

above "Notes" and a simple usage example in an "Examples" section below "Notes"?

hakonanes (changes requested):
Very good re-use of existing functionality. Also good additions to the test by checking the diagonal.

Non-inline requests:
* Add method to API reference
* Add note to "Added" in changelog

We must remember to update "Returns" to a NumPy array if the PR removing the scalar module is merged before this.

Since this computation can be memory intensive, I think we should still allow `lazy`, `chunk_size` and `progressbar`, even though it is not possible with `angle_with()` etc. We should keep in mind that supporting more operations with `dask` to keep memory in check could be necessary in the future.

harripj:
I have just added `angle_with_outer` to `Rotation` as well, as it was missing.

It has brought up an interesting problem, however. Due to the different implementations of `dot_outer` between `Orientation` and `Rotation`, the shape of the output is inconsistent. We should address this. This comes back in part to @pc494's previous comment that we should reuse code where possible #254.

An example of the current problem:
```python
>> from orix.quaternion import Orientation Rotation
>>> r1 = Rotation.random((5,))
>>> r2 = Rotation.random((6,))
>>> dist_r = r1.angle_with_outer(r2)
>>> dist_r.shape
(5, 6)

>>> o1 = Orientation(r1)
>>> o2 = Orientation(r2)
>>> dist_o = o1.angle_with_outer(o2)
>>> dist_o.shape
(6, 5)

>>> np.allclose(dist_r.data, dist_o.data.T)
True
```

@hakonanes @pc494 what do you think? Should the shape be `(self.shape, other.shape)` or the other way around?

hakonanes:
> Should the shape be (self.shape, other.shape) or the other way around?

Well spotted. I suggest to use the shape returned by the parent class `Rotation`, since this was implemented first. A test assuring that shapes returned from equivalent `Rotation` and `Orientation` methods would be good in this case.

orix/quaternion/orientation.py
```diff
@@ -551,59 +551,20 @@ def angle_with(self, other):
 
         Returns
         -------
-        numpy.ndarray
+        Scalar
```
hakonanes, line 554: These can now be updated since #301 is merged.

CHANGELOG.rst
```diff
@@ -10,6 +10,13 @@ this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.htm
 Unreleased
 ==========
 
+Added
```
hakonanes, line 13: We should add a note to "Fixed" explaning that results from `Orientation.dot_outer()` are now `self.shape + other.shape`, in line with results shape from `Rotation.dot_outer()`. I think "Fixed" instead of "Changed" because the shapes from both classes should be equal.

hakonanes (changes requested):
Another very good addition to the rotation classes. I think this is in good shape except for my two minor requests.

I anticipate to approve and merge after these two have been addressed.

CHANGELOG.rst
harripj: Good idea, I will add it.

CHANGELOG.rst
```diff
@@ -27,6 +34,11 @@ Deprecated
   deprecated, in favour of `direction` in the former. This parameter will be removed in
   release 1.0.
 
+Fixed
+-----
+- The results from `Orientation.dot_outer()` are now now returned as
```
hakonanes, line 39:
```suggestion
- The results from `Orientation.dot_outer()` are now returned as 
```

hakonanes (approved):
I find this a very good addition! Merging.

Issue #296, opened 2022-02-25
Consider the use of `unique` for identical symmetries in `Orientation.dot`


harripj:
Consider the case where one wishes to find the angle between two `Orientation` with the same symmetry, say `Oh`. The following calculation is currently performed: https://github.com/pyxem/orix/blob/d474f246fedd1fabc44f65f103db7ee096759a53/orix/quaternion/orientation.py#L570-L575

In my testing on my machine the total time for this operation is `~5.5ms`, however around `5ms` of this is the `unique` operation. Perhaps we introduce a test where the `unique` function is only performed if the symmetries are different.

Further to this, if we compute the maximum angle after the symmetry unique operation, we find that nothing changes:
```python
from orix.quaternion import symmetry

for name in dir(symmetry):
    s = getattr(symmetry, name)
    if isinstance(s, symmetry.Symmetry):
        u = s.outer(s).unique()
        if s.size != u.size:
            print(name, u, s)
        else:
            delta = (s * ~u).angle.data
            print(name, np.allclose(delta, 0))
```
```
C1 True
C2 True
C2h True
C2v True
C2x True
C2y True
C2z True
C3 True
C3h True
C3v True
C3x True
C3y True
C3z True
C4 True
C4h True
C4v True
C4x True
C4y True
C4z True
C6 True
C6h True
C6v True
Ci True
Cs True
Csx True
Csy True
Csz True
D2 True
D2d True
D2h True
D3 True
D3d True
D3h True
D3x True
D3y True
D4 True
D4h True
D6 True
D6h True
O True
Oh True
S4 True
S6 True
T True
Td True
Th True
_cubic Symmetry (3,) 
[[ 1.   0.   0.   0. ]
 [ 0.5  0.5  0.5  0.5]
 [-0.5  0.5  0.5  0.5]] Symmetry (2,) 
[[1.  0.  0.  0. ]
 [0.5 0.5 0.5 0.5]]
_mirror_xy True
```

which seems to be unnecessary computation apart from `_cubic` symmetry.

What I would suggest then is to add a check to see whether to the two symmetries are the same, and if they are, avoid the unique computation.

On this note, there is an `angle_with` function and I think it would be a good idea to include an `angle_with_outer` function too.

harripj:
As some symmetries are related, perhaps a different test would be:
```python
s1, s2 = symmetry.Oh, symmetry.Th

axis = np.array((s1.size, s2.size)).argmax()
[...]

PR #303, opened 2022-03-01
Speed up `dot` and `dot_outer` by not calling `unique` for identical `Symmetry`

harripj:
As discussed in #296. The use of `unique` is slow and causes a significant performance hit when not needed. Currently in `Orientation.dot` and `.dot_outer` unique values of the two symmetries are computed, however when the two `Symmetry` are identical, eg. both `Th`, then this call to unique is not needed. I suggest we add a check to the function to test whether the two `Symmetry` are unique, and if so avoid the `unique` call. This may be further extended to related symmetries, eg. `Oh` and `Th`.

#### Progress of the PR

```python
from orix.quaternion import Orientation, Rotation, symmetry
import numpy as np

o1 = Orientation.random((6, 4))
o2 = Orientation.random((6, 4))

o1.symmetry = symmetry.Th
o2.symmetry = symmetry.Th

def dot(o1, o2):
        """Symmetry reduced dot product of orientations in this instance
        to orientations in another instance, returned as
        :class:`~orix.scalar.Scalar`.
        See Also
        --------
        dot_outer
        """
        symmetry = o1.symmetry.outer(o2.symmetry).unique()
        misorientation = o2 * ~o1
        all_dot_products = Rotation(misorientation).dot_outer(symmetry).data
        highest_dot_product = np.max(all_dot_products, axis=-1)
        return highest_dot_product

def dot_skip_unique(o1, o2):
        """Symmetry reduced dot product of orientations in this instance
        to orientations in another instance, returned as
        :class:`~orix.scalar.Scalar`.
        See Also
        --------
        dot_outer
        """
        if o1.symmetry == o2.symmetry:
                symmetry = o1.symmetry
        else:
                symmetry = o1.symmetry.outer(o2.symmetry).unique()
        misorientation = o2 * ~o1
        all_dot_products = Rotation(misorientation).dot_outer(symmetry).data
        highest_dot_product = np.max(all_dot_products, axis=-1)
        return highest_dot_product

>>> d = dot(o1, o2)
>>> dsu = dot_skip_unique(o1, o2)
>>> np.allclose((d - dsu).angle.data, 0)
True

>>> %timeit dot(o1, o2)
1.77 ms ± 107 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

>>> %timeit dot_skip_unique(o1, o2)
280 µs ± 8.43 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

#### For reviewers


orix/quaternion/orientation.py
```diff
@@ -570,7 +570,11 @@ def dot(self, other):
         --------
         dot_outer
         """
-        symmetry = self.symmetry.outer(other.symmetry).unique()
+        # this is checking the symmetry ids
```
hakonanes, line 573: I suggest we add a `Symmetry.__eq__` method which returns True if `data ` and `improper` are equivalent in a try/except? Checking `data` with `numpy.allclose()` will ensure that the shapes are identical as well. I don't think the `name` needs to be identical.

hakonanes (changes requested):
This is a great speed-up and good additions to the tests.

In addition to the other requests, could you update the changelog?

orix/quaternion/orientation.py
harripj: Perhaps this can be added to the `Rotation` class instead?

hakonanes: Yes, even better.

orix/tests/quaternion/test_symmetry.py
```diff
@@ -41,6 +42,24 @@ def vector(request):
     return Vector3d(request.param)
 
 
+def _all_symmetries():
+    # fmt: off
+    symmetries = [C1, Ci,  # triclinic
```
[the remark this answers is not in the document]
harripj: Thanks for this, much cleaner.

harripj:
In fact I believe this can be generalised further by checking whether either `Symmetry` exists in the subgroup of the `other`.

```python
all_sym_sg = []

for s1 in symmetry._groups:
    res = []
    for s2 in s1.subgroups:
        # outer of symmetry with its subgroups
        u = s1.outer(s2).unique()
        # assert that unique is same size as main symmetry
        if u.size != s1.size:
            res.append(False)
        else:
            # check that there is no difference between unique
            # and main symmetry
            diff = (s1 * ~u).angle.data
            if np.allclose(diff, 0):
                res.append(True)
            else:
                res.append(False)
    all_sym_sg.append(all(res))
all(all_sym_sg)
True
```

pc494 (commented):
I've got no comments on this, the coverage has dropped though - happy to let @hakonanes lead reviewing on this.

harripj:
> I've got no comments on this, the coverage has dropped though.

Thanks, I will get on it.

harripj:
I have written a function that aims to test whether one symmetry exists within the subgroup of the other, and hence avoid the unique operation:

```python
def get_unique_symmetry_elements(sym1, sym2):
     """Returns the unique symmetry elements between two `Symmetry`."""
     if sym1 == sym2:
         return sym1
     # test whether either symmetry is in the subgroup of the other
     sym2_in_sym1_sg = True if sym2 in sym1.subgroups else False
     sym1_in_sym2_sg = True if sym1 in sym2.subgroups else False
     # if either one of in_sym1 or in_sym2 is satisfied then return the
     # larger parent group
     if sym2_in_sym1_sg + sym1_in_sym2_sg == 1:
         if sym2_in_sym1_sg:
             return sym1
         else:  # in sym2
             return sym2
     else:
         # either no common parent
         # or both in_sym1 and in_sym2 are True, in which case ambiguous
         return sym1.outer(sym2).unique()
```

This works in the large majority of cases, however, for example, in the case of some symmetries:
```python
>>> sym1 = symmetry.C2h
>>> sym2 = symmetry.S4

>>> sym1 in sym2.subgroups, sym2 in sym1.subgroups
True, True

>>> sym1.outer(sym2).unique()
Symmetry (4,) 
[[ 1.  0.  0.  0.]
 [ 0.  0.  0.  1.]
 [ 0.  0.  0.  1.]
 [-1.  0.  0.  0.]]

>>> sym2.outer(sym1).unique()
Symmetry (4,) 
[[1. 0. 0. 0.]
 [0. 0. 0. 1.]
 [1. 0. 0. 0.]
 [0. 0. 0. 1.]]
```
ie. the `unique` value between these sets of symmetry operations is order dependent. Any thoughts?

pc494:
```
>>> sym1 in sym2.subgroups, sym2 in sym1.subgroups
True, True
```

This feels like it's probably at least part of the bug as I doubt two groups can both be subgroups of one another.

hakonanes:
I'm wondering if we should focus on the operations themselves in this PR and leave the equality in subgroups for a later PR?

> the unique value between these sets of symmetry operations is order dependent. Any thoughts?

Hm, sorting of operations can be done using [`numpy.lexsort()`](https://numpy.org/doc/stable/reference/generated/numpy.lexsort.html#numpy.lexsort):

```python
>>> import numpy as np
>>> from orix.quaternion.symmetry import C2h, S4
>>> C2h[np.lexsort(C2h.data)]
Symmetry (4,) 
[[ 1.  0.  0.  0.]
 [ 0.  0.  0.  1.]
 [ 0.  0.  0.  1.]
 [-1.  0.  0.  0.]]
>>> S4[np.lexsort(S4.data)]
Symmetry (4,) 
[[0. 0. 0. 1.]
 [1. 0. 0. 0.]
 [1. 0. 0. 0.]
 [0. 0. 0. 1.]]
```

Don't know whether this helps, though.

harripj:
> This feels like it's probably at least part of the bug as I doubt two groups can both be subgroups of one another.

This makes sense to me. In fact the `_differentiator` of both symmetries are the same, which doesn't appear to make sense.

Looking at the symmetry stereographic projections halfway down [this page](http://xrayweb.chem.ou.edu/notes/symmetry.html) it appears as through the rotations between `S4` and `C4` should be the same (different of course as to whether the rotations are improper (circles vs crosses in figure)). The rotational elements of `C2h` do not look as though they match `S4`, so I am surprised that the differentiators between `S4` and `C2h` are identical.

hakonanes:
Thanks for making this notebook. It is the best visualization of the symmetry operations we have so far, and I suggest to add this as the default projection of a `Symmetry.plot()` method. What do others think?

What do you think about renaming the axes to "a" and "b" instead of "X" and "Y", since the operations are expressed in the crystal direct lattice reference system?

Now that others have started contributing more user guide notebooks, I suggest we restructure it as suggested in #248. What do you think?

harripj:
It looks like the answer was already written in an example in the comments: https://github.com/pyxem/orix/blob/0d975204a2298f59b71bb599524b62ab0b45d308/orix/quaternion/symmetry.py#L375-L377

Except that is the 4-fold rotoinversion around z is combined with `C2`, we get to what I believe is the expected result:
```python
>>> _4fold_rotoinversion_z = Symmetry([(1, 0, 0, 0), (0.5**0.5, 0, 0, 0.5**0.5)])
>>> _4fold_rotoinversion_z.improper = [0, 1]

>>> S4 = Symmetry.from_generators(C2, _4fold_rotoinversion_z)
>>> S4
Symmetry (4,) 
 [ 0.      0.      0.      1.    ]
 [ 0.7071  0.      0.      0.7071]
 [-0.7071  0.      0.      0.7071]]

>>> S4.improper
array([False, False,  True,  True])

>>> S4.subgroups
[Symmetry (1,) 1
 [[1. 0. 0. 0.]],
 Symmetry (2,) 112
 [[1. 0. 0. 0.]
  [0. 0. 0. 1.]],
 ```

and itself is of course part of its subgroup.

File: orix/quaternion/symmetry.py [hunk 3]
```diff
@@ -837,3 +838,23 @@ def _get_laue_group_name(name):
         return "m-3m"
     else:
         return None
+
+
+def get_unique_symmetry_elements(sym1, sym2):
```
hakonanes, line 843: Should this be public? If you don't anticipate to use it directly yourself, I say no. If yes, then a complete docstring explaining parameters and returns must be added. It should also then be added to the list of functions in the `symmetry` module in the reference API.

orix/tests/quaternion/test_orientation.py
```diff
@@ -661,3 +661,31 @@ def test_set_symmetry_deprecation_warning_misorientation():
     o = Misorientation.random((3, 2))
     with pytest.warns(np.VisibleDeprecationWarning, match="Function `set_symmetry()"):
         _ = o.set_symmetry(C2, C2)
+
+
+# TODO: remove in 1.0
+def test_from_euler_convention_mtex():
```
hakonanes: Why is this test added?

orix/tests/quaternion/test_symmetry.py
```diff
@@ -31,11 +31,11 @@
     C4, S4, C4h, D4, C4v, D2d, D4h,  # tetragonal
     C3, S6, D3x, D3y, D3, C3v, D3d,  # trigonal
     C6, C3h, C6h, D6, C6v, D3h, D6h,  # hexagonal
-    T, Th, O, Td, Oh,  # cubic
+    T, Th, O, Td, Oh, get_unique_symmetry_elements,  # cubic
```
hakonanes, line 34: Put this on the next line, since it isn't a cubic point group.

orix/tests/quaternion/test_symmetry.py
```diff
@@ -196,6 +201,41 @@ def test_symmetry(symmetry, vector, expected):
     assert set(vector_calculated) == set(expected)
 
 
+def test_same_symmetry_unique(all_symmetries):
+    symmetry = all_symmetries
```
hakonanes, line 205: Could you explain how this fixture works? Is `symmetry == _groups`, or one element in `_groups`?

hakonanes (changes requested):
The speed-up and much more robusts tests are great!

I have a couple of questions. Also, could you add to the changelog stating that the S4 symmetry has been corrected?

orix/tests/quaternion/test_symmetry.py
harripj: This is one element in groups.

orix/tests/quaternion/test_symmetry.py
harripj: Thanks, will do.

orix/tests/quaternion/test_symmetry.py
hakonanes: Does the fixture operate like `pytest.mark.parametrize()`, when a list is passed as `params`? This test is then executed number of times equal to elements in `_groups`?

orix/quaternion/symmetry.py
hakonanes: Fantastic!

orix/tests/quaternion/test_symmetry.py
harripj: Yes pretty much, this allows the parameterization to be written once and then used for each test which calls the fixture. https://docs.pytest.org/en/6.2.x/fixture.html#parametrizing-fixtures

hakonanes: Cool, I haven't used a fixture in this way. Thanks for the explanation.

hakonanes (approved):

harripj:
Just a note on performance:
```python
>>> import orix
>>> %timeit Oh == Oh
58.8 µs ± 784 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

>>> %timeit Oh in Oh.subgroups
10 ms ± 124 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

>>> %timeit Oh.outer(Oh).unique()
3.25 ms ± 36.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Checking subgroups is currently detrimental to perfoomance, I will add a boolean parameter to the `_get_unique_symmetry_elements` function to define whether subgroups are tested. This performance hit is probably due to computation of `_differentiators` and `subgroups` at run time. Perhaps these could be computed at instantiation in the future, which may help.

hakonanes:
> Perhaps these could be computed at instantiation in the future, which may help.

We would then have to make `Symmetry.data` immutable (unchangable). Alternatively, we could try to implement a check that updates `subgroups` whenever `data` changes. I don't know how to do such a thing, but it might be worth looking into it.

> Checking subgroups is currently detrimental to perfoomance

I assume the timings in the top comment still hold, roughly?

harripj:
> We would then have to make Symmetry.data immutable (unchangable).

This I am okay with, it is essentially the premise with which we implemented the hash. I'm not considering user-created symmetries for now.

> I don't know how to do such a thing, but it might be worth looking into it.

I suspect that this feature could be implemented with a setter, at least for the whole array.

> I assume the timings in the top comment still hold, roughly?

Yes, for identical symmetries. I think for this PR we can keep this identical symmetry check for the performance benefit. `_get_unique_symmetry_elements` by default now computes whether the symmetries are identical, and if not explicitly computes the `unique()` as at the moment this is faster than checking for a subgroup relationship. I have left the subgroup checking code there in the case where we could use it in the future.  Otherwise the implemented tests are of benefit for the codebase.

orix/quaternion/symmetry.py
```diff
@@ -837,3 +838,43 @@ def _get_laue_group_name(name):
...
+    To improve computation speed some checks are performed prior to
+    explicit computation of the unique elements. If `sym1 == sym2`
+    then the unique elements are just the symmetries themselves, and so
+    are returned. If `sym2` is a `subgroup` of `sym1` then the unique
+    symmetry elements will be identical to `sym1`, in this case `sym1`
+    is returned. This check is made if `check_subgroups` is True. As the
+    symmetry order matters, this may not be the case if `sym1` is a
+    subgroup of `sym2`, so this is not checked here.
+
+    If no relationship can is determined between the symmetries then the
```
hakonanes, line 856:
```suggestion
    If no relationship is determined between the symmetries then the
```

orix/tests/quaternion/test_symmetry.py
```diff
@@ -394,16 +443,70 @@ def test_get_point_group():
...
+
+
+def test_two_symmetries_are_not_in_each_others_subgroup(all_symmetries):
+    # if given two symmetries, test that both do not exist in the
+    # subgroup of the other
+    sym1 = all_symmetries
+    # identify place in list by name, cannot test symmetry directy as D3
+    # and D3x are the same and causes an index issue
+    i = [s.name for s in _groups].index(sym1.name)
+    if i + 1 == len(_groups):
```
hakonanes, line 467: I find this if/else clause a bit strange, can't we just do `if i + 1 != len(_groups):` and then do `assert not any(v == 2 for v in values)` as the only assert?

harripj: Good idea, we don't need the `if else`.

hakonanes: What I meant was for the assert to be within the if clause, that should make pytest happy, right?

harripj: Ah okay! 👍

hakonanes: `pytest` doesn't require anything to be asserted in a test, so the else *could* be left out entirely. But of course the solution you chose now is OK.

hakonanes:
I'm very happy with this PR, and I'll leave it up to @pc494 to review and merge when he finds the time.

pc494 (approved):
LGTM, merging

PR #305, opened 2022-03-02
Add equality operator to `Rotation` and `hash` to `Symmetry`

harripj:
As discussed in #303 we should override `Rotation.__eq__` to better test for identical rotations. This will involve checking that the `Rotation` `shape`, `data`, and `improper` are all identical (or close in the case of `data`). This, however means that `__hash__` must be overridden ie. for certain [hash-based tests](https://github.com/pyxem/orix/blob/d7f15cfe2490bf7e799b205ba1cc1726fba0a211/orix/tests/quaternion/test_symmetry.py#L238).

I do not think that `Rotation` should be hashable, as it is by definition immutable. I think that it is fair that `Symmetry` are hashable however, as their data should not change. What I propose is a simple key-based hash system, where by the `Symmetry` hash is `hash(tuple(symmetry.name, size))`, my feeling is that this should be quick and robust enough for the little amount of hashing typically performed.

#### Progress of the PR

```python
>>> from orix.quaternion import symmetry, Rotation
>>> hash(symmetry.C1)
-4311019315580370579

>>> hash(symmetry.Oh)
-295940913969180145

>>> from copy import deepcopy
>>> hash(deepcopy(symmetry.Oh)) == hash(symmetry.Oh)
True

>>> r1 = Rotation.random((5,))
>>> r1_copy = Rotation(r1)
>>> r2 = Rotation.random((5,))
>>> assert r1 == r1
True
>>> assert r1_copy == r1
True
>>> # test values not equal
>>> assert r1 != r2
True
>>> # test improper not equal
>>> r1_copy.improper = ~r1.improper
>>> assert r1_copy != r1
True
>>> # test shape not equal
>>> r3 = Rotation.random((5, 1))
>>> assert r3 != r1
True
```

#### For reviewers


orix/tests/quaternion/test_orientation.py
```diff
@@ -307,6 +307,12 @@ def test_symmetry_property_wrong_number_of_values_misorientation(error_type, val
         o.symmetry = value
 
 
+@pytest.mark.parametrize("symmetry", [C1, C2, C4, D6, T, Oh])
+def test_equality(symmetry):
+    # test that inherited equality is properly tested
+    assert Orientation(symmetry) == symmetry
```
hakonanes, line 313: Hm, do we want this to be true? The orientation instance has in general some symmetry attached, which should be used to check whether two orientation instances are equal. I think we should override the equality method in `Orientation` to also check the `.symmetry` attribute.

harripj: Hmm yes good point, I hadn't thought of overriding it for `Orientation`, but you are right this makes sense. Two `Orientation` should only be equal if `Rotation.__eq__` is `True` and  _also_ if their symmetries are the same. I think the test you highlighted above should be `assert Rotation(symmetry) == symmetry`, I wanted to check that despite the fact the two instances are different classes (although one is inherited) that the equality check passes.

harripj:
> Does this mean that a Symmetry with name "m-3m" and size 48, irrespective of operations, has the same hash as Oh? And will then equal each other when comparing sets of symmetries?

It does. Another option could be `hash(symmetry.data.tobytes())`, which hashes the data directly and is probably more robust, although involves copying the data.

hakonanes:
> It does. Another option could be `hash(symmetry.data.tobytes())`, which hashes the data directly and is probably more robust, although involves copying the data.

I think the operations should be considered when considering the equality of two symmetry instances. The operations are small in bytes, as you say, so including them in the hash seems like an OK solution to me.

harripj:
> I think the operations should be considered when considering the equality of two symmetry instances. The operations are small in bytes, as you say, so including them in the hash seems like an OK solution to me.

I agree. Perhaps we should also include `.improper` in this hash, what do you think?

hakonanes:
Yes, adding .improper should complete the unique description of a symmetry instance.

orix/quaternion/orientation.py
```diff
@@ -145,6 +145,17 @@ def __getitem__(self, key):
...
 
+    def __eq__(self, other):
+        v1 = super().__eq__(other)
+        if not v1:
+            return v1
+        else:
+            # check symmetries are also equivalent
+            v2 = [
+                sym_s == sym_o for sym_s, sym_o in zip(self._symmetry, other._symmetry)
+            ]
```
hakonanes, line 156: Given that this operation must be spread over three lines anyway, I suggest to replace the list comprehension with standard list decleration and looping.

hakonanes: To increase readability.

orix/quaternion/rotation.py
```diff
@@ -118,6 +118,19 @@ def __invert__(self):
...
+        if not isinstance(other, Rotation):
+            return False
+        # only return equal if shape, values, and improper arrays are equal
+        if self.shape != other.shape:
+            return False
+        if not np.allclose(self.data, other.data):
+            return False
+        if not np.allclose(self.improper, other.improper):
+            return False
+        return True
```
hakonanes, line 132: I'd argue for the readability to increase if this was written as

```python
if (
    not isinstance(other, Rotation)
    or self.shape != other.shape
    or not np.allclose(self.data, other.data)
    or not np.allclose(self.improper, other.improper)
):
    return False
else:
    return True

hakonanes:
I'd say [the failing test](https://github.com/pyxem/orix/runs/5479565736?check_suite_focus=true#step:9:83), unrelated to this PR, is OK to change to be equal to 0 within a precision of 1e-6. If you agree, could you add this change?

orix/quaternion/rotation.py
harripj: Yes, in this case I think I will invert the negation as well, so the `if` statement is `and` rather than `not or`

harripj:
> I'd say [the failing test](https://github.com/pyxem/orix/runs/5479565736?check_suite_focus=true#step:9:83), unrelated to this PR, is OK to change to be equal to 0 within a precision of 1e-6. If you agree, could you add this change?

I agree, looks to be rounding errors, I have added a tolerance to the test. I suppose that is the penalty for using random data in tests.

orix/quaternion/rotation.py
hakonanes: Yes, that is much better!

hakonanes (approved):
I'm sure this can be used many places in the code to simplify logic; a good addition!

PR #306, opened 2022-03-22
Add Symmetry plot method

### LINKED ISSUE #171: Plotting of symmetry elements
I think we should have a `Symmetry.plot()` method to plot the symmetry elements in the stereographic projection, as I've done in the [stereographic plot notebook](https://orix.readthedocs.io/en/latest/stereographic_projection.html#Experimental-functionality).

Would also be nice to have stereograms with filled and non-filled markers etc., but that is a bit more work, I think.

harripj:
As discussed in #303 and #171, it would be a good idea to add a `plot` method to `Symmetry` to display symmetry elements in the crystal reference frame under the stereographic projection.

Closes #171.

#### Progress of the PR

```python
>>> from orix.quaternion import symmetry
>>> symmetry.O.plot()
```
image

#### For reviewers


orix/quaternion/symmetry.py
```diff
@@ -451,6 +451,77 @@ def fundamental_zone(self):
...
+        dict : optional
+            Keyword arguments passed to
+            :func:`~orix.plot.StereographicPlot.scatter`, which passes
+            these on to :meth:`matplotlib.axes.Axes.scatter`.
+
+        Returns
+        -------
+        fig : matplotlib.figure.Figure
+            The created figure, returned if `return_figure` is True.
+
```
hakonanes, line 485: Remove this blank line to maintain the style.

hakonanes, line 485: Remove this blank line to maintain the style.

orix/quaternion/symmetry.py
```diff
@@ -451,6 +451,77 @@ def fundamental_zone(self):
...
+            `matplotlib.Axes.scatter`.
+        dict : optional
+            Keyword arguments passed to
+            :func:`~orix.plot.StereographicPlot.scatter`, which passes
+            these on to :meth:`matplotlib.axes.Axes.scatter`.
+
+        Returns
+        -------
+        fig : matplotlib.figure.Figure
+            The created figure, returned if `return_figure` is True.
```
hakonanes, line 484: I suggest to clarify that `return_figure` has to be passed as a keyword argument.

hakonanes, line 484: I suggest to clarify that `return_figure` has to be passed as a keyword argument.

orix/quaternion/symmetry.py
```diff
@@ -451,6 +451,77 @@ def fundamental_zone(self):
...
+
+        """
+
+        if isinstance(orientation, str) and orientation.lower() == "default":
```
hakonanes, line 488: This will then be `if orientation is None:`

hakonanes, line 488: This will then be `if orientation is None:`

orix/quaternion/symmetry.py
```diff
@@ -451,6 +451,77 @@ def fundamental_zone(self):
...
+            # orientation chosen to mimic stereographic projections as
+            # shown: http://xrayweb.chem.ou.edu/notes/symmetry.html
+            orientation = Rotation.from_axes_angles((-1, 8, 1), np.deg2rad(65))
+        if not isinstance(orientation, Rotation):
+            raise TypeError("Orientation must be a Rotation instance or 'default'.")
```
hakonanes, line 493: Must be updated as well.

hakonanes, line 493: Must be updated as well.

orix/quaternion/symmetry.py
```diff
@@ -451,6 +451,77 @@ def fundamental_zone(self):
...
+        # reproject vectors on lower hemisphere to upper hemisphere
+        v_lower.z *= -1
+
+        kwargs.setdefault("return_figure", False)
+        return_figure = kwargs.pop("return_figure")
+
+        kwargs.setdefault("s", 100)
+        kwargs.setdefault("ec", "tab:blue")
+
+        figure = v_upper.scatter(
```
hakonanes, line 509: This is just an idea for the future, but it might be that we should support a boolean parameter to `Vector3d.scatter()` to set whether vectors on the other hemisphere (northern if `hemisphere="south` is passed and vice versa) should be projected up to the projection plane as well. We should then force the scatter points to be different.

hakonanes, line 509: This is just an idea for the future, but it might be that we should support a boolean parameter to `Vector3d.scatter()` to set whether vectors on the other hemisphere (northern if `hemisphere="south` is passed and vice versa) should be projected up to the projection plane as well. We should then force the scatter points to be different.

orix/tests/quaternion/test_symmetry.py
```diff
@@ -19,6 +19,8 @@
 from copy import deepcopy
 
 from diffpy.structure.spacegroups import GetSpaceGroup
+from matplotlib.figure import Figure
```
hakonanes, line 22: `Figure` can be accessed from `matplotlib.pyplot.Figure`, so you can skip this import.

hakonanes, line 22: `Figure` can be accessed from `matplotlib.pyplot.Figure`, so you can skip this import.

orix/tests/quaternion/test_symmetry.py
```diff
@@ -406,6 +408,39 @@ def test_hash_persistence():
     assert all(h1a == h2a for h1a, h2a in zip(h1, h2))
 
 
+@pytest.mark.parametrize("symmetry", [C1, C4, Oh])
+def test_symmetry_plot(symmetry):
+    figure = symmetry.plot(return_figure=True)
+    assert isinstance(figure, Figure)
+    assert len(figure.axes) == 1
+    ax = figure.axes[0]
+    num = 2 if symmetry.improper.any() else 1
```
hakonanes, line 417: We have `Symmetry.is_proper` for this check.

hakonanes, line 417: We have `Symmetry.is_proper` for this check.

orix/tests/quaternion/test_symmetry.py
```diff
@@ -406,6 +408,39 @@ def test_hash_persistence():
...
+    assert len(ax.collections) == num
+    c0 = ax.collections[0]
+    assert len(c0.get_sizes()) == np.count_nonzero(~symmetry.improper)
+    assert c0.get_label().lower() == "upper"
+    assert np.allclose(c0.get_sizes(), 100)  # default value
```
hakonanes, line 422: Is this check necessary? What is tested here, that it is a circle? If it's not very important, I suggest to drop these checks.

hakonanes, line 422: Is this check necessary? What is tested here, that it is a circle? If it's not very important, I suggest to drop these checks.

orix/tests/quaternion/test_symmetry.py
```diff
@@ -406,6 +408,39 @@ def test_hash_persistence():
...
+    assert len(c0.get_paths()[0].vertices) == 4  # for crosses
+    if num > 1:
+        c1 = ax.collections[1]
+        assert len(c1.get_sizes()) == np.count_nonzero(symmetry.improper)
+        assert c1.get_label().lower() == "lower"
+        assert np.allclose(c1.get_sizes(), 100)  # default value
+        # in my experience there are 26 points in circle marker path
```
hakonanes, line 429: Remove first person phrasing here, but leave the comment if these checks are kept. I've had to change many tests in orix and elsewhere with hardcoded numbers "deep" within Matplotlib that change between releases.

hakonanes, line 429: Remove first person phrasing here, but leave the comment if these checks are kept. I've had to change many tests in orix and elsewhere with hardcoded numbers "deep" within Matplotlib that change between releases.

doc/point_groups.ipynb
```diff
@@ -0,0 +1,112 @@
+{
+    "cells": [
+        {
+            "cell_type": "markdown",
+            "metadata": {},
```
hakonanes, line 5: Add "nbsphinx": "hidden" to the metadata, so that this notice isn't shown when built on the Read the Docs web page (because there links actually work).

hakonanes, line 5: Add "nbsphinx": "hidden" to the metadata, so that this notice isn't shown when built on the Read the Docs web page (because there links actually work).

doc/point_groups.ipynb
```diff
@@ -0,0 +1,112 @@
...
+        {
+            "cell_type": "code",
+            "execution_count": null,
+            "metadata": {},
+            "outputs": [],
+            "source": [
+                "from matplotlib import pyplot as plt\n",
+                "import numpy as np\n",
+                "\n",
+                "import orix.plot.stereographic_plot\n",
```
hakonanes, line 59: The projection is registered with `from orix import plot`, I suggest to use this command instead since its simpler.

hakonanes, line 59: The projection is registered with `from orix import plot`, I suggest to use this command instead since its simpler.

doc/point_groups.ipynb
```diff
@@ -0,0 +1,112 @@
...
+        },
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
+                "## Visualizing point groups\n",
+                "\n",
+                "Point group symmetry operations are shown here under the stereographic projection.\n",
+                "\n",
+                "Vectors located on the upper (`z > 0`) hemisphere are displayed as crosses (`+`), whereas vectors on the lower hemisphere are reprojected onto the upper hemisphere and shown as open circles (`o`)."
```
hakonanes, line 18: I suggest to link to for example http://xrayweb.chem.ou.edu/notes/symmetry.html#point to give the reader an opportunity to understand the visualizations. Either that, or we add some theory here ourselves. I think adding the link is OK for now.

hakonanes, line 18: I suggest to link to for example http://xrayweb.chem.ou.edu/notes/symmetry.html#point to give the reader an opportunity to understand the visualizations. Either that, or we add some theory here ourselves. I think adding the link is OK for now.

hakonanes (changes requested):
Very helpful and nice visualization of symmetry operations! All requests and comments are related to documentation and tests.

Should add `plot` to the list of `Symmetry` methods below `from_generators` in the [reference](https://github.com/pyxem/orix/blob/master/doc/reference.rst#symmetry).

hakonanes:
Seems like GitHub has submitted all my comments twice. I suggest to leave the extra comments, in case both are deleted when one is deleted...

orix/tests/quaternion/test_symmetry.py
harripj: Thanks, didn't know this.

orix/quaternion/symmetry.py
harripj: This seems like a good idea. We could move the code here to `Vector3d.scatter` and the keyword argument could be called something like `reproject_upper`. This would then simplify the method here too.

harripj:
Doc build error looks like it is part of jinja2 or nbconvert and is being experienced elsewhere recently: https://github.com/sphinx-doc/sphinx/issues/10289

harripj:
> Doc build error looks like it is part of jinja2 or nbconvert and is being experienced elsewhere recently: [sphinx-doc/sphinx#10289](https://github.com/sphinx-doc/sphinx/issues/10289)

Looks like a fairly minor fix. I suggest we wait until `nbconvert` pushes the patch which will hopefully sort everything from our side without change, An alternative is to pin `jinja2` to `3.0` in requirements, but it seems that like this will be cleared up pretty quickly.

doc/point_groups.ipynb
```diff
@@ -0,0 +1,116 @@
...
+            "cell_type": "code",
+            "execution_count": null,
+            "metadata": {
+                "tags": [
+                    "nbsphinx-thumbnail"
+                ]
+            },
+            "outputs": [],
+            "source": [
+                "from orix.quaternion import symmetry\n",
```
hakonanes, line 42: All notebooks must use the "inline" backend, so add `%matplotlib inline` here (don't know how to set this in `doc/conf.py`, checked nbsphinx docs now).

hakonanes:
> I suggest we wait until `nbconvert` pushes the patch which will hopefully sort everything from our side without change

I agree. Checked the notebook locally, just had one important comment.

hakonanes:
After the notebook change, I have a final suggestion: could you update the locations of other relevant packages references in `conf.py`

from

```python
    "h5py": ("http://docs.h5py.org/en/stable/", None),
    "matplotlib": ("https://matplotlib.org", None),
    "numpy": ("https://docs.scipy.org/doc/numpy", None),
    "scipy": ("https://docs.scipy.org/doc/scipy/reference", None),
```

to

```python
    "h5py": ("https://docs.h5py.org/en/stable", None),
    "matplotlib": ("https://matplotlib.org/stable", None),
    "numpy": ("https://numpy.org/doc/stable", None),
    "scipy": ("https://docs.scipy.org/doc/scipy", None),
```

Sphinx tells us nicely that they have changed when running `make html`.

harripj:
> > Doc build error looks like it is part of jinja2 or nbconvert and is being experienced elsewhere recently: [sphinx-doc/sphinx#10289](https://github.com/sphinx-doc/sphinx/issues/10289)
> 
> Looks like a fairly minor fix. I suggest we wait until `nbconvert` pushes the patch which will hopefully sort everything from our side without change, An alternative is to pin `jinja2` to `3.0` in requirements, but it seems that like this will be cleared up pretty quickly.

NB this is the PR in question https://github.com/jupyter/nbconvert/pull/1737.

harripj:
> NB this is the PR in question https://github.com/jupyter/nbconvert/pull/1737.

This appears to have been fixed.

hakonanes:
This PR can now use reprojection functionality of vectors on the other hemisphere!

doc/point_groups.ipynb
```diff
@@ -0,0 +1,126 @@
...
+                "for i, s in enumerate(schoenflies):\n",
+                "    sym = getattr(symmetry, s)\n",
+                "\n",
+                "    ori_sym = sym.outer(orientation)\n",
+                "    v = (ori_sym * Vector3d.zvector())\n",
+                "    \n",
+                "    # reflection in the projection plane (x-y) is performed internally in\n",
+                "    # Symmetry.plot() or when using the `reproject=True` argument for\n",
+                "    # Vector3d.scatter()\n",
+                "    v_reproject = Vector3d(v.data * (1, 1, -1))\n",
```
hakonanes, line 91: This can be done as suggested in https://github.com/pyxem/orix/pull/308#issuecomment-1081939951.

hakonanes (approved):
This is a very good way to visualize symmetry operations; I'm sure I'll use it many times!

I've suggested an insignificant change to the user guide notebook. Performing the change isn't required to merge, but I think it shows the user a slightly simpler way to reflect a vector about the projection plane.

doc/point_groups.ipynb
harripj:
```python
v_reproject = Vector3d(v.data)
v_reproject += -1
```

hakonanes: Yes, that looks cleanest. And then negate only the Z coordinate!

harripj: Doesn't work actually, the memory is shared.

harripj:
```python
v_reproject = Vector3d(v.data.copy())
v_reproject += -1
```
This is okay if that works for you?

hakonanes: Yes, but it's only the z coordinate that should be negated?

pc494 (approved):
This looks good, assuming the tests green light. I'll let @hakonanes merge.

PR #313, opened 2022-04-01
Remove functions deprecated for removal in 0.9

harripj:
Functions and properties previously deprecated and tagged for removal in release `0.9` have been removed. The changelog has been updated.

Question, `Misorientation.distance()` was not flagged for deprecation (whereas `Orientation.distance()` was. I believe we want to remove both here?

#### Progress of the PR

#### For reviewers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -54,7 +54,7 @@ Removed
 -------
 - `orix.scalar.Scalar` class has been removed and the data held by `Scalar` is now
   returned directly as a `numpy.ndarray`.
-- Functions: `(Mis)Orientation.set_symmetry` and `Orientation.distance`  and property:
+- Functions: `(Mis)Orientation.set_symmetry` and `Orientation.distance` and property:
  `Object3d.data_dim` were previously deprecated and have now been removed. 
 
 2022-02-21 - version 0.8.2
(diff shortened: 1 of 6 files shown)


hakonanes:
Just ping whenever this is ready for review!

harripj:
Just a thought that even though `Orientation.distance()` was tagged for removal, if we keep `Misorientation.distance()` then `Orientation` will keep a `distance()` method be inheritance. So I think it is best to either keep or remove both.

hakonanes:
> So I think it is best to either keep or remove both.

Good point. Removing `Misorientation.distance()` removes functionality, so I suggest to keep both but leave the deprecation warning. We should leave that method alone until an alternative is implemented.

That the deprecation warning says that the method will be removed in v0.8 isn't important I think, what's important is that users know to use `Orientation.get_distance_matrix()`.

harripj:
> Good point. Removing Misorientation.distance() removes functionality, so I suggest to keep both but leave the deprecation warning. We should leave that method alone until an alternative is implemented.

Good idea, I have left the `distance()` functions in. This is ready for another pair of eyes!

hakonanes (approved):
Thanks for making this PR, looks good to me (inspected branch locally and the docs build log [no warnings or errors])! As I've said before, it's always nice to decrease the code base without loosing functionality or clarity.

pc494 (approved):
LGTM, merging

Issue #315, opened 2022-04-05
Memory efficient alternative to Misorientation.distance()


hakonanes:
Computing the outer disorientation angle (geodesic distance) with itself is implemented in a memory efficient way with Dask in `Orientation.get_distance_matrix()`. This replaced `Orientation.distance()`. There does not exist a similar alternative for `Misorientation.distance()`. Until this is done, the deprecation warning for `Orientation.distance()` is left in.

harripj:
Can this be done using `Quaternion._outer_dask()`?

```python
>>> from orix.quaternion import symmetry, Orientation, Misorientation, Rotation

>>> o1 = Orientation.random((5,))
>>> o2 = Orientation.random((5,))
>>> o1.symmetry = symmetry.T
>>> o2.symmetry = symmetry.T

>>> # existing method
>>> m = o1 - o2
>>> dist = m.distance()

>>> # explicit computation, warning memory for large-size symmetries
>>> s1, s2 = m.symmetry
>>> mm = s1.outer(m).outer(s2).outer(s1).outer(~m).outer(s2)
>>> dist2 = mm.angle.min(axis=(0, 2, 3, 5))
>>> np.allclose(dist, dist2)
True

>>> # reduced computation where s1 and s2 are the same
>>> assert s1 == s2
>>>mm2 = s1.outer(m).outer(s1).outer(~m)
>>> dist3 = mm2.angle.min(axis=(0, 2))
>>>np.allclose(dist, dist3)
True

>>> # using `Quaternion._outer_dask` with the two symmetry expanded misorientations
>>> q = Rotation(s1.outer(m).outer(s2))._outer_dask(s1.outer(~m).outer(s2)).compute()
>>> dist4 = Rotation(q).angle.min(axis=(0, 2, 3, 5))
>>> np.allclose(dist, dist4)
True
```

hakonanes:
Good idea, I'll have a look at how it performs with the outer products in the misorientation clustering notebook.

hakonanes:
The computation could be done within 1min 19s on my machine, with minimal memory requirements, since there are only 860 misorientations considered. I'll set up a PR to add `Misorientation.get_distance_matrix()`.

hakonanes:
> The computation could be done within 1min 19s

[...]

PR #317, opened 2022-04-06
Add memory efficient alternative to Misorientation.distance() using Dask

hakonanes:
This PR is a direct result of @harripj's comment https://github.com/pyxem/orix/issues/315#issuecomment-1089348212. It adds `Misorientation.get_distance_matrix()` which computes the disorientation angle of the rotation between two misorientations.

I've only made a lazy implementation using Dask; no direct NumPy equivalent exists, like for `Orientation.get_distance_matrix()`.

Note that the above method is the only new method implemented, private or public. Parts of the method can perhaps be generalized to other methods, like was done with `Orientation.angle_with_outer()` from `Orientation.get_distance_matrix()`, but I haven't focussed on that here.

The misorientation clustering notebook now uses this method: the distance matrix is computed within 18s on my machine without any memory overhead. The results are the same as in the paper. **EDIT: The notebook is now executed on every commit to `master` and PR by Read the Docs, made possible by downloading the Ti orientations via the data module.**

I couldn't find any tests for `Misorientation.distance()` specifically, only `Orientation.distance()`. I took the five first misorientations from the clustering notebook and the resulting distance matrix as a validation in a test.

After this method is added, we can deprecate `Misorientation.distance()` for 0.9 (in another PR), and finally remove it completely for 1.0.

**EDIT: Code cells in the documentation now have a copy button. This requires a new documentation dependency in `sphinx-copybutton`.**

#### Progress of the PR
- [x] Remove cell output from notebook once #318 is merged, since the Ti orientations can be downloaded by Read the Docs

```python
>>> import numpy as np
>>> from orix.quaternion import Misorientation, symmetry
>>> m = Misorientation.from_axes_angles((1, 0, 0), [0, np.pi / 2])
>>> m.symmetry = (symmetry.D6, symmetry.D6)
>>> d = m.get_distance_matrix()
>>> d
[[0.         1.57079633]
 [1.57079633 0.        ]]
```

#### For reviewers


hakonanes:
I anticipate to make some more PRs in the near future. I'd like to respect your time, Phillip and Patrick, so I've added a new label "status: needs review" which I'll use to mark PRs when they're ready, instead of pinging.

harripj:
> I anticipate to make some more PRs in the near future. I'd like to respect your time, Phillip and Patrick, so I've added a new label "status: needs review" which I'll use to mark PRs when they're ready, instead of pinging.

Looking forward to seeing it! I will do my best to review when I can.

harripj (commented):
I think that a Dask implementation of the distance method is great and the right way to go forward. But before we go ahead I have a few things I think we should discuss.

I have written a few ways to compute the mismisorientation below and compared them. When `symmetry1` and `symmetry2` are the same:
```python
>>> from orix.quaternion import Orientation, Misorientation, symmetry
>>> import numpy as np

>>> o1 = Orientation.random((5,))
>>> o2 = Orientation.random((5,))

>>> # same symmetry
>>> o1.symmetry = symmetry.D6
>>> o2.symmetry = symmetry.D6

>>> m = o1 - o2

>>> dist1 = m.distance()

>>> s1, s2 = m.symmetry
>>> p1 = s1.outer(m).outer(s2)
>>> p2 = s1.outer(~m).outer(s2)
>>> dist2 = p1.outer(p2).angle.min(axis=(0, 2, 3, 5))
>>> np.allclose(dist1, dist2)
True

>>> dist3 = p1.outer(~m).angle.min(axis=(0, 2))
>>> np.allclose(dist1, dist3)
True

>>> dist4 = m.get_distance_matrix()
>>> np.allclose(dist1, dist4)
True
```

But for differing symmetries:
```python
>>> o1 = Orientation.random((5,))
>>> o2 = Orientation.random((5,))

>>> # different unrelated symmetry
>>> o1.symmetry = symmetry.D6
>>> o2.symmetry = symmetry.T

>>> m = o1 - o2

>>> dist1 = m.distance()

>>> s1, s2 = m.symmetry
>>> p1 = s1.outer(m).outer(s2)
>>> p2 = s1.outer(~m).outer(s2)
>>> dist2 = p1.outer(p2).angle.min(axis=(0, 2, 3, 5))
>>> np.allclose(dist1, dist2)
False

>>> (dist2 <= dist1).all()
True

>>> dist3 = p1.outer(~m).angle.min(axis=(0, 2))
>>> np.allclose(dist1, dist3)
True

>>> dist4 = m.get_distance_matrix()
>>> np.allclose(dist1, dist4)
False
```

So we need to work this out. The formation of this calculation, from the 2020 Johnstone paper: `d = s1.outer(~m).outer(s2).outer(s1).outer(m).outer(s2)`, which I believe is using right-hand multiplication, so here (left-hand multiplication) as: `d = s1.outer(m).outer(s2).outer(s1).outer(~m).outer(s2)`, is what I believe we should treat as the truth, and any optimisations should work around this. What is confusing is that this disagrees with `distance()`.

Although when just looking at angles the returned by this calculation the order doesn't matter:
```python
>>> p12 = p1.outer(p2)
>>> p21 = p2.outer(p1)
>>> np.allclose(p12.data, p21.data)
False

>>> np.allclose(p12.angle.min(axis=(0, 2, 3, 5)), p21.angle.min(axis=(0, 2, 3, 5)).T)
True
```

It think the best approach would be that we have a known truth to test against.

harripj:
It looks like in `distance()` the calculation being performed is:
```python
>>> s1_outer_s1 = s1.outer(s1)
>>> m5 = s2.outer(~m).outer(s1_outer_s1).outer(m).outer(s2)
>>> dist5 = m5.angle.min(axis=(0, 2, 3, 5))
>>> np.allclose(dist5, dist1)
True
```
For identical symmetries of course this doesn't matter, but for differing symmetries this creates a difference. I don't think there is a need to swap the symmetry order of `m`, as is currently implemented in `distance`? The Krakow paper mentions that the symmetry order matters in Appendix B. What do you both think?

harripj:
> ```matlab
>     0    0.0111    0.2414    0.2573    0.0279
>     0.0111         0    0.2198    0.2355    0.0207
>     0.2414    0.2198         0         0    0.2345
>     0.2573    0.2355         0         0    0.2496
>     0.0279    0.0207    0.2345    0.2496    0.0238
> ```

In orix running this as:
```python
>>> from orix.quaternion import Orientation, Misorientation, symmetry
>>> import numpy as np

>>> q = np.array([[-0.8541, -0.5201, -0.0053, -0.0002],
    [-0.8486, -0.5291, -0.0019, -0.0018],
    [-0.7851, -0.6194, -0.0043, -0.0004],
    [-0.7802, -0.3136, -0.5413, -0.0029],
    [-0.8518, -0.5237, -0.0004, -0.0102]])

>>> q = Misorientation(q)
>>> q.symmetry = symmetry.D6, symmetry.D6

>>> s1, s2 = q.symmetry
>>> a = s1.outer(~q).outer(s2).outer(s1).outer(q).outer(s2)
>>> a.angle.min(axis=(0, 2, 3, 5))
array([[0.        , 0.02239326, 0.2419903 , 0.25797761, 0.02385484],
       [0.02239326, 0.        , 0.22095847, 0.23673986, 0.02118541],
       [0.2419903 , 0.22095847, 0.        , 0.01837405, 0.234388  ],
       [0.25797761, 0.23673986, 0.01837405, 0.        , 0.24965008],
       [0.02385484, 0.02118541, 0.234388  , 0.24965008, 0.        ]])
```
A lot of the values are close. I am confused why MTEX doesn't give all `0` on the diagonal though.

harripj:
Does MTEX define symmetries in the same way as orix?

hakonanes:
> A lot of the values are close.

These are the values in the `D` matrix on Google Drive for these misorientations, and are the hard-coded angles in the mentioned test in this PR.

> Does MTEX define symmetries in the same way as orix?

Not 100% sure, they are [defined dynamically](https://github.com/mtex-toolbox/mtex/blob/develop/geometry/%40symmetry/symmetry.m#L159) after setting a certain crystal axis alignment (e.g. e1 || a, e3 || c*), by rotations about direct lattice vectors.

hakonanes:
The reason for the force-push is that I rebased this branch on `master` after #318. I've updated the top comment with the new additions.

harripj:
> Not 100% sure, they are [defined dynamically](https://github.com/mtex-toolbox/mtex/blob/develop/geometry/%40symmetry/symmetry.m#L159) after setting a certain crystal axis alignment (e.g. e1 || a, e3 || c*), by rotations about direct lattice vectors.

Let's continue this discussion in #319.

>  I am confused why MTEX doesn't give all 0 on the diagonal though.

I would like to address this before going further. Is there any way that the angle between an orientation and itself can be anything other than 0, as MTEX suggests?

For an implementation of `get_distance_matrix()`, which is a `Misorientation` with itself, the symmetries would be the same, so I think what is implemented here is okay. Perhaps what is discussed here is out of the scope for this issue https://github.com/pyxem/orix/pull/317#issuecomment-1091553965 and should be addressed elsewhere?

hakonanes:
> Perhaps what is discussed here is out of the scope for this issue https://github.com/pyxem/orix/pull/317#issuecomment-1091553965 and should be addressed elsewhere?

Yes, I think a new issue on implementations of `Misorientation` methods `dot()`, `dot_outer()`, `angle_with()` and `angle_with_outer()` respecting symmetry would be the more suited.

> Is there any way that the angle between an orientation and itself can be anything other than 0, as MTEX suggests?

I don't understand it either, and I think it must be a rounding issue.

harripj:
@hakonanes I would like to add a couple tests to this PR if that is okay? I think we should test that for all symmetries the implementation of `distance()` and `get_distance_matrix()` produces the same result (this can be removed when `distance()` is removed, but it is then in the commit history). Also a test that can be kept, explicitly computing the mismisorientation (eq 6 in Johnstone et al. orix paper) and comparing with the implementation of `get_distance_matrix()`.

hakonanes:
Sounds very good, go ahead!

harripj:
Happy with this PR 👍, but as I have contributed I will ask @pc494 to have a look when he can.

hakonanes:
Could you have a look at this when you find time, @pc494?

pc494:
Reminding myself to look at this tomorrow :)

pc494 (approved):
This looks good, merging.

Issue #149, opened 2021-01-22
Improving the orix demos and documentation


hakonanes:
I would like for orix to adopt the demos-and-documentation-setup we now have in kikuchipy from v0.3.0:
* Jupyter notebook demos as [part of the package documentation](https://github.com/pyxem/kikuchipy/tree/master/doc)
* Uses [nbsphinx](https://nbsphinx.readthedocs.io/en/0.8.1/) (called indirectly by Sphinx when the docs are built via `make html`) to convert the notebooks to reStructuredText files included in the [Sphinx documentation](https://kikuchipy.org/en/stable/)
* Users can run the notebooks interactively [via Binder in their browsers](https://mybinder.org/v2/gh/pyxem/kikuchipy/v0.3.1?filepath=doc)
* Documentation can be written solely in notebooks
* Whether the notebooks run or fail is determined by Read the Docs on every PR check
* No need for a separate demo repository

Necessary changes for orix to adopt this setup:
- [x] Some test data as part of a `orix.data` module for the notebooks to use. A few small datasets (less than 0.5 MB, e.g.) should be part of the package. Larger data can be downloaded from an external source (similar to [kikuchipy-data](https://github.com/pyxem/kikuchipy-data), with the help of the `pooch` package (used by e.g. `scikit-image`), into a cache on the user's machine *once*, and then loaded directly from the `orix.data` via a safe registry. This introduces a pooch dependency.
- [x] nbsphinx dependency (#152)
- [x] Some notebooks in the orix/doc directory (like the crystal map notebook) (#152)

If there is interest, I can set this up.

pc494:
This seems like a good idea. No pressure/rush

hakonanes:
Since I've done it once already it shouldn't be too difficult to add the module with a crystal map and port the notebook over here. The challenge is, if we want to archive the orix-demos, to make the other more heavy-duty notebooks over.

hakonanes:
I've bumped a `orix.data` module to v0.7.

hakonanes:
A data module was added in #318, and will be released as part of 0.9.

PR #318, opened 2022-04-06
Add orix data module

hakonanes:
Add `orix.data` module as suggested in #149. Data sets, like instances of `CrystalMap`, `Orientation` etc., used in tests and/or user guide notebooks can be conveniently (down)loaded from the module. Whenever a data set is desired, e.g. by calling `orix.data.sdss_austenite()`, a registry is checked for the file location in the local cache (by default ~/.cache/orix, but can be set via a global variable `ORIX_DATA_DIR`) and loaded if it is there. If it's not in the cache, it is downloaded from an HTTPS address to the local cache, provided that `allow_download=True` is passed to the function. File hashes are hard-coded in the registry, preventing harmful files being downloaded, and also making sure that a file is re-downloaded whenever we change the hash in the registry and online. The implemented approach is used in e.g. [kikuchipy.data](https://kikuchipy.org/en/stable/reference.html#module-kikuchipy.data) and [skimage.data](https://scikit-image.org/docs/stable/api/skimage.data.html#module-skimage.data).

This requires a new dependency, [`pooch`](https://github.com/fatiando/pooch): it has a BSD3 license and is [lightweight](https://github.com/fatiando/pooch/blob/7ddc04505de237c7bc8b50c01ad7bc05d440646a/setup.cfg#L37-L46). If people really want to, we can make this dependency optional via a `orix[data]` extra requirement set in setup.py.

So far, only two data sets, from EBSD of super duplex stainless steel (SDSS), are added to the module: `orix.data.sdss_austenite()` (orix HDF5 file) and `orix.data.sdss_ferrite_austenite()` (.ang file). They are downloaded upon request from [here](https://folk.ntnu.no/hakonwii/files/orix-demos/). The latter data set is used in the crystal map user guide. It was previously downloaded with urllib rather inconveniently, but is now downloaded with one call to `sdss_ferrite_austenite(allow_download=True)` (see docs built from this PR).

Google provides a new file hash for "Ti_orientations.ctf" every time it is downloaded from Google Drive, preventing it from being downloaded from `orix.data`. We should add this file to an `orix-data` repo, similar to https://github.com/pyxem/kikuchipy-data. All code lines are added but commented out for a `orix.data.ti_orientations()` function.

#### Progress of the PR

```python
>>> from orix import data

# Create new local data directory and download file here
>>> xmap = data.sdss_ferrite_austenite(allow_download=True)
Downloading file 'data/sdss/sdss_ferrite_austenite.ang' from 'https://folk.ntnu.no/hakonwii/files/orix-demos/sdss_ferrite_austenite.ang' to '/home/hakon/.cache/orix/master'.
100%|████████████████████████████████████████| 652k/652k [00:00<00:00, 967MB/s]
>>> xmap
Phase   Orientations       Name  Space group  Point group  Proper point group       Color
    1   5657 (48.4%)  austenite         None          432                 432    tab:blue
    2   6043 (51.6%)    ferrite         None          432                 432  tab:orange
Properties: iq, dp
Scan unit: um

>>> xmap_di = data.sdss_austenite(allow_download=True)
Downloading file 'data/sdss/sdss_austenite_dp.h5' from 'https://folk.ntnu.no/hakonwii/files/orix-demos/sdss_austenite_dp.h5' to '/home/hakon/.cache/orix/master'.
100%|█████████████████████████████████████| 10.8M/10.8M [00:00<00:00, 13.0GB/s]
>>> xmap_di
Phase    Orientations       Name  Space group  Point group  Proper point group     Color
    0  11700 (100.0%)  austenite         None         m-3m                 432  tab:blue
Properties: AvDotProductMap, CI, IQ, ISM, KAM, OSM, RefinedDotProducts, TopDotProductList, TopMatchIndices
Scan unit: um

# Reload data from cache, also passing kwargs on to orix.io.plugins.orix_hdf5.file_reader()

>>> xmap_ref = data.sdss_austenite(refined=True)
>>> xmap_ref
Phase    Orientations       Name  Space group  Point group  Proper point group     Color
    0  11700 (100.0%)  austenite         None         m-3m                 432  tab:blue
Properties: AvDotProductMap, CI, IQ, ISM, KAM, OSM, RefinedDotProducts, TopDotProductList, TopMatchIndices
Scan unit: um
>>> xmap_di.rotations_per_point
50
>>> xmap_ref.rotations_per_point
1
```

#### For reviewers

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -206,23 +206,42 @@ framework. The tests reside in a ``test`` directory within each module. Tests ar
 methods that call functions in orix and compare resulting output values with known
 answers. Install necessary dependencies to run the tests::
 
-   $ pip install --editable .[tests]
+   pip install --editable .[tests]
 
 Some useful `fixtures <https://docs.pytest.org/en/latest/fixture.html>`_ are available
 in the ``conftest.py`` file.
 
 To run the tests::
 
-   $ pytest --cov --pyargs orix
+   pytest --cov --pyargs orix
 
 The ``--cov`` flag makes `coverage.py <https://coverage.readthedocs.io/en/latest/>`_
 print a nice report in the terminal. For an even nicer presentation, you can use
 ``coverage.py`` directly::
 
-   $ coverage html
+   coverage html
 
 Then, you can open the created ``htmlcov/index.html`` in the browser and inspect the
 coverage in more detail.
+
+Adding data to the data module
+==============================
+
+Test data for user guides and tests are included in the :mod:`orix.data` module via the
+`pooch <https://www.fatiando.org/pooch/latest>`_ Python library. These are listed in a
+file registry (`orix.data._registry.py`) with their file verification string (hash,
+SHA256, obtained with e.g. `sha256sum <file>`) and location, the latter potentially not
+within the package but from the `orix-data <https://github.com/pyxem/orix-data>`_
+repository, since some files are considered too large to include in the package.
+
+If a required dataset isn't in the package, but is in the registry, it can be downloaded
+from the repository when the user passes `allow_download=True` to e.g.
+`sdss_austenite()`. The dataset is then downloaded to a local cache, e.g.
+`/home/user/.cache/orix/`. The data cache directory can be set with a global
+`ORIX_DATA_DIR` variable locally, e.g. by setting export `ORIX_DATA_DIR=~/orix_data` in
+`~/.bashrc`. Pooch handles downloading, caching, version control, file verification
+(against hash) etc. If we have updated the file hash, pooch will re-download it. If the
+file is available in the cache, it can be loaded as the other files in the data module.
 
 Continuous integration (CI)
 ===========================
--- a/orix/data/__init__.py
+++ b/orix/data/__init__.py
@@ -1,0 +1,156 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018-2022 the orix developers
+#
+# This file is part of orix.
+#
+# orix is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# orix is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with orix.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test data.
+
+Some datasets must be downloaded from the web. For more test datasets,
+see :doc:`open datasets <open_datasets>`.
+"""
+
+import os
+from pathlib import Path
+
+# import numpy as np
+import pooch as ppooch
+
+from orix import __version__, io
+from orix.data._registry import registry, registry_urls
+
+# from orix.quaternion import Orientation, symmetry
+
+
+__all__ = [
+    "sdss_austenite",
+    "sdss_ferrite_austenite",
+    #    "ti_orientations",
+]
+
+
+fetcher = ppooch.create(
+    path=ppooch.os_cache("orix"),
+    base_url="",
+    version=__version__.replace(".dev", "+"),
+    env="ORIX_DATA_DIR",
+    registry=registry,
+    urls=registry_urls,
+)
+cache_data_path = fetcher.path.joinpath("data")
+package_data_path = Path(os.path.abspath(os.path.dirname(__file__)))
+
+
+def _has_hash(path, expected_hash):
+    """Check if the provided path has the expected hash."""
+    if not os.path.exists(path):
+        return False
+    else:
+        return ppooch.file_hash(path) == expected_hash  # pragma: no cover
+
+
+def _cautious_downloader(url, output_file, pooch):
+    if pooch.allow_download:
+        delattr(pooch, "allow_download")
+        # HTTPDownloader() requires tqdm
+        download = ppooch.HTTPDownloader(progressbar=True)
+        download(url, output_file, pooch)
+    else:
+        raise ValueError(
+            "The dataset must be (re)downloaded from the orix-data repository on GitHub"
+            " (https://github.com/pyxem/orix-data) to your local cache with the pooch "
+            "Python package. Pass `allow_download=True` to allow this download."
+        )
+
+
+def _fetch(filename, allow_download):
+    resolved_path = os.path.join(package_data_path, "..", filename)
+    expected_hash = registry[filename]
+    if _has_hash(resolved_path, expected_hash):  # File already in data module
+        return resolved_path  # pragma: no cover
+    else:  # Pooch must download the data to the local cache
+        fetcher.allow_download = allow_download  # Extremely ugly
+        resolved_path = fetcher.fetch(filename, downloader=_cautious_downloader)
+    return resolved_path
+
+
+def sdss_austenite(allow_download=False, **kwargs):
+    """Single phase Austenite crystallographic map of shape (100, 117)
+    produced from dictionary indexing of electron backscatter
+    diffraction patterns of a super duplex stainless steel (SDSS) sample
+    containing both Austenite and Ferrite grains.
+
+    Parameters
+    ----------
+    allow_download : bool, optional
+        Whether to allow downloading the dataset from the internet to
+        the local cache with the pooch Python package. Default is False.
+    kwargs
+        Keyword arguments passed to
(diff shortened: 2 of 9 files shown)


CONTRIBUTING.rst
```diff
@@ -206,24 +206,49 @@ framework. The tests reside in a ``test`` directory within each module. Tests ar
...
+   pip install --editable .[tests]
 
 Some useful `fixtures <https://docs.pytest.org/en/latest/fixture.html>`_ are available
 in the ``conftest.py`` file.
 
+.. note::
+
+    Some :mod:`orix.data` module tests check that data not part of the package
+    distribution can be downloaded from the web, thus downloading some datasets of total
+    size ~11 MB to your local cache (by default to `/home/<user>/.cache/orix`).
```
harripj, line 218: On Mac this was `/Users/<user>/Library/Caches/orix/master` for me.

harripj (commented):
Looks good to me. Is there a way to remove the cache?

> Google provides a new file hash for "Ti_orientations.ctf" every time it is downloaded from Google Drive, preventing it from being downloaded from orix.data. 

We should wait until this is sorted. Who owns the file?

CONTRIBUTING.rst
hakonanes: Right, ttating the default on Linux is too specific in this case, and can confuse the user. It's better to specify "by default to whatever `pooch.os_cache("orix")` returns", since this is always true.

orix/data/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+__all__ = [
+    "sdss_austenite",
+    "sdss_ferrite_austenite",
+    #    "ti_orientations",
+]
+
+
+fetcher = ppooch.create(
```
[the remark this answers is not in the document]
hakonanes: Copy/paste from kikuchipy, but I forgot why I did it there. Will correct.

hakonanes:
> Is there a way to remove the cache?

Manually the directory in a file browser, or from Python

```python
>>> import os
>>> import pooch
>>> os.rmdir(pooch.os_cache("orix"))
```

Do you think this is useful information to include in the documentation?

> We should wait until this is sorted. Who owns the file?

Agree. The original location of a ~45 MB .zip file containing the ~28 MB .ctf file is here [https://doi.org/10.17863/CAM.8815](https://doi.org/10.17863/CAM.8815). The file was uploaded by Robert Krakow and Ralf Hielscher for their paper on 3D misorientation spaces [http://dx.doi.org/10.1098/rspa.2017.0274](http://dx.doi.org/10.1098/rspa.2017.0274). Thankfully, they licensed it with [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/). As long as we provide the file origin and license in the new orix-data repo and in the docstrings, I think we are good.

I'll create the orix-data repo and upload the files with origin and licenses once @pc494 approves.

hakonanes:
I hadn't thought about deleting the cache before you mentioned it... I looked at the cache directory of data sets from `kikuchipy.data`:

bilde

A total of 250 MB is used. A note on deleting old versions is definitely useful to have in the documentation.

orix/data/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+    else:
+        raise ValueError(
+            "The dataset must be (re)downloaded from the orix-data repository on GitHub"
+            " (https://github.com/pyxem/orix-data) to your local cache with the pooch "
+            "Python package. Pass `allow_download=True` to allow this download."
+        )
+
+
+def _fetch(filename, allow_download):
+    resolved_path = os.path.join(package_data_path, "..", filename)
```
harripj, line 86: I had a bit of a deeper look into this. I think that we shouldn't mix `os` and `pathlib` and I would advocate for `pathlib`. 

For `resolved_path` shouldn't we be looking at the `cache_data_path` rather than the `package_data_path`? We don't expect any data to be found in the latter if I understand correctly?

orix/data/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+
+fetcher = ppooch.create(
+    path=ppooch.os_cache("orix"),
+    base_url="",
+    version=__version__.replace(".dev", "+"),
+    env="ORIX_DATA_DIR",
+    registry=registry,
+    urls=registry_urls,
+)
+cache_data_path = fetcher.path.joinpath("data")
```
harripj, line 59: As `"data"` is already explicitly defined in the filenames, I think this should just be `fetcher.path`.

orix/data/__init__.py
harripj: This would become `resolved_path = str(cache_data_path.joinpath(filename))`

orix/data/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+            "Python package. Pass `allow_download=True` to allow this download."
+        )
+
+
+def _fetch(filename, allow_download):
+    resolved_path = os.path.join(package_data_path, "..", filename)
+    expected_hash = registry[filename]
+    if _has_hash(resolved_path, expected_hash):  # File already in data module
+        return resolved_path  # pragma: no cover
+    else:  # Pooch must download the data to the local cache
```
harripj, line 90: In this `else` statement we could then check `if allow_download`. The `else` would raise the error in `_cautious_downloader`, so that should be moved here.

This then avoids setting and deleting the `fetcher.allow_download` property.

harripj (changes requested):

orix/data/__init__.py
harripj: Something like this:
```python
def _fetch(filename, allow_download):
    resolved_path = str(cache_data_path.joinpath(filename))
    expected_hash = registry[filename]
    if _has_hash(resolved_path, expected_hash):  # File already in data module
        return resolved_path  # pragma: no cover
    else:  # Pooch must download the data to the local cache
        if allow_download:
            resolved_path = fetcher.fetch(filename, downloader=_cautious_downloader)
        else:
            raise ValueError(
            "The dataset must be (re)downloaded from the orix-data repository on GitHub"
            " (https://github.com/pyxem/orix-data) to your local cache with the pooch "
            "Python package. Pass `allow_download=True` to allow this download."
        )
    return resolved_path
```

orix/data/__init__.py
hakonanes: Thank you very much for looking at this in more detail. We're now using `pathlib` only.

hakonanes: You're right in that it's the cache path we want. In kikuchipy some data sets are part of the package included in the `data` module, which is why we check the `package_data_path` first. This is corrected.

orix/data/__init__.py
hakonanes: I've opted to remove "data" from the path entirely, resulting in "$HOME/.cache/orix/0.9.0/sdss/sdss_ferrite_austenite.ang" etc.

orix/data/__init__.py
hakonanes: Yup, very well spotted. I've simplified the logic based on your suggestion, and also made it safer without having to attach and remove the attribute. Thank you!

orix/data/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+    version=__version__.replace(".dev", "+"),
+    version_dev="master",
+    env="ORIX_DATA_DIR",
+    registry=registry_hashes,
+    urls=registry_urls,
+)
+
+
+def _fetch(filename, allow_download):
+    file_path_in_cache = _fetcher.path / filename
```
harripj, line 60: I didn't know you could use `/` to join paths- nice!

harripj (approved):

harripj:
This seems to be much cleaner and leaner now- thanks for your efforts!

pc494:
@hakonanes do you want a review on this from me?

hakonanes:
Thank you for offering, but I think @harripj's good review suffices.

But I'd like your green light on creating an "orix-data" repo in the pyxem organization similar to "kikuchipy-data". I plan to put the Ti_orientations.ctf file and the two SDSS data files (orix HDF5 and .ang) used elsewhere in the documentation (see https://github.com/pyxem/orix/pull/318#issuecomment-1090509677).

pc494:
> Thank you for offering, but I think @harripj's good review suffices.
> 
> But I'd like your green light on creating an "orix-data" repo in the pyxem organization similar to "kikuchipy-data". I plan to put the Ti_orientations.ctf file and the two SDSS data files (orix HDF5 and .ang) used elsewhere in the documentation (see [#318 (comment)](https://github.com/pyxem/orix/pull/318#issuecomment-1090509677)).

"orix-data" repo sounds like an excellent idea, I think making it clear where our data comes from and under what license is a big improvement.

hakonanes:
I've updated the top comment now that the [orix-data](https://github.com/pyxem/orix-data) repo is created.

hakonanes:
Fantastic, thanks @harripj and @pc494 for quick discussions and reviews so that we got a data module added so quickly.

PR #321, opened 2022-04-21
Add Dask-based lazy computation of quaternion-vector outer product

harripj:
There is already a Dask-based implementation of the quaternion-quaternion outer product, but currently no such implementation for quaternion-vector outer products. In this PR no new functions are added, but `Quaternion._outer_dask()` will now also work with `Vector3d` inputs, which is consistent with the behaviour of `Quaternion.outer()`.

I have also corrected the missing parentheses in the `Quaternion` docstring relating to quaternion-vector multiplication introduced when we switched to `numpy-quaternion` in #281 [here](https://github.com/pyxem/orix/pull/281/files#diff-5fa95cd8f92d9bc27b217689cd19a42fbf336678e6e4c2354f63a5fc37a512acL113-L124) (sorry about that).

#### Progress of the PR

```python
>>> import numpy as np
>>> v = Vector3d(np.random.rand(7, 5, 3)).unit
>>> q = Quaternion(np.random.rand(6, 2, 4)).unit
>>> arr = q._outer_dask(v)
>>> type(arr)
dask.array.Array
>>> v_new = Vector3d(arr.compute())
>>> v_new.shape
(6, 2, 7, 5)
```

#### For reviewers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Unreleased
 
 Added
 -----
+- Dask implementation of the `Quaternion` - `Vector3d` outer product.
 - Point group `Symmetry` elements can now be viewed under the stereographic projection
   using `Symmetry.plot()`. The notebook point_groups.ipynb has been added to the
   documentation.
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -58,11 +58,11 @@ class Quaternion(Object3d):
     :math:`v' = q * v * q^{-1}` and follows as:
 
     .. math::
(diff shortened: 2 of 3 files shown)


hakonanes (approved):
Looks very good, thanks!

`_outer_dask(quaternion)` is used in `Orientation.get_distance_matrix()`, while as you say `_outer_dask(vector)` isn't used in the public API. Many computations done with orix include large multiplications with a subsequent reduction based on minimization or symmetry operations... We should consider introducting a `lazy` boolean parameter where we can so that memory is kept in check during these larger computations. Something for another PR!

harripj:
> We should consider introducting a lazy boolean parameter where we can so that memory is kept in check during these larger computations. Something for another PR!

A very good idea!

Issue #323, opened 2022-04-25
Consolidate generator functions with `diffsims`


harripj:
There is a bug in `sample_S2_uv_mesh` due to inverted ranges for the azimuthal and polar angles. This can be fixed in a PR with some added tests, but as this code is essentially duplicated from `diffsims` I think it would be good to discuss where this code should live first, to improve code reuse and reduce bugs.

Both `orix` and `diffsims` have overlapping generator functions and I think it would be wise to consolidate these into just one package. Given the tight relationship between both packages I don't think this would cause too many problems. I know in my experience I often have to check in which package these generator functions live, so this would help in this case too. I see that `diffsims` imports from `orix` and not vice-versa and, to avoid circular imports, perhaps this is why some of these functions were duplicated into `orix` in the first place?

I think it would be a good idea to bring the `sphere_mesh_generators` module into orix, for example `get_uv_sphere_mesh_vertices` and `sample_S2_cube_mesh` already exist in `orix's` `S2_sampling` module, as discussed. I think the `rotation_list_generators` may also be a good idea to move over, but not so sure on this.

All of this said, I do not use `diffsims` too much so forgive me if I this is a reach- it would be great to get some thoughts on this from the `diffsims` developers!

harripj:
> Where does it show itself? Could you add an example?

This is the example using the current code:
```python
resolution = 10
steps_azimuth = int(np.ceil(360 / resolution))
steps_polar = int(np.ceil(180 / resolution)) + 1
azimuth = np.linspace(0, np.pi, num=steps_azimuth, endpoint=True)
polar = np.linspace(0, 2 * np.pi, num=steps_polar, endpoint=False)
azimuth_prod, polar_prod = np.meshgrid(azimuth, polar)
azimuth_prod = azimuth_prod.ravel()
polar_prod = polar_prod.ravel()
v = Vector3d.from_polar(azimuth=azimuth_prod, polar=polar_prod).unit

fig, ax = plt.subplots(subplot_kw=dict(projection='3d'))
ax.scatter(*v.data.T)
```
For a resolution of 10 there should be 19 elevation levels, which is not the case:
<img width="460" alt="image" src="https://user-images.githubusercontent.com/16853829/165113551-e58c7e72-f96c-4aad-992a-500748768117.png">

The bug is that `azimuth` should range from `[0..2*pi)` and `polar` from `[0..pi]`, and in the current implementation they are inverted. Using the updated code:
```python
[...]

PR #324, opened 2022-04-25
Fix `sample_S2_uv_mesh`

harripj:
As discussed in #323 there was a bug in `sample_S2_uv_mesh`. This has been corrected and extra tests have been introduced.

#### Progress of the PR

```python
>>> from orix.sampling.S2_sampling import sample_S2_uv_mesh
>>> from mpl_toolkits.mplot3d import Axes3D
>>> from matplotlib import pyplot as plt

>>> v = sample_S2_uv_mesh(10)

>>> fig, ax = plt.subplots(subplot_kw=dict(projection='3d'))
>>> ax.scatter(*v.data.T)

>>> np.isclose(v.z, 1).sum()
1
```
image

#### For reviewers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -50,7 +50,7 @@ Deprecated
 
 Fixed
 -----
-- Fixed bug in `sample_S2_uv_outer()` and removed duplicate vectors at poles.
+- Fixed bug in `sample_S2_uv_mesh()` and remove duplicate vectors at poles.
 - The results from `Orientation.dot_outer()` are now returned as 
   `self.shape + other.shape`, which is consistent with `Rotation.dot_outer()`.
 - Writing of property arrays in .ang writer from masked CrystalMap.


orix/sampling/S2_sampling.py
```diff
@@ -48,11 +48,21 @@ def sample_S2_uv_mesh(resolution):
...
     azimuth_prod = azimuth_prod.ravel()
     polar_prod = polar_prod.ravel()
+    # remove duplicated vectors at north (polar == 0) and
+    # south (polar == np.pi) poles. Keep the azimuth == 0 vector in each
+    # case. Masks here are vectors to remove
+    mask_azimuth = azimuth_prod > 0
+    mask_polar_0 = np.isclose(polar_prod, 0) * mask_azimuth
+    mask_polar_pi = np.isclose(polar_prod, np.pi) * mask_azimuth
+    # create mask of vectors to keep
+    mask = ~np.logical_or(mask_polar_0, mask_polar_pi)
```
hakonanes, line 63: Might `np.logical_or.reduce((mask_azimuth, mask_polar_0, mask_polar_pi))` be a bit more efficient, to prevent two multiplications?

hakonanes (approved):
Great, thanks for this bug fix!

I've got one minor suggestion for simplification of the vector masking, but it's not necessary to address this before merging.

orix/sampling/S2_sampling.py
harripj: I don't think this works as the two multiplications with `mask_azimuth` are `logical_and` operators. For `mask_polar_0` I am calculating a mask of where the points have `0` polar angle and non-zero azimuthal angle (to remove duplicates), and the same for polar `pi`.

hakonanes: Of course, I was too quick there. Sorry. Mergin!

PR #325, opened 2022-04-27
Add `lazy` parameter to `Quaternion` and `Rotation` `.outer()` for easy `Dask` computation

harripj:
Now that `Vector3d` outer products can be computed in `Quaternion._outer_dask()`, and as previously discussed, this PR allows this functionality to be accessed using the public API through `.outer()`. In line with `._outer_dask()` 2 keyword arguments `lazy` and `chunksize` are added, and `lazy` defaults to `False` to conserve the current behaviour.

`Rotation._outer_dask()` has also been removed as it can be called though the inherited `Quaternion._outer_dask()` method.

#### Progress of the PR

```python
>>> import numpy as np
>>> r1 = Rotation.random((5, 3))
>>> r2 = Rotation.random((11, 4))
>>> r12 = r1.outer(r2)
>>> r12_lazy = r1.outer(r2, lazy=True, chunk_size=20)
>>> np.allclose(r12.data, r12_lazy.data)
True

>>> r = Rotation.random((5, 3))
>>> v = Vector3d(np.random.rand(6, 4, 3)).unit
>>> v2 = r.outer(v)
>>> v2_lazy = r.outer(v, lazy=True, chunk_size=20)
>>> np.allclose(v2.data, v2_lazy.data)
True
```

#### For reviewers

### DIFF
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -230,9 +230,9 @@ class Quaternion(Object3d):
         ----------
         other : orix.quaternion.Quaternion or orix.vector.Vector3d
         lazy : bool
-            Whether to computer this computation using Dask. This is
-            option can be used to reduce memory usage when working with
-            large arrays. Default is False.
+            Whether to computer this computation using Dask. This option
+            can be used to reduce memory usage when working with large
+            arrays. Default is False.
         chunk_size : int, optional
             When using `lazy` computation, `chunk_size` represents the
             number of objects per axis for each input to include in each
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -230,10 +230,12 @@ class Quaternion(Object3d):
         ----------
         other : orix.quaternion.Quaternion or orix.vector.Vector3d
         lazy : bool
-            Whether to computer this computation using Dask. Default is
-            False.
+            Whether to computer this computation using Dask. This is
+            option can be used to reduce memory usage when working with
+            large arrays. Default is False.
         chunk_size : int, optional
-            Number of objects per axis for each input to include in each
+            When using `lazy` computation, `chunk_size` represents the
+            number of objects per axis for each input to include in each
             iteration of the computation. Default is 20.
 
         Returns
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -19,6 +19,7 @@
 import warnings
 
 import dask.array as da
+from dask.diagnostics import ProgressBar
 import numpy as np
 import quaternion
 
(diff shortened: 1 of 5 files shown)


hakonanes:
Fantastic! I've got time to review from tomorrow. Just ping me.

orix/quaternion/quaternion.py
```diff
@@ -222,34 +222,49 @@ def mean(self):
         w_max = np.argmax(w)
         return self.__class__(v[:, w_max])
 
-    def outer(self, other):
+    def outer(self, other, lazy=False, chunk_size=20):
```
hakonanes, line 225: `get_distance_matrix()` supports controlling whether a progressbar is shown during computation, this should be added here to maintain a consistent lazy API.

orix/quaternion/quaternion.py
```diff
@@ -222,34 +222,49 @@ def mean(self):
...
         """Compute the outer product of this quaternion and the other
         quaternion or vector.
 
         Parameters
         ----------
         other : orix.quaternion.Quaternion or orix.vector.Vector3d
+        lazy : bool
```
hakonanes, line 232:
```suggestion
        lazy : bool, optional
```

orix/quaternion/quaternion.py
hakonanes: The calls to the initialization methods are those that should then be inside a `with ProgressBar():` context if `progressbar` is `True`, because that's when the Dask arrays are converted to NumPy arrays.

orix/quaternion/rotation.py
```diff
@@ -259,22 +259,35 @@ def angle_with_outer(self, other):
         angles = np.nan_to_num(np.arccos(2 * dot_products**2 - 1))
         return angles
 
-    def outer(self, other):
+    def outer(self, other, lazy=False, chunk_size=20):
```
hakonanes, line 262: The `progressbar` parameter should be supported here as well.

orix/quaternion/quaternion.py
hakonanes: Likewise below.

orix/quaternion/quaternion.py
hakonanes: A couple extra lines to the tests catching the cases when `progressbar=True` must of course be added. I think it's unnecessary to actually check that the progressbar is shown.

hakonanes (changes requested):
Very nice enhancement and clean-up!

I've suggested that we support the `progressbar` parameter in the methods newly supporting lazy computation, to maintain a consistent lazy API. Wherever `lazy` is supported, some control over chunk size via `chunksize` and whether to show the progressbar via `progressbar` is a sufficient lazy API at the moment, I think.

orix/quaternion/quaternion.py
harripj: Very good point. I think this makes the API simple and consistent.

hakonanes (approved):
Great, thanks for adding the `progressbar` parameter. I'll merge tomorrow if you don't say otherwise, or you can merge yourself :)

PR #331, opened 2022-05-10
Fix reST formatting in project description in source distribution

hakonanes:
The source distribution couldn't be uploaded to PyPI ([log](https://github.com/pyxem/orix/runs/6367065668?check_suite_focus=true)) due to our newly added "raw" HTML directive in "README.rst", which PyPI doesn't allow. I've now removed this part before adding it to `setup()`. I'll manually re-run the publish action after this PR is merged.

Also, the weekly action checking the user guide notebooks with `nbval` now installs the `doc` dependencies. I've also added a manual trigger for this workflow.

#### Progress of the PR

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,6 +7,10 @@
       </h1>
     </p>
 
+.. Content above here until EXCLUDE plus one line is excluded from the long description
+.. in the source distributions uploaded to PyPI
+.. EXCLUDE
+
 |binder|_ |build_status|_ |Coveralls|_ |docs|_ |pypi_version|_  |downloads|_ |black|_ |doi|_
 
 .. |binder| image:: https://mybinder.org/badge_logo.svg
--- a/.github/workflows/test_documentation_notebooks.yml
+++ b/.github/workflows/test_documentation_notebooks.yml
@@ -8,6 +8,8 @@ name: Test documentation notebooks
 on:
   schedule:
     - cron: '15 6 * * 1' # Run every Monday morning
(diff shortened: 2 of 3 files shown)


hakonanes:
I checked that PyPI allows the newly generated long description by running

```bash
python -m build
twine check dist/*
```

locally.

README.rst
```diff
@@ -7,6 +7,10 @@
       </h1>
     </p>
 
+.. Content above here until EXCLUDE plus one line is excluded from the long description
```
harripj, line 10: I have read that comments are best to not be defined inline: https://stackoverflow.com/a/4783854/12063126

Although I think in this case this will be okay, what do you think?

harripj (commented):
Just one comment, otherwise LGTM!

README.rst
hakonanes: I don't know of a way we can test how PyPI will display the `long_description` file (reST) we provide, so it is good to be on the safe side.

hakonanes: Actually, the [documentation](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#comments) states quite clearly that this is valid syntax, so I think we're safe.

hakonanes: I suggest to leave as is.

harripj (approved):
Merging!

hakonanes:
Great, thanks!

Issue #299, opened 2022-02-28
Release 0.9.0


hakonanes:
Due to recent improvements to the code base by @harripj, we've decided to make a 0.9 release. The improvements break the API, to the point where many of kikuchipy's tests fail (in a straightforward way), so it would be good to get them released and tested by users and in downstream packages before releasing 1.0 🚀 

To include in 0.9 (add or remove as you see fit):
- [x] #229
- [x] #296
- [x] #297
- [x] #306 
- [x] #308 
- [x] #303 
- [x] #294
- [x] #301
- [x] #302
- [x] #304
- [x] #305 
- [x] #307
- [x] #310   
- [x] #312
- [x] #321 

I propose to make a release candidate (rc) 0.9.0rc1 on PyPI (I think conda-forge is unnecessary) so that kikuchipy can be updated to work with the coming 0.9.0. I can handle these releases unless someone else wants to.

Apart from the benfit of having some time between 0.9 and 1.0, there is no particular rush to get this out, at least not from my end.

Input on this plan is as always welcome!

harripj:
I think a `0.9` release is a good idea for the reasons you mentioned @hakonanes.

Regarding the changes #296 and #298 I think are improvements and additions which do _should_ not break anything. #295 may require some minor reworking to existing code to avoid `memoryviews` popping up everywhere. #294 however may change existing code and should be checked thoroughly on affected code, as discussed in the PR.

I agree that there is not necessarily a rush to package this up, and that we should give some time after this release for other projects to adopt the changes and catch up.

pc494:
> I propose to make a release candidate (rc) 0.9.0rc1 on PyPI (I think conda-forge is unnecessary)

I agree with a release candidate, and iirc conda-forge don't accept release candidates anyway.

harripj:
I think we're getting pretty close to finishing this up, after #303 and #308 are reviewed and merged. #296 will be closed by #303. @hakonanes will you look further into #297? It doesn't seem to be a big issue as discussed in that PR. Finally a bit of tidying up with #229.

hakonanes:
I'll do the tasks in #297, and also try to find a way to include the misorientation clustering notebook in the Read the Docs documentation. We can then archive the orix-demos repo :tada:

pc494:
Ah also! After this is out the door we should move from `master` to `main` (unless we've already discussed this)

hakonanes:
[...]

PR #333, opened 2022-05-10
Remove restriction to upload to PyPI via action

hakonanes:

See https://github.com/pyxem/orix/issues/299#issuecomment-1122451887.

#### Progress of the PR

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -40,7 +40,6 @@ jobs:
         password: ${{ secrets.TEST_PYPI_API_TOKEN }}
         repository_url: https://test.pypi.org/legacy/
     - name: Publish package to PyPI
-      if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
       with:
         user: __token__


hakonanes:
Good questions.

The "publish" action is only triggered if we
1. publish a tagged release or
2. trigger it manually

Currently, an upload to PyPI is attempted in case 1. If this PR is merged, an upload is also attempted in case 2.

An upload to PyPI is successful if the attempted upload's version isn't available on PyPI. This means that we cannot upload two 0.9.0, for example. If 0.9.0rc1 isn't on PyPI, it will be uploaded. See for example [NumPy's](https://pypi.org/project/numpy/#history) PyPI release history.

harripj:
Thanks for answering! I see, the RC release is not tagged, so the upload needs to be triggered manually, but the action currently prevents the release due to the `if` statement. Yes we should correct this behaviour.

harripj (approved):

hakonanes:
Actually, the RC release *was* tagged as `v0.9.0rc1`, but the tagged version of `master` couldn't be uploaded due to the unsupported "raw" directive included in the reST string passed to `long_description` in `setup()`. We fixed this, but there is no way to update the tag via GitHub. The alternative is to
1. Delete the release on GitHub (upstream)
2. Overwrite the upstream tag locally, including the updates
3. Force push the new local tag to upstream
4. Create new release on GitHub pointing to the new tag

hakonanes:
It's worth mentioning that we can delete uploads to PyPI via their web page if we uploaded something accidentally.

harripj:
Merging.

PR #334, opened 2022-05-10
Migrate S2 samplings from diffsims, add hexagonal sampling

din14970:
This is a first in some PRs I still hope to make in order to clean up and enhance `diffsims` a bit by taking out functionality which I believe should reside in `orix`. May take me quite a long time to do so but we'll see how far we get. It seems not all S2 samplings from `diffsims` were copied to orix. This PR moves all of them and gives them names consistent with the other functions. In addition, I've implemented a new "hexagonal" sampling:

```python
>>> from orix.sampling import S2_sampling
>>> points = S2_sampling.sample_S2_hexagonal_mesh(5)
>>> points.scatter(s=2)
```

image

My reasoning for this is that for a "beam direction grid" sampling you usually want the edges of the fundamental zone to be sampled. For the existing grids, the sampled fundamental zones for trigonal and hexagonal crystals will have a "rough" edge. This mesh solves that.

In addition, I've refactored the cube mesh sampling a bit to make it more readable and hopefully maintainable.

What is not included in this PR:
* in diffsims, these meshes are [filtered by crystal system](https://github.com/pyxem/diffsims/blob/2b7532946092fe4a8b133f8b2f7ad638fe277cf8/diffsims/generators/rotation_list_generators.py#L173) and converted to orientations and Euler angles, using the [beam directions grid to euler fuction](https://github.com/pyxem/diffsims/blob/2b7532946092fe4a8b133f8b2f7ad638fe277cf8/diffsims/generators/sphere_mesh_generators.py#L479). I'm wondering how this can be streamlined. In my vision, everything related to orientations and vectors is moved to orix. That means:
  * the "filtering" of orientations should ideally use orix' symmetry functionality based on the (diffpy) crystal structure provided. now a user has to specify crystal symmetry with a string.
  * the conversion of a beam direction to an orientation should follow the very specific constraint that phi_1 = 0. The question is where such a function can fit best into orix.

#### Progress of the PR

see above

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


hakonanes:
Great that you're taking the time to add this functionality to orix!

> the "filtering" of orientations should ideally use orix' symmetry functionality based on the (diffpy) crystal structure provided. now a user has to specify crystal symmetry with a string.

This is straight forward using the `FundamentalSector` (derived from MTEX using orix' `OrientationRegion`) which defines a region on *S2*. It has implemented `__eq__`, which checks whether a vector is inside the region, allowing simple slicing. I've added a Gist showing this [here](https://gist.github.com/hakonanes/72d03a886a8d06aa06a64b382637aa92).

Note the different crystal axes alignment in orix compared to diffpy.structure's default, described [here](https://orix.readthedocs.io/en/latest/crystal_reference_frame.html). This might be the main reason for the differences between the monoclinic, orthorhombic, trigonal and hexagonal grids in `diffsims` and `orix`.

I'll review this today or tomorrow. We should follow this PR up with a "Vectors/Sampling" user guide notebook!

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def _compose_from_faces(corners, faces, n):
+    """
+    Helper function to refine a grid starting from a platonic solid,
+    adapted from meshzoo
+
+    Parameters
+    ----------
+    corners: numpy.ndarray (N, 3)
```
hakonanes, line 99: Space around colon here.

hakonanes: This docstring could need some capitalization and periods.

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,265 @@
...
+
+    See also
+    --------
+    :func:`get_icosahedral_mesh_vertices`
+    """
+    # create corner nodes
+    vertices = [corners]
+    vertex_count = len(corners)
+    corner_nodes = np.arange(len(corners))
+    # create edges
```
hakonanes, line 120: I suggest to add some blank lines to separate sections.

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,265 @@
...
+def _edge_grid_spherified_corner_cube(resolution: float):
+    center_of_face_to_corner = np.sqrt(2)
+    number_of_steps = _number_of_equiangular_steps(resolution, center_of_face_to_corner)
+    grid_on_diagonal = _sample_length_equiangular(
+        number_of_steps, center_of_face_to_corner
+    )
+    return grid_on_diagonal / center_of_face_to_corner
+
+
+def _compose_from_faces(corners, faces, n):
```
hakonanes, line 92: I'm not able to follow the code in this function. Has it been much used in diffsims? If not, do you think it will be more used if in orix? I'm asking because I'm not sure if I'm able to maintain the function. Of course we can leave it in if you're willing to maintain it.

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,265 @@
+import numpy as np
+from scipy.spatial import cKDTree
+
+
+def _get_start_and_end_index(
+    number_of_steps: float,
```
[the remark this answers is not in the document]
harripj: I was thinking at some point that adding type hints would be a good addition to orix.

I also think that this doesn't need to be done all in one go but can be done gradually, and maybe completed before 1.0? Happy to include them starting from here! In that case I would prefer if all functions in this PR used typehints.

On a separate note, I could not find how best to include type hints in the [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html) but currently it means possibly duplicating the parameter types as both type hints and in the docstring, which I think is redundant. If we can get it to work with Sphinx then I would prefer to have type hints and remove type parameters in the docstrings. What do you think?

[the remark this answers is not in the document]
harripj: Looks like [this comment](https://stackoverflow.com/a/66295922) could be a solution.

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +134,166 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+    Vector3d
+    """
+    number_of_steps = int(np.ceil(2 / np.tan(np.radians(resolution))))
+    if number_of_steps % 2 == 1:
+        # an even number of steps is required to get a point in the middle
+        # of the hexagon edge
+        number_of_steps += 1
+    grid_1D = _sample_length_equidistant(
+        number_of_steps,
+        length=1.0,
```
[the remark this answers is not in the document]
din14970: No, but I like to write this explicitly to indicate it is a float

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,265 @@
...
+        number of times the mesh is refined
+
+    Returns
+    -------
+    vertices: numpy.ndarray (N, 3)
+        The coordinates of the refined mesh vertices.
+
+    See also
+    --------
+    :func:`get_icosahedral_mesh_vertices`
```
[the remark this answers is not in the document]
din14970: it was the name of the function that calls `_compose_from_faces` in diffsims, but I don't think this is necessary here, especially since I've moved it to a different module. So now I've removed it

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,265 @@
...
+        include_end,
+        positive_and_negative,
+    )
+    grid_spacing = length / number_of_steps
+    grid_on_edge = np.arange(start_index, end_index) * grid_spacing
+    return grid_on_edge
+
+
+def _number_of_equiangular_steps(resolution: float, length: float) -> int:
+    total_angle = np.arctan(length)
```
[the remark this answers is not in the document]
din14970: Length is actually a unit of distance. The sphere used to generate the mesh has radius 1, so this "length" is actually the distance on the cube around the sphere.

orix/sampling/_polyhedral_sampling.py
din14970: To be honest, I'm not able to follow the code either, it came more or less directly from meshzoo. The problem with using meshzoo directly was that the functions not only generated points, but also the triangles connecting them. Plus, adding meshzoo as dependency to diffsims seemed a bit overkill for a pretty small feature -- hence a copy of only the relevant bits of code. This specific function is necessary for the icosahedral sampling which I think might be a nice feature to have. The function has no complex operations or dependencies which would have a high chance of getting depricated in future versions of numpy or python, so I'm pretty sure it can be left as-is. If it ever turns out to be unmaintainable, both icosahedral sampling and this function could be removed.

orix/sampling/_polyhedral_sampling.py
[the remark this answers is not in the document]
din14970: In my current job we use type annotations for everything mainly to be able to use [mypy](http://mypy-lang.org/). Has quite a lot of advantages as you can catch dumb issues early during development using a linter (I'm using python language server with a mypy plugin) instead of at runtime. For me this is the main selling point over just annotating them in the documentation. A downside of type annotiations as they currently exist is that you can't specify more details about the types, like shapes of numpy arrays. This IS often done in documentation, since return type or parameter type `np.ndarray` is usually not very informative. There is now `numpy.typing` <https://numpy.org/devdocs/reference/typing.html> with which dtypes of arrays can be specified. It seems shapes for type annotations are still on the roadmap https://numpy.org/neps/roadmap.html.

orix/sampling/_polyhedral_sampling.py
din14970: I have had another look at this function and it seems a large part of it was only necessary for generating edges/faces. I've now stripped down the function to the bare essentials for only generating the vertices. I've also improved the icosahedral sampling by now using basic geometry to calculate how large `n` should be, instead of finding `n` through a trial and error loop. This means that the `_compose_from_faces` function only needs to be called once.

din14970:
@hakonanes I could work out a short notebook but I might copy paste some stuff from the gist you shared if that's OK with you. Will also probably take some time still.

> This might be the main reason for the differences between the monoclinic, orthorhombic, trigonal and hexagonal grids in diffsims and orix.

For diffsims to continue to work properly for the template matching in pyxem, I'm wondering whether it is important which fundamental zone is sampled. I guess we'll find out when I get to it.

@harripj docs don't seem to have an issue with the type hints AFAICT

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +134,166 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+def sample_S2_random_mesh(resolution, seed=None):
+    """Vectors of a random mesh on *S2*
+
+    Parameters
+    ----------
+    resolution : float
+        The expected mean angle between nearest neighbor
+        grid points in degrees.
+    seed : int, optional
+        passed to np.random.default_rng(), defaults to None which
```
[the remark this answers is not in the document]
hakonanes: We have to spell out "numpy" for intersphinx to catch the reference (see [here](https://orix--334.org.readthedocs.build/en/334/reference.html#orix.sampling.sample_S2_random_mesh)).

orix/sampling/_polyhedral_sampling.py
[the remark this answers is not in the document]
hakonanes: We use type hints in kikuchipy without duplication in the docstring, see e.g. [doc](https://kikuchipy.org/en/stable/reference.html#kikuchipy.signals.EBSD.dictionary_indexing) and [code](https://github.com/pyxem/kikuchipy/blob/2b6feb33a8852b89cbd089721fecc99c4890a6c5/kikuchipy/signals/ebsd.py#L952). We use the [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) to help Sphinx with the type hints. But, based on [the docs](https://orix--334.org.readthedocs.build/en/334/reference.html#orix.sampling.sample_S2_random_mesh) from one of the functions in this PR, it seems we don't need that? Would be good to check the same function without duplication in the docstring.

orix/sampling/_polyhedral_sampling.py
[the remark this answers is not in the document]
hakonanes: Thanks for clarifying!

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,232 @@
...
+    # create edges
+    edges = set()
+    for face in faces:
+        edges.add(tuple(sorted([face[0], face[1]])))
+        edges.add(tuple(sorted([face[1], face[2]])))
+        edges.add(tuple(sorted([face[2], face[0]])))
+
+    # create edge nodes:
+    t = np.linspace(1 / n, 1.0, n - 1, endpoint=False)
+    k = corners.shape[0]
```
hakonanes, line 151: Seems to me like `k` can be removed?

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,232 @@
...
+# orix is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with orix.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Tuple, List
+import numpy as np
```
hakonanes, line 20: Final comments on code style not picked up by `black`:
1. Add a blank line between imports of standard libraries and third party libraries
2. Start docstring top description on line with """, to be in line with (all?) other docstrings in orix
3. One space missing around colon in docstring parameter on line 132

orix/sampling/_polyhedral_sampling.py
hakonanes: Thank you for making the effort, it is much easier to follow now!

hakonanes:
> I would suggest we push this to post 0.9.0 as it might well need a couple of weeks. That way we can have a "stable" trio of 0.14 (pyxem) 0.9.0 (orix) and 0.5.0 (diffsims) to expand on.

Sounds good to me, since making a nice vector sampling user guide isn't done by tomorrow when we plan to release orix 0.9. If it turns out some downstream package needs this functionality to build upon, I'm OK with releasing a 0.10 with this in soon if necessary.

> I might copy paste some stuff from the gist you shared if that's OK with you.

Sure, go ahead! I was thinking that as a temporary solution in diffsims, we can hard-code the fundamental sectors with the existing corners in diffsims to reduce the vectors. This should produce the exact list of current vectors in diffsims.

> Will also probably take some time still. Otherwise, I would say this PR should be in decent shape.

That's OK, of course. Since we don't have any better way to cut a release in orix than from `master`, I suggest to merge this PR *after* 0.9 is released.

hakonanes (changes requested):
As you say, this PR is in decent shape!

I've got four outstanding review comments, and also the two points in https://github.com/pyxem/orix/pull/334#pullrequestreview-971070726.

orix/sampling/_polyhedral_sampling.py
din14970: good catch

orix/sampling/_polyhedral_sampling.py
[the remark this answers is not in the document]
hakonanes: > Would be good to check the same function without duplication in the docstring.

I checked. From this PR with docstring duplication:

bilde

Without docstring duplication and without `sphinx-autodoc-typehints` (built locally):

bilde

Without docstring ducplication and with `sphinx-autodoc-typehints` (built locally):

bilde

My preference is for without docstring duplication and without `sphinx-autodoc-typehints`.

Another question is if the docs can be built nicely with the minimal version of Python orix supports, 3.6.

orix/sampling/_polyhedral_sampling.py
din14970: > Add a blank line between imports of standard libraries and third party libraries

Have you ever looked into [`isort`](https://pycqa.github.io/isort/)? I can never remember the rules on imports so I just use that. But it seems when I ran it on the orix codebase tests stopped working, so it must be that in some files there is a very deliberate order for imports.

orix/sampling/_polyhedral_sampling.py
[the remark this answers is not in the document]
din14970: Ok, so I will remove the duplicated type hints in the documentation then

orix/sampling/_polyhedral_sampling.py
hakonanes: Thank you for making the changes.

I haven't used it. I always group imports into three groups of std lib, 3rd party and internal, and list these alphabetically, unless another order is recquired. By the way, this is stated in the [contributing guide](https://orix.readthedocs.io/en/stable/contributing.html#code-style). If using isort hadn't caused any issues as you say, I'd be happy for use to start using it. I run black via pre-commit, and it seems like isort integrates well with those two.

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +140,169 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+    include_points = z > tolerance
+    points_one_face = np.stack([coordinate[include_points] for coordinate in [x, y, z]])
+
+    # repeat 6 times by rotating 60 degrees
+    def rotation(r):
+        return np.array(
+            [[np.cos(r), -np.sin(r), 0], [np.sin(r), np.cos(r), 0], [0, 0, 1]]
+        )
+
+    angle = np.radians(60)
```
harripj, line 201: As @hakonanes mentioned could this be made `deg2rad` too?

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,232 @@
...
+    if not include_start:
+        start = start + 1
+    end = number_of_steps
+    if include_end:
+        end = end + 1
+    return start, end
+
+
+def _number_of_equidistant_steps(resolution: float, length: float) -> int:
+    maximum_grid_spacing = np.tan(np.radians(resolution))
```
harripj, line 44: `deg2rad` here too.

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,232 @@
...
+        positive_and_negative,
+    )
+    grid_spacing = length / number_of_steps
+    grid_on_edge = np.arange(start_index, end_index) * grid_spacing
+    return grid_on_edge
+
+
+def _number_of_equiangular_steps(resolution: float, length: float) -> int:
+    total_angle = np.arctan(length)
+    return int(np.ceil(total_angle / np.radians(resolution)))
```
harripj, line 68: `deg2rad` here too.

harripj (changes requested):
A very nice addition to orix which greatly improves the sampling capabilities. I just few a few small comments below.

din14970:
I've added a very basic notebook under `vectors`, would be happy to get some feedback (or direct modifications) on this

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e39db862",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib inline"
```
harripj, line 31: This cell could be combined with the one below.

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "\n",
+    "from orix import plot\n",
+    "from orix.crystal_map import Phase\n",
+    "from orix.quaternion import Orientation, symmetry\n",
+    "from orix.sampling import S2_sampling\n",
+    "from orix.vector import Vector3d"
```
harripj, line 48: From my testing `Phase`, `Orientation`, and `Vector3d` are not needed to run this notebook, so I suggest to remove them.

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "a8fe0e70",
+   "metadata": {},
+   "source": [
+    "## Sampling the fundamental zone\n",
+    "Sampling of directions is most relevant when combined with crystal symmetry functionality.\n",
+    "In this case, we would want to sample vectors in the real crystallographic basis.\n",
+    "Usually one only wants to sample the fundamental zone, not the entire sphere, as in the context of a crystal, many vectors are symmetrically equivalent.\n",
+    "The fundamental zone of a point group represents the smallest region of directions unique by symmetry.\n",
```
harripj, line 89: Could I suggest rewriting this as: "the smallest region of symmetrically unique directions."? At least for myself this is more clear!

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+    "The higher the symmetry of the point group, the smaller the fundamental zone.\n",
+    "\n",
+    "The symmetry operations are independent of the S2 sampling methods, so any set of vectors can be limited to the fundamental zone.\n",
```
harripj, line 92: Could you use LaTeX formatting on $S_2$ (as done in the top of the notebook).

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+   "outputs": [],
+   "source": [
+    "fig, axes = plt.subplots(2, 4, subplot_kw=dict(projection=\"stereographic\"), \n",
+    "                         figsize=(20, 10), facecolor=\"white\")\n",
+    "axes = axes.ravel()\n",
+    "\n",
+    "resolution = 5\n",
+    "\n",
+    "for i, (name, sampling_method) in enumerate(S2_sampling.SAMPLING_METHODS.items()):\n",
+    "    vectors = sampling_method(resolution = 5)\n",
```
harripj, line 75: Could you remove the spaces around `=`?

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+    "    axes[i].set_title(name, fontsize=20)\n",
+    "    axes[i].scatter(vectors, s=2);"
+   ]
+  },
```
harripj, line 79: It would also be nice to fill in the final plot. I have created an equal area sampling implementation on S2 in #340 which I think would fit nicely here, if that would be okay with you once that PR is accepted. What do you think?

harripj (changes requested):
I really like the notebook- a great addition to the docs, thanks @din14970! I have just a few comments below.

doc/s2_sampling.ipynb
hakonanes: I suggest to just write "S2" in the title, because the title is used in the sidebar and it takes bit of time for "\(S_2\)" to be rendered in LaTeX, and the browser tab title does not support LaTeX so it will always read "\(S_2\)".

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +144,185 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+    # icosahedron edge length
+    a = np.linalg.norm(corners[0]) / np.sin(2 * np.pi / 5)
+    # icosahedron inscribed sphere radius
+    r_i = np.sqrt(3) / 12 * (3 + np.sqrt(5)) * a
+    n = int(np.ceil(a / (r_i * np.tan(np.deg2rad(resolution)))))
+    vertices = _compose_from_faces(corners, faces, n)
+    return Vector3d(vertices).unit
+
+
+SAMPLING_METHODS: Mapping[str, Callable] = {
```
hakonanes, line 318: This dictionary prompts me to review the import structure of the `orix.sampling` module, which I've never really liked. Do you have a suggestion for how this could be simpler, @din14970?

Since this dictionary is used in the user guide, people will use it as well. I'm not sure I understand the benefit of the dictionary other than it is convenient to use in the user guide, was this your motivation?
 
I find it confusing that using for example `orix.sampling.S2_sampling.SAMPLING_METHODS["normalized_cube"]` and `orix.sampling.S2_sampling.sample_S2_cube_mesh()` have different defaults for grid type, is there a reason for this?

hakonanes, line 318: This dictionary prompts me to review the import structure of the `orix.sampling` module, which I've never really liked. Do you have a suggestion for how this could be simpler, @din14970?

Since this dictionary is used in the user guide, people will use it as well. I'm not sure I understand the benefit of the dictionary other than it is convenient to use in the user guide, was this your motivation?
 
I find it confusing that using for example `orix.sampling.S2_sampling.SAMPLING_METHODS["normalized_cube"]` and `orix.sampling.S2_sampling.sample_S2_cube_mesh()` have different defaults for grid type, is there a reason for this?

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+   "execution_count": null,
+   "id": "4c696bea",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, axes = plt.subplots(2, 4, subplot_kw=dict(projection=\"stereographic\"), \n",
+    "                         figsize=(20, 10), facecolor=\"white\")\n",
+    "axes = axes.ravel()\n",
+    "\n",
+    "resolution = 5\n",
```
hakonanes, line 72: This variable is unused as far as I can tell.

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+    "\n",
+    "for i, (name, sampling_method) in enumerate(S2_sampling.SAMPLING_METHODS.items()):\n",
+    "    vectors = sampling_method(resolution = 5)\n",
+    "    axes[i].set_title(name, fontsize=20)\n",
+    "    axes[i].scatter(vectors, s=2);"
```
hakonanes, line 77: I suggest to add the axes labels X, Y to the plots via `...scatter(vectors, s=2, axes_labels=["X", "Y"])`

doc/s2_sampling.ipynb
hakonanes: An alternative is to hide it with `axes[-1].set_visible(False)`.

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a8fe0e70",
+   "metadata": {},
+   "source": [
+    "## Sampling the fundamental zone\n",
+    "Sampling of directions is most relevant when combined with crystal symmetry functionality.\n",
+    "In this case, we would want to sample vectors in the real crystallographic basis.\n",
+    "Usually one only wants to sample the fundamental zone, not the entire sphere, as in the context of a crystal, many vectors are symmetrically equivalent.\n",
```
hakonanes, line 88: I suggest to replace "vectors" with "directions" in "... many vectors are symmetrically equivalent".

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+    "The fundamental zone of a point group represents the smallest region of directions unique by symmetry.\n",
+    "The higher the symmetry of the point group, the smaller the fundamental zone.\n",
+    "\n",
+    "The symmetry operations are independent of the S2 sampling methods, so any set of vectors can be limited to the fundamental zone.\n",
+    "However, different meshes may be most appropriate for different point groups, as we illustrate.\n",
+    "In general, to sample vectors for crystals with $\\alpha=\\beta=\\gamma=90^\\circ$, one of the `cube` meshes is recommended.\n",
```
hakonanes, line 94: A link to the "cube" mesh function(s) in the API reference would be good here as well.

hakonanes: Also the case for the hexagonal mesh function(s) and icosahedral mesh function.

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+    "    \"triclinic\": \"icosahedral\",\n",
+    "    \"monoclinic\": \"icosahedral\",\n",
+    "    \"orthorhombic\": \"spherified_cube_edge\",\n",
+    "    \"tetragonal\": \"spherified_cube_edge\",\n",
+    "    \"cubic\": \"spherified_cube_edge\",\n",
+    "    \"trigonal\": \"hexagonal\",\n",
+    "    \"hexagonal\": \"hexagonal\",\n",
+    "}\n",
+    "\n",
+    "fig, axes = plt.subplots(len(point_groups), 2, subplot_kw=dict(projection=\"stereographic\"), figsize=(10, 5 * len(point_groups)), facecolor=\"white\")\n",
```
hakonanes, line 133: Breaking this line up ensures we don't have to scroll horizontally in the user guide cell.

doc/s2_sampling.ipynb
hakonanes: Also, setting the Matplotlib preferences in the top cell makes reading the plotting commands simpler:

```python
plt.rcParams.update({"font.size": 20, "lines.markersize": 2})
```

hakonanes (changes requested):
Very nice and pedagogical user guide notebook. I've made some new comments and also added to Paddy's comments. The comment on the `orix.sampling` module import structure is the most important one.

I don't think this PR should include more functionality, so after the current review comments are sorted, I suggest to merge.

doc/s2_sampling.ipynb
din14970: I will make it S2 in the title and $S_2$ elsewhere then?

hakonanes: Perfect!

din14970: Actually might drop S2 and just write sampling unit vectors, S2 is redundant

doc/s2_sampling.ipynb
din14970: @harripj is the "random" meshing not equivalent to this? https://mathworld.wolfram.com/SpherePointPicking.html?

I will for now go with hiding the axes, thanks @hakonanes

doc/s2_sampling.ipynb
din14970: yes I did it in the cell above but forgot in this one, thanks for the catch

doc/s2_sampling.ipynb
hakonanes: Sorry, this is no surprise, since `axes_labels` is only a parameter of `Vector3d.scatter()` and not `StereographicPlot.scatter()`. See https://github.com/pyxem/orix/blob/14639a15a4c66c87891debb8656ef187c3b6e9ea/orix/vector/vector3d.py#L722.

`axes[i].set_labels("X", "Y")` should work.

orix/sampling/S2_sampling.py
din14970: When making the notebook I realized that from a user perspective it doesn't make much sense to access each sampling method from different functions with different arguments. It makes a lot more sense to have a function like `sample_s2` where a string argument specifies which type of gridding should be used. The `cube` sampling can then also be "unraveled" using `partial` to map each type of cube sampling to a string, so that each type of sampling has the same interface - just provide a resolution value. One could just define a mapping in a function and then use that instead, the advantage of a mapping is that you can also query the available keys (this would be hidden or only accessible from documentation in the case of a function). The downside is that mappings which return a callable are a kind of weird construct, and directly "calling" something that is returned from a dictionary looks weird. I have to have a think on something that makes more sense, but in my view a simple interface is having a single function to have access to all kinds of mappings.

doc/s2_sampling.ipynb
harripj: Yes the parameterisation looks similar. The function `sample_S2_random_mesh()` produces random vectors on S2, but the function `sample_S2_equal_area_mesh()` in #340 produces a grid of cells with equal area on S2. This was necessary in that PR as to produce the pole density distribution we need the chance a given vector intercepts each cell to be the same.

Here is a comparison with the UV mesh:
```python
from orix.sampling.S2_sampling import sample_S2_equal_area_mesh, sample_S2_uv_mesh
from orix import plot

fig, ax = plt.subplots(ncols=2, figsize=(8, 4), subplot_kw=dict(projection='stereographic'))

uv = sample_S2_uv_mesh(10, hemisphere='upper')
ea = sample_S2_equal_area_mesh(10, hemisphere='upper')

ax[0].scatter(uv)
ax[1].scatter(ea)

ax[0].set_title('UV mesh')
ax[1].set_title('Equal area mesh')
```
dc7eb148-d9e5-4ed4-bc28-b9bba0538b4b

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,170 @@
...
+    "## Sampling methods\n",
+    "Currently 7 sampling methods are supported, demonstrated below by plotting in stereographic projection, each with different regions of higher and lower sampling density.\n",
+    "The user should mainly decide on the `resolution` parameter, which determines the maximum angle (in degrees) between nearest neighbors except in the `random` sample where it determines the **average** angle between nearest neighbors."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4c696bea",
+   "metadata": {},
```
harripj, line 65: I noticed in the built docs that currently there is no thumbnail set for this notebook. Could you set this figure as the thumbnail? (I think the final figure may be a bit too large!)

If you can add something like the following to the `metadata` tag:
```
"metadata" : {
    "nbsphinx-thumbnail": {
         "tooltip": "Sampling directions on orientation space S2"
    },
    "tags": [
     "nbsphinx-thumbnail"
    ]
}
```

harripj (changes requested):

doc/s2_sampling.ipynb
din14970: I've now made all the different individual mesh functions private and centralized them into one function.

hakonanes:
[This documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch#updating-a-local-clone-after-a-branch-name-changes) might be useful with regards to changing branch names.

hakonanes:
> Heads up, I've add to change the pre-commit config due to a black dependency issue see https://github.com/OCA/oca-addons-repo-template/issues/133

I think this is unnecessary with black v22.3, which is the current version in `develop`? Anway, going over this PR again now.

doc/reference.rst
```diff
@@ -144,14 +144,14 @@ ang
 ~~~
 .. automodule:: orix.io.plugins.ang
     :members:
-    :undoc-members:
```
hakonanes, line 147: Could you revert these changes?

doc/s2_sampling.ipynb
hakonanes: Hehe, I meant rephrasing the marked line "... many VECTORS are symmetrically equivalent" to "... many DIRECTIONS are symmetrically equivalent", while the line "Sampling of VECTORS is most relevant..." was fine.

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +144,228 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+        _sample_S2_cube_mesh,
+        grid_type="spherified_corner",
+    ),
+    "icosahedral": _sample_S2_icosahedral_mesh,
+    "hexagonal": _sample_S2_hexagonal_mesh,
+    "random": _sample_S2_random_mesh,
+}
+sampling_methods: Sequence[str] = list(_sampling_method_registry.keys())
+
+_s2_sampling_docstring = (
```
hakonanes, line 338: Why not add this directly to `sample_S2()`?

hakonanes: Nevermind, didn't see the use of `format()`.

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +144,228 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+    """Generate unit vectors that sample S2 with a specific angular resolution
```
hakonanes, line 339:
```suggestion
    """Generate unit vectors that sample S2 with a specific angular resolution.
```

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +144,228 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+
+    Parameters
+    ----------
+    resolution
+        Maximum angle between nearest neighbour grid points, in degrees.
+    method
+        Sphere meshing method. Options are: {}.
+
```
hakonanes, line 347:
```suggestion
    kwargs
        Keyword arguments passed to the sampling function.
```

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +144,228 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+    n = int(np.ceil(a / (r_i * np.tan(np.deg2rad(resolution)))))
+    vertices = _compose_from_faces(corners, faces, n)
+    return Vector3d(vertices).unit
+
+
+_sampling_method_registry: Mapping[str, Callable] = {
+    "uv": _sample_S2_uv_mesh,
+    "normalized_cube": partial(
+        _sample_S2_cube_mesh,
+        grid_type="normalized",
```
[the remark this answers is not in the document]
hakonanes: This is also the case for all function signatures above.

orix/sampling/__init__.py
```diff
@@ -20,14 +20,17 @@
 
 from orix.sampling.sample_generators import get_sample_fundamental, get_sample_local
 from orix.sampling.SO3_sampling import uniform_SO3_sample
-from orix.sampling.S2_sampling import sample_S2_cube_mesh, sample_S2_uv_mesh
+from orix.sampling.S2_sampling import (
+    sample_S2,
+    sampling_methods as sampling_methods_S2,
```
hakonanes, line 25: How about `sample_s2_methods`, so we have "sample_S2" for both?

doc/s2_sampling.ipynb
hakonanes: I think this is a good idea.

orix/sampling/S2_sampling.py
hakonanes: I agree with you and I think you have found a very clean and neat solution. Linking to the docstrings of the individual sampling functions could be good for the interested user to know more.

hakonanes:
> Basically I've made all the individual sampling methods private

This makes the individual docstrings unreachable, and I think they should be part of the public API since they explain the sampling functions. What do you think about making them public, but basically only using `sample_S2` in the user guide, and also stating that this is the function meant to be used in the individual docstrings?

orix/sampling/S2_sampling.py
```diff
@@ -131,3 +144,228 @@ def sample_S2_cube_mesh(resolution, grid_type="spherified_corner"):
...
+        Maximum angle between nearest neighbour grid points, in degrees.
+    method
+        Sphere meshing method. Options are: {}.
+
+    Returns
+    -------
+    vectors
+        Vectors that sample the unit sphere.
+    """
+).format(sampling_methods)
```
hakonanes, line 353: I suggest to make the private functions public and link to them in `sample_S2` like so:

```python
_sampling_method_names = []
for method in list(_sampling_method_registry.values()):
    if isinstance(method, partial):
        method = method.func
    name = method.__name__
    if name not in _sampling_method_names:
        _sampling_method_names.append(name)

_s2_sampling_docstring = (
    """
Generate unit vectors that sample S2 with a specific angular
resolution.

Parameters
----------
resolution
    Maximum angle between nearest neighbour grid points, in degrees.
method
    Sphere meshing method. Options are: {}.

Returns
-------
vectors
    Vectors that sample the unit sphere.

See Also
--------
{}
"""
).format(sampling_methods, "\n".join(_sampling_method_names))
```

orix/sampling/S2_sampling.py
hakonanes: Also, the default method must be mentioned in the `method` parameter explanation.

doc/reference.rst
din14970: don't know what happened here, very strange.

orix/sampling/S2_sampling.py
din14970: > This will only be relevant for the random seed, as far as I can see from the individual docstrings.

Currently yes, but it might be that there are sampling methods that would require more fine-grained control or additional parameters passed to it, which is why I allow the option. I'll mention that currently this is only relevant for the `seed`.

> Also, the default method must be mentioned in the method parameter explanation.

What is your reasoning for this? Personally I disagree since the default defined in the function definition automatically shows up in the docs; otherwise it is repeated in two places. I would agree if the default is something like `None` which then gets interpreted to something else inside the function.

hakonanes: One use for repeating the default in the parameter explanation is that when you have a long docstring with many parameters and read the description of one of them, you don't have to scroll to the function signature at the top to learn what the default is. Also, I think most users expect it from using other Python packages, see e.g. the `mode` parameter in [numpy.choose()](https://numpy.org/doc/stable/reference/generated/numpy.choose.html), the `histtype` parameter in [matplotlib.pyplot.hist()](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hist.html#matplotlib.pyplot.hist), or the `mode` parameter in [skimage.filters.difference_of_gaussians()](https://scikit-image.org/docs/stable/api/skimage.filters.html#skimage.filters.difference_of_gaussians).

hakonanes: > but it might be that there are sampling methods that would require more fine-grained control or additional parameters passed to it

Yes, this is a good reason for putting the keyword arguments in the docstring.

din14970: Ok. I think this is more or less resolved now.

hakonanes: I think it is too. Thank you!

hakonanes:
I'm very happy with this PR now. Is it OK if I make some minor changes to the functions references and add the additions to the changelog before approving?

din14970:
@hakonanes yep, feel free!

hakonanes (approved):
I suggest to merge!

CHANGELOG.rst
```diff
@@ -14,6 +14,8 @@ Added
 -----
 - Support for type hints has been introduced and a section on this topic has been added
   to the contributing guide.
+- Seven methods for sampling unit vectors in spherical grids in *S2* via
```
harripj, line 17: I'm a bit confused by the nomenclature here, what is the meaning of "spherical grid"? Also I think this should be `on *S2*`, as S2 is 2d, but happy to be convinced otherwise.

orix/sampling/_polyhedral_sampling.py
```diff
@@ -0,0 +1,223 @@
...
+        this tree is constructed depends on leaf_size, so this parameter
+        will influence speed of tree construction and search.
+
+    Returns
+    -------
+    nn1_vec
+        Point cloud represented by an array of shape (N, D) with N points in
+        D dimensions, representing the nearest neighbor point of each point
+        in "points"
+    """
```
harripj, line 195:
```suggestion
    """Get array of first nearest neighbor points for all points in a point cloud

    Parameters
    ----------
    points
        Point cloud with shape (N, D) representing N points in D dimensions.
    leaf_size
        The NN search is performed using a cKDTree object. The way
        this tree is constructed depends on leaf_size, so this parameter
        will influence speed of tree construction and search.

    Returns
    -------
    nn1_vec
        Point cloud represented by an array of shape (N, D) with N points in
        D dimensions, representing the nearest neighbor point of each point
        in `points`.
    """
```

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,181 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a59fef6b",
+   "metadata": {},
+   "source": [
+    "# Sampling unit vectors\n",
+    "\n",
+    "Sampling unit vectors or directions in 3D space is equivalent to sampling points from the surface of the unit sphere ($S_2$).\n",
```
harripj, line 20:
```suggestion
    "Sampling unit vectors or directions in 3D space is equivalent to sampling points on the surface of the unit sphere ($S_2$).\n",
```

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,181 @@
...
+   "source": [
+    "## Sampling the fundamental zone\n",
+    "Sampling of directions is most relevant when combined with crystal symmetry functionality.\n",
+    "In this case, we would want to sample directions in the real crystallographic basis.\n",
+    "Usually one only wants to sample the fundamental zone, not the entire sphere, as in the context of a crystal, many directions are symmetrically equivalent.\n",
+    "The fundamental zone of a point group represents the smallest region of symmetrically unique directions.\n",
+    "The higher the symmetry of the point group, the smaller the fundamental zone.\n",
+    "\n",
+    "The symmetry operations are independent of the $S_2$ sampling methods, so any set of vectors can be limited to the fundamental zone.\n",
+    "However, different meshes may be most appropriate for different point groups, as we illustrate below.\n",
```
harripj, line 100:
```suggestion
    "However, different meshes may be more appropriate for different point groups, as we illustrate below.\n",
```

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,181 @@
...
+    "In general, to sample vectors for crystals with $\\alpha=\\beta=\\gamma=90^\\circ$, one of the `cube` meshes is recommended.\n",
+    "For trigonal or hexagonal crystals that have a 3 or 6-fold axis along [001], a hexagonal mesh is recommended.\n",
+    "This ensures that the boundary of the fundamental zone is also sampled at regular intervals, otherwise some of these edges may appear \"ragged\".\n",
+    "For lower symmetry crystals, most samplings will be fine, and the icosahedral sampling may be recommended as points are most evenly spread.\n",
```
harripj, line 104:
```suggestion
    "For lower symmetry crystals, most sampling methods are suitable, and icosahedral sampling is recommended as the points are most evenly spaced.\n",
```

doc/s2_sampling.ipynb
```diff
@@ -0,0 +1,181 @@
...
+    "\n",
+    "Below we demonstrate a sampling of the fundamental zones for all 32 point groups."
```
harripj, line 106:
```suggestion
    "Below we demonstrate sampling of the fundamental zone for all 32 point groups."
```

harripj (changes requested):
Looks great to me! Just a few very small suggestions and a question.

harripj (approved):
Looks great! Thanks @din14970 for your efforts porting this over. LGTM.

Issue #320, opened 2022-04-15
Calculate, export and import the Polefigure (PDF) or the Inverse Polefigure (IPDF) 


tarekiraki:
Dear Sir or Madam, 

I am quite new to the orix framework. I've been analyzing orientations with mtex so far. I would like to process orientations either with  Polefigure (PDF) or Inverse Polefigure (IPDF) and test if there is a 'orix-way' to accomplish this.

So I would like to ask, if there is a way with orix to (1) calculate an ODF, (2) calculate the PDF/IPDF (3) export the  PDF/IPDF and (4) import the  PDF/IPDF again. The source code in mtex for this pipeline is listed below.

Thanks for help.
Cheers Tarek

```
% calculate ODF
ebsd = EBSD.load(
	filename_ori_data, CS, 'interface', 'generic', 'ColumnNames', 
	{'x' 'y' 'phi1' 'Phi' 'phi2' 'weights'}, 'Bunge');
oris = ebsd.orientations;
odf = calcDensity(oris, 'halfwidth', 3*degree, 'weights', ebsd.weights);

% calculate Pole Figure
pf_odf = calcPoleFigure(odf, h, 'upper');

% export Pole Figure
export(pf_odf, folder_pf, 3*degree)

% import Pole Figure
pf_import = loadPoleFigure(
	fnames, cs, h, 'interface','generic', 'ColumnNames', 
	{'Polar Angle' 'Azimuth Angle' 'Intensity'}, 'RADIANS');

% calulate ODF based on Pole Figure data
odf_recon = calcODF(pf_import,'silent');

% export odf_recon
export(odf_recon, filename_odf_export)
```

hakonanes:
Hi Tarek,

unfortuantely no, orix cannot compute an orientation distribution function. So far, we only have the possibility to compute discrete pole figures "by hand", for example something like this should compute a <111> pole figure assuming point group symmetry m-3m 

```python
import numpy as np
from orix.crystal_map import Phase
from orix.quaternion Orientation, symmetry
from orix.vector import Miller

# Create orientations
x, y, phi1, Phi, phi2 = np.loadtxt("file.txt", unpack=True)
oris = Orientation.from_euler(np.deg2rad([phi1, Phi, phi2]), symmetry.Oh)

# Create direct lattice vectors
uvw111 = Miller(uvw=[1, 1, 1], phase=Phase(point_group=symmetry.Oh))
[...]

Issue #338, opened 2022-05-16
Introducing Type Hints to orix


harripj:
As discussed in https://github.com/pyxem/orix/pull/334#discussion_r873158671 with @hakonanes and @din14970 we should gradually introduce type hints into orix. This opens up the possibility to perform type checking at during the development phase and assures that data types are well-defined in the documentation.

Apart from #334 the code base currently uses docstring types, whereas @hakonanes suggested that we migrate towards defining types in the function definitions rather than in the docstrings and I would be happy with this change (see comment above), this would mean that type hinting is used and types are removed from the doctring to avoid type definition duplication.

As this may take some time, I suggest that we aim to convert the code base gradually with completion due in 1.0. If everyone is in agreement then new PRs should use type hinting.

pc494:
I had been the hold out on this for some time, but am happy for this to go ahead.

harripj:
Copying this [comment](https://github.com/pyxem/orix/pull/340#issuecomment-1128775422) here for clarity in the discussion:

> Related to https://github.com/pyxem/orix/issues/338, __annotations__ within the future package was only introduced in Python 3.7, hence the CI Python 3.6 build fails.
>
> __annotations__ are needed for type hints when a class operates with another instance of the class, for example: Vector3d.outer(other: Vector3d). There is likely a lot of this in the code base, so introducing type hints while testing with Python 3.6 will cause problems.

pc494:
Given that Python 3.6 is deprecated we should probably stop testing for it?

harripj:
> Given that Python 3.6 is deprecated we should probably stop testing for it?

This could be an option. Python 3.7 is currently supported for ~another year. https://www.python.org/downloads/

hakonanes:
Glad that you brought this up @harripj. It seems to me like the course of action is to:
1. Make a PR dropping support for Python 3.6
2. Continue with type hints including `__annotations__`

harripj:
Sounds like a good plan to me!

hakonanes:
`__annotations__` can now be used in PRs to `develop`. Note that Python 3.6 is still supported for orix 0.9.x, so any patches have to respect this.

hakonanes:
Relevant comment on type *checking* by @harripj: https://github.com/pyxem/orix/pull/340#discussion_r883537419

hakonanes:
Completed in #378.

Issue #265, opened 2021-12-22
(Inverse) pole figure convenience method(s)


hakonanes:
We should have some convenience methods for plotting pole figures (PFs) and inverse pole figures (IPFs). The following are on my todo-list, but any help is appreciated if people find the time.

## IPF
At the moment, IPFs are plotted via `Orientation.scatter("ipf", direction=Vector3d((0, 0, 1)))` or via `orix.plot.InversePoleFigurePlot.scatter()`, which accepts `Vector3d` (of which `Miller` is a subclass), seen in [this user guide](https://orix.readthedocs.io/en/stable/clustering_orientations.html#Visualisation). Should we add "ipf" as an option to `Miller`? I think so.
- [ ] `Miller.scatter("ipf")` option

## PF
At the moment, we don't have any "official" pole figure plotting, however, almost all the functionality is there. The current stereographic plots are pole figures when fed sample directions r parallel to some crystal direction h, given by r = ~ori * h. We need to support outer multiplication of `Orientation` with `Vector3d` to enable multiplication of multiple orientations with a symmetrised `Miller` family, say <111> in a cubic crystal. I think we should add an `Orientation.scatter("pf", direction=Miller)` option as well.
- [ ] Multiplication of `Orientation` with `Vector3d` of different shapes
- [ ] `Orientation.scatter("pf", direction=Miller)` option

harripj:
> At the moment, IPFs are plotted via Orientation.scatter("ipf", direction=Vector3d((0, 0, 1))) or via orix.plot.InversePoleFigurePlot.scatter(), which accepts Vector3d (of which Miller is a subclass), seen in this user guide. Should we add "ipf" as an option to Miller? I think so.

I think that this may be confusing. For IPF the sample axes are shown in the crystal reference frame, and whilst `Miller` is just a subclass of `Vector3d` in orix, my understanding is that `Miller` should be reserved for crystal directions? Perhaps there is an interpretation I am missing?

By the same logic `Miller` should definitely be included for PF.

hakonanes:
> I think that this may be confusing. For IPF the sample axes are shown in the crystal reference frame, and whilst `Miller` is just a subclass of `Vector3d` in orix, my understanding is that `Miller` should be reserved for crystal directions? Perhaps there is an interpretation I am missing?

[...]

PR #340, opened 2022-05-17
Add Pole Density Function (PDF) functionality to `Vector3d`

harripj:
orix currently supports plotting poles (vectors) in the stereographic projection, however another useful crystallographic tool is the Pole Density Function (PDF). This representation allows the density of poles to be compared, which can be difficult when studying (overlapping) point data. This functionality was also requested in #320.

The PDF is calculated in angular space on S2 and uses (added) equal area sampling. This follows the work done by [Rohrer et al.](http://www.hanser-elibrary.com/doi/abs/10.3139/146.017934) which is also added to the bibliography. The resulting PDF can be plotted on both hemispheres in units of Multiples of Random Density (MRD).

I have also added lazy `Vector3d-Vector3d` outer product, as I needed this at some point when working on this PR.

#### Progress of the PR

Fot the `Ti_orientation` data from `orix.data`:
```python
from orix.vector import Miller, Vector3d
from orix.quaternion.symmetry import D6

from orix.data import ti_orientations

ori = ti_orientations(allow_download=True)[::3]
m = Miller(uvw=(1, 0, 0), phase=Phase(point_group=D6)).symmetrise().unique()

poles_100 = (~ori).outer(m).flatten()[::10]
poles_100.scatter(s=2, axes_labels=['x', 'y', None], alpha=0.005, hemisphere='both')
```
4ccfe8e4-bad6-47cd-bae9-60092e773109

```python
poles_100.pdf(return_figure=True, colorbar=True, sigma=5, log=True, cmap='magma', resolution=5, hemisphere='both')
```
bea9efd7-229a-4cc1-be18-76a76d049c89

And for random data:
```python
poles = Vector3d(np.random.randn(100_000, 3)).unit
poles.scatter(s=2, axes_labels=['x', 'y', None], alpha=0.01, hemisphere='both')
```

7a9c590c-f07b-4659-a788-1638c94a7c70

```python
poles.pdf(return_figure=True, colorbar=True, sigma=5, log=False, cmap='magma', resolution=5, hemisphere='both')
```
045c82b7-bc42-41b7-8505-3e48ea13245d

In this last example you can see that the PDF is ~1 MRD, which is expected for random data.

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


harripj:
I wasn't able to check the failing [test](https://github.com/pyxem/orix/runs/6469869427?check_suite_focus=true#step:8:86) on my machine, even after cleaning my environment and using the same matplotlib version. I have removed these minor checks.

The test checked that there was only one `collection` on each axes. On my machine this was correct, but perhaps I have an environment problem somewhere.

harripj:
Related to #338, `__annotations__` within the `future` package was only introduced in Python 3.7, hence the CI Python 3.6 build fails.

`__annotations__` are needed for type hints when a class operates with another instance of the class, for example: `Vector3d.outer(other: Vector3d)`. There is likely a lot of this in the code base, so introducing type hints while testing with Python 3.6 will cause problems.

harripj:
I can't replicate the [failing test](https://github.com/pyxem/orix/runs/6470207686?check_suite_focus=true#step:9:151) on my machine. If anybody else would have the time to pull this PR onto their machine, run the testing suite, and let me know the result I would be grateful!

orix/sampling/S2_sampling.py
```diff
@@ -18,12 +18,125 @@
...
+    offset : float, optional
+        Mesh points are offset in angular space by this fraction of the
+        step size, must be in the range [0..1]. Default is 0.
+    azimuthal_endpoint : bool, optional
+        If True then endpoint of the azimuthal array is included in the
+        calculation. Default is False.
+    
+    Returns
+    -------
+    azimuth, polar : ndarray
```
hakonanes, line 86: numpy.ndarray

orix/sampling/S2_sampling.py
```diff
@@ -18,12 +18,125 @@
...
+    """
+    hemisphere = hemisphere.lower()
+    if hemisphere not in ("upper", "lower", "both"):
+        raise ValueError('Hemisphere must be one of "upper", "lower", or "both".')
+    if offset < 0 or offset > 1:
```
hakonanes, line 91: I find `not 0 < offset < 1` more readable.

orix/sampling/S2_sampling.py
hakonanes: This is required other places as well.

orix/sampling/S2_sampling.py
```diff
@@ -41,28 +154,108 @@ def sample_S2_uv_mesh(resolution):
...
+        calculation. Default is False.
+
+    Returns
+    -------
+    azimuth, polar : ndarray
+    """
+    hemisphere = hemisphere.lower()
+    if hemisphere not in ("upper", "lower", "both"):
+        raise ValueError('Hemisphere must be one of "upper", "lower", or "both".')
+    # calculate number of steps and step size angular spacing, = D [1]
```
hakonanes, line 204: What does "= D [1]" refer to here?

orix/sampling/S2_sampling.py
```diff
@@ -41,28 +154,108 @@ def sample_S2_uv_mesh(resolution):
...
+    """Vectors of a cube mesh on a unit sphere *S2* according to equal
+    area spacing.
+
+    Parameters
+    ----------
+    resolution : float
+        The angular resolution in degrees of the azimuthal vectors.
+    hemisphere : str, optional
+        Generate mesh points on either the "upper", "lower" or "both"
+        hemispheres. Default is "both".
```
hakonanes, line 192: Either the docstring or the default value must be updated here.

orix/sampling/S2_sampling.py
```diff
@@ -41,28 +154,108 @@ def sample_S2_uv_mesh(resolution):
...
+    polar = np.arccos(polar)
+
+    return azimuth, polar
+
+
+def sample_S2_equal_area_mesh(
+    resolution: float, hemisphere: str = "both", remove_pole_duplicates: bool = True
+) -> Vector3d:
+    """Vectors of a cube mesh on a unit sphere *S2* according to equal
+    area spacing. Parameterization taken from :cite:`rohrer2004interfaces`.
```
hakonanes, line 234: I suggest to just write "area spacing :cite:\`rohrer2004interfaces\`.

orix/sampling/S2_sampling.py
```diff
@@ -41,28 +154,108 @@ def sample_S2_uv_mesh(resolution):
...
+        polar_max,
+        num=steps,
+        endpoint=True,
+    )
+    polar = np.arccos(polar)
+
+    return azimuth, polar
+
+
+def sample_S2_equal_area_mesh(
```
hakonanes, line 230: Am I right in thinking that this will then be an alternative to the sampling options to be introduced by @din14970?

orix/tests/sampling/test_s2_sampling.py
```diff
@@ -58,3 +60,44 @@ def test_cube_mesh(self, grid_type, resolution, size):
...
+    )
+    def test_equal_area_angle_coordinate_arrays(self, hemisphere, min_polar, max_polar):
+        azi, polar = S2_sampling._sample_S2_equal_area_arrays(
+            10, hemisphere=hemisphere, azimuthal_endpoint=False
+        )
+        assert isinstance(azi, np.ndarray)
+        assert isinstance(polar, np.ndarray)
+        assert azi.ndim == 1
+        assert polar.ndim == 1
+        assert np.allclose(polar.min(), min_polar)
```
hakonanes, line 76: I suggest to use `==` to compare numbers here.

orix/tests/sampling/test_s2_sampling.py
```diff
@@ -58,3 +60,44 @@ def test_cube_mesh(self, grid_type, resolution, size):
     def test_cube_mesh_raises(self):
         with pytest.raises(ValueError, match="The `grid_type` hexagonal"):
             _ = sampling.sample_S2_cube_mesh(2, "hexagonal")
+
+    @pytest.mark.parametrize(
+        "hemisphere, min_polar, max_polar",
+        [("upper", 0, np.pi / 2), ("lower", np.pi / 2, np.pi), ("both", 0, np.pi)],
+    )
+    def test_equal_area_angle_coordinate_arrays(self, hemisphere, min_polar, max_polar):
```
hakonanes, line 68: I find "equal_area_angle" a bit confusing, could you just write "equal_area_coordinate_arrays"?

orix/tests/test_vector3d.py
```diff
@@ -436,6 +443,44 @@ def test_rmul_notimplemented(self, vector):
         "cantmul" * vector
 
 
+def test_pdf():
```
hakonanes, line 446: I suggest to add a new class for testing the PDF funtionality, so that potential future tests will be added in the correct place, and it will be easier to just run the PDF tests by running `pytest -k TestPDF`.

orix/tests/test_vector3d.py
```diff
@@ -436,6 +443,44 @@ def test_rmul_notimplemented(self, vector):
...
+    v = Vector3d(np.random.randn(10_000, 3)).unit
+    fig1 = v.pdf(return_figure=True, colorbar=True, sigma=5, log=True)
+    assert len(fig1.axes) == 2
```
hakonanes, line 449: Some blank lines separating parts would be good here.

orix/vector/vector3d.py
```diff
@@ -16,10 +16,16 @@
 # You should have received a copy of the GNU General Public License
 # along with orix.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
```
hakonanes, line 20: Remove this blank line here, as both are imports from the standard library.

orix/vector/vector3d.py
```diff
@@ -584,6 +633,171 @@ def get_circle(self, opening_angle=np.pi / 2, steps=100):
...
+        Notes
+        -----
+        The PDF is calculated in terms of Multiples of Random
+        Distribution (MRD), ie. multiples of the expected density if the
+        pole distribution was completely random, see reference [1].
+
+        References
+        ----------
+        [1]: The Distribution of Internal Interfaces in Polycrystals,
```
hakonanes, line 718: I suggest to add :cite:\`rohrer2004distribution\` at the end of the docstring's first sentence instead.

doc/bibliography.bib
```diff
@@ -118,3 +118,14 @@ @article{singh2016orientation
 	volume = {24},
 	year = {2016}
 }
+@article{rohrer2004interfaces,
+	author = {Rohrer, Gregory S. and Saylor, David M. and Dasher, Bassem El and Adams, Brent L. and Rollett, Anthony D. and Wynblatt, Paul},
+	doi = {10.3139/146.017934},
```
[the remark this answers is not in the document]
hakonanes: Also, the bibtex key should be "rohrer2004distribution" to be in line with the other entries in this file.

orix/vector/vector3d.py
```diff
@@ -584,6 +633,171 @@ def get_circle(self, opening_angle=np.pi / 2, steps=100):
...
+            x, y = x.reshape(v_mesh.shape), y.reshape(v_mesh.shape)
+            # plot mesh
+            kwargs.setdefault("cmap", "magma")
+            pc = ax.pcolormesh(x, y, hist, **kwargs)
+            if colorbar:
+                label = "MRD"
+                if log:
+                    label = f"log({label})"
+                plt.colorbar(pc, label=label, ax=ax)
+        if return_figure:
```
hakonanes, line 798: Blank line here.

orix/vector/vector3d.py
```diff
@@ -584,6 +633,171 @@ def get_circle(self, opening_angle=np.pi / 2, steps=100):
             circles[i] = v.rotate(v.perpendicular, oa).rotate(v, full_circle)
         return circles
 
+    def pdf(
```
hakonanes, line 636: I suggest to rename this to `pole_density_function()`: Yes, it takes longer to type, but I think PDF is a less familiar abbreviation than IPF, which is the only used abbreviation in orix currently. I'm open to other views.

hakonanes: And it cannot be confused with a probability density function or other things abbreviated PDF.

hakonanes (changes requested):
As far as I can tell, this is the first addition of *quantification* of orientations/vectors to orix, and thus a very important step forward!

Looking at the example you included in the top comment, I think we should reconsider #265 and whether we should add something like the following to generate the first figure in your example

```python
from orix import data


g = data.ti_orientations(allow_download=True)[::3]
m = Miller(uvw=(1, 0, 0), phase=Phase(point_group=symmetry.D6))
g.scatter("pf", direction=m, hemisphere="both")  # Pole figure
```

Other things:
 - [x] Changelog entry
 - [ ] Update API reference
 - [ ] User guide notebook(s), either a new one or adding to existing ones

orix/sampling/S2_sampling.py
```diff
@@ -18,12 +18,125 @@
 
 """Generation of spherical grids in *S2*."""
 
+from typing import Tuple
+
 import numpy as np
+from numpy.typing import NDArray
```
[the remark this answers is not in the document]
harripj: Good point, I will convert to explicit `np.ndarray`.

orix/sampling/S2_sampling.py
harripj: From the cited Rohrer paper, but I agree it is not clear. I will update this

doc/bibliography.bib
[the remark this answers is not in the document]
harripj: > Visiting https://doi.org/10.3139/146.017934 gets me nowhere, I suggest to add the URL instead.

Yes this is an interesting observation- it goes against what I thought the purpose of a DOI was! I have changed it for the (working) URL in any case.

orix/sampling/S2_sampling.py
harripj: I have updated the `hemisphere` argument to have default value of `None`. In this case it is converted to `"both"` internally, but is more consistent with other API. Is this what you were thinking?

harripj:
> I can't replicate the [failing test](https://github.com/pyxem/orix/runs/6470207686?check_suite_focus=true#step:9:151) on my machine. If anybody else would have the time to pull this PR onto their machine, run the testing suite, and let me know the result I would be grateful!

Still not quite sure why there was a difference between running the same code in the same environment with `pytest` vs `ipython`. In any case the problem was that there were multiple collections in the axes when testing using `pytest` and on the CI. In any case the `--pdb` flag for `pytest` comes in handy!

harripj:
The failing build should be fixed after #347 is merged.

hakonanes:
Thank you for introducing `Orientation.plot()`, but I suggest to leave the (I)PF convenience methods for another PR. These methods will be very much used, so we should carefully consider the public facing functions and methods. I anticipate some discussion on the naming and parameter choices. We can also then focus on where the existing documentation can benefit from the added functionality.

hakonanes:
> we should carefully consider the public facing functions and methods

By this I mean it will be difficult to change later on since they will be much used.

harripj:
No problem, this is fine with me- I had thought your previous comment was a suggestion but I agree it is best to leave this for a later PR. I will remove these changes and we can consider this later.

hakonanes:
> I had thought your previous comment was a suggestion but I agree it is best to leave this for a later PR

Sorry, my bad, I should have stated that I ment in another PR...

harripj:
No problem! There is a commit [here](https://github.com/pyxem/orix/pull/340/commits/4f5888bce2f8fc3fc8c78a5a1d43593eeb6f73b5) that could potentially be used as a starting point for that discussion in another issue!

orix/sampling/S2_sampling.py
harripj: This has been implemented.

hakonanes:
I don't think the notebook was added to git? We have to add the `-f` flag when we `git add` notebooks (see first point in the list in the [contributing guide](https://orix--340.org.readthedocs.build/en/340/contributing.html#build-and-write-documentation)).

orix/sampling/S2_sampling.py
hakonanes: Should we add as an alternative method in `orix.sampling.sample_S2` which will be introduced in #334?

harripj:
> I don't think the notebook was added to git? We have to add the -f flag when we git add notebooks (see first point in the list in the [contributing guide](https://orix--340.org.readthedocs.build/en/340/contributing.html#build-and-write-documentation)).

Well spotted yep! I saw a change to another notebook and assumed that was the one! Newly added now.

orix/sampling/S2_sampling.py
harripj: Yes, I think it is easiest if I reformat the sampling method here to fit in with the syntax in #334 after that PR gets merged.

hakonanes: Good idea, let's wait for that PR to be merged first (very close).

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+   },
+   "source": [
+    "This notebook is part of the *orix* documentation https://orix.readthedocs.io. Links to the documentation won’t work from the notebook."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Visualizing Crystal Poles in the Pole Density Function (PDF)\n",
```
hakonanes, line 16: I suggest to remove "(PDF)" so that the title fits on two lines in the sidebar, and rather introduce the abbreviation in the introductory cell.

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "\n",
+    "This notebook demonstrates how to quantify the distribution of crystallographic poles,\n",
+    "which is useful, for example, in texture analysis.\n",
+    "\n",
+    "First, we load some sample orientations from a Titanium sample dataset which represent\n",
+    "crystal orientations in the sample reference frame:\n",
+    "\n",
+    "NB. if not previously downloaded this will download some example data from an online\n",
```
hakonanes, line 24: Might I suggest to use an info box (admonition) here instead? nbsphinx supports [warning and info boxes](https://nbsphinx.readthedocs.io/en/0.8.8/markdown-cells.html#Info/Warning-Boxes):

```markdown
<div class="alert alert-info">

Note

If not previously downloaded this will download some example data from an online
repository to a local cache, see the docstring of
[ti_orientations](reference.rst#orix.data.ti_orientations) for more details.

</div>
```

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "import numpy as np\n",
+    "\n",
+    "from orix.crystal_map import Phase\n",
+    "from orix.data import ti_orientations\n",
+    "from orix import plot\n",
+    "from orix.quaternion import symmetry\n",
+    "from orix.sampling.S2_sampling import sample_S2_equal_area_mesh\n",
+    "from orix.vector import Miller, Vector3d\n",
+    "\n",
+    "ori = ti_orientations(allow_download=True)\n",
```
hakonanes, line 47: I always load data in the cell *below* imports, since I often add to the imports as I build up a notebook and I don't want to reload the data every time I do this. I suggest to load the data in a new cell below.

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "ori.shape"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's look at the sample's $\\{011\\}$ texture plotted in the stereographic projection.\n",
```
hakonanes, line 55: Perhaps we should use {hkil} = {01-11} since the orientations describe unit cells of hexagonal symmetry? This is supported in the [Miller class](https://orix.readthedocs.io/en/stable/crystal_directions.html#Trigonal-and-hexagonal-index-conventions).

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "unique $(011)$ poles: "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m = Miller(hkl=(0, 1, 1), phase=Phase(point_group=symmetry.D6))\n",
```
hakonanes, line 67: By using `ori.symmetry` here you can drop the import of `symmetry`. To show the user what `ori.symmetry` contains, we can print the `ori` representation after loading it above: this will show the size and symmetry in one line.

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "m = m.symmetrise().unique()\n",
```
hakonanes, line 68: I suggest to pass `symmetrise(unique=True)` instead of the extra method call (see [docstring](https://orix.readthedocs.io/en/stable/reference.html#orix.vector.Miller.symmetrise)).

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig = poles.scatter(hemisphere='both', s=2, alpha=0.05, return_figure=True)\n",
+    "\n",
+    "w, h = plt.rcParams['figure.figsize']\n",
```
hakonanes, line 109: Is it possible to set some Matplotlib parameters in the import cell to avoid some of these cosmetic calls, e.g. a default figure size suitable for this plot? This makes following the notebook simpler. We should work towards making the convenience methods handle as much as possible of this in the future, without loosing the existing customizability...

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this case there are many individual data points, which makes it difficult to\n",
+    "interpret whether regions contain higher or lower pole density.\n",
+    "\n",
+    "In this case we can use the [Vector3d.pole_density_function()](reference.rst#orix.vector.Vector3d.pole_density_function)\n",
```
hakonanes, line 121: This method must be listed in the [autosummary](https://github.com/pyxem/orix/blob/develop/doc/reference.rst#vector3d) of `Vector3d`.

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "projection with a resolution of 10&deg;:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, ax = plt.subplots(ncols=2, figsize=(2 * h, h), subplot_kw=dict(projection='stereographic'))\n",
```
hakonanes: I suggest to split this into three lines so that we don't have to scroll horizontally to see the end.

hakonanes: You should get the same figure by just doing `v_mesh.scatter(hemisphere="both", axes_labels=["x", "y"])`, right? Possibly return the figure and call `tight_layout()` on a second line.

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "    ax[i].set_labels('x', 'y', None)\n",
+    "    ax[i].set_title(str(n))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can also change the sampling angular `resolution` on $S_2$, the colormap with the\n",
+    "`cmap` parameter, and broadening of the density distrbution with `sigma`:"
```
hakonanes, line 194: distribution

hakonanes (changes requested):
I've gone over the notebook and have some suggestions.

doc/pole_density_function.ipynb
harripj: Good idea.

doc/pole_density_function.ipynb
harripj: Missed that- thanks!

doc/pole_density_function.ipynb
harripj: Yes, you're right. In fact I have updated this figure as I found it a bit redundant to show the sampling on both hemispheres. This has been updated with a 2-axes plot to show the the sampling in the stereographic projection and in 3D.

doc/pole_density_function.ipynb
harripj: Good idea. I remembered that there was actually a `figure_kwargs` parameter for this purpose too.

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,377 @@
...
+    "from orix.data import ti_orientations\n",
+    "from orix.sampling.S2_sampling import sample_S2_equal_area_mesh\n",
+    "from orix.vector import Miller, Vector3d\n",
+    "\n",
+    "# We'll want our plots to look a bit larger than the default size\n",
+    "plt.rcParams.update(\n",
+    "    {\n",
+    "        \"figure.figsize\": (10, 5),\n",
+    "        \"lines.markersize\": 2,\n",
+    "        \"font.size\": 20,\n",
```
hakonanes, line 44: Is a font size of 15 enough? Text overlap with figure with 20.

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,377 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's look at the sample's $\\{01\\overline{1}1\\}$ texture plotted in the stereographic projection.\n",
+    "\n",
+    "First we must define the crystal's point group and generate the set of symmetrically\n",
+    "unique $(01\\overline{1}1)$ poles: "
```
hakonanes, line 87: I think \bar{} is what you want, otherwise the negative index is lower than the others.

orix/plot/stereographic_plot.py
```diff
@@ -167,6 +169,91 @@ def plot(self, *args, **kwargs):
 
         super().plot(x, y, **updated_kwargs)
 
+    def pole_density_function(
```
hakonanes, line 172: This must be listed in the API reference and in the autosummary of StereographicPlot. Why don’t you accept azimuth and polar as keyword arguments here?

orix/plot/stereographic_plot.py
```diff
@@ -167,6 +169,91 @@ def plot(self, *args, **kwargs):
...
+        log: bool = False,
+        colorbar: bool = True,
+        **kwargs,
+    ):
+        """Compute the Pole Density Function (PDF) of vectors in the
+        stereographic projection.
+
+        Parameters
+        ----------
+        args
```
hakonanes, line 186: Should be stated that these are in radians.

orix/plot/stereographic_plot.py
```diff
@@ -167,6 +169,91 @@ def plot(self, *args, **kwargs):
...
+            azimuth, polar, bins=(azimuth_arr, polar_arr), density=False
+        )
+        # Normalize by the average number of counts per cell on the
+        # unit sphere to calculate in terms of Multiples of Random
+        # Distribution (MRD). See :cite:`rohrer2004distribution`.
+        hist = hist / hist.mean()
+
+        # apply gaussian filtering to plot
+        # "wrap" along azimuthal axis, "reflect" along polar axis
+        hist = gaussian_filter(hist, sigma / resolution, mode=("wrap", "reflect"))
```
hakonanes, line 235: Is it useful to the user to have access to this histogram outside of this function? Say to compare the mean or max for different distributions, or to obtain the location of the max or n max? If yes, we should consider, probably in another PR, some separate module for this I think. orix.measure is an option, but might be too broad of a name…

hakonanes (commented):
Two minor comments on aesthetics and some questions for the PDF calculation function.

orix/plot/stereographic_plot.py
harripj: I was following the syntax of the other `StereographicPlot` methods, such as `scatter`, but you are right, there is no need as here we don't use the internal methods. I will replace with `azimuth` and `polar`.

orix/plot/stereographic_plot.py
harripj: Thanks!

orix/plot/stereographic_plot.py
harripj: Yes, in fact this is accessible through `fig.axes[0].collections` and `get_array()`, but admittedly not that obvious. Perhaps your suggestion is the best and that this calculation becomes part of another module in the future, which returns the histogram and bin coordinates. This function would then simply call that function.

harripj: I think `orix.measure` is a clear name, any quantification could go in there. Familiar to those who use `skimage` too.

doc/pole_density_function.ipynb
harripj: Good spot, will update

doc/pole_density_function.ipynb
harripj: Looks good to me!

orix/plot/stereographic_plot.py
hakonanes: We can follow this up in #315.

hakonanes:
We agreed to wait for #334 to be merged, so that the equal area sampling added here becomes an alternative in the `sample_S2()` function to the unit vector samplings in that PR. From my point of view, this is the only outstanding task in this PR.

orix/sampling/S2_sampling.py
harripj: I have now added `"equal_area"` to the `_sampling_method_registry` introduced in #334, which I think fixes this issue.

harripj:
I have now merged `upstream/develop` into this PR and added the `equal_area_mesh` sampling method to the new sampling syntax introduced in #334. This PR is good for another look!

CHANGELOG.rst
```diff
@@ -14,6 +14,9 @@ Added
 -----
 - Support for type hints has been introduced and a section on this topic has been added
   to the contributing guide.
+- `Vector3d.pole_density_function()` has been implemented which allows for calculation
```
hakonanes, line 17:
```suggestion
- ``Vector3d.pole_density_function()`` has been implemented which allows for calculation 
```

doc/reference.rst
```diff
@@ -564,6 +564,7 @@ Vector3d
     get_nearest
     in_fundamental_sector
     mean
+    pole_density_function
```
hakonanes, line 567: Could you add the corresponding method to the autosummary of `StereographicPlot` as well?

orix/plot/stereographic_plot.py
```diff
@@ -167,6 +169,90 @@ def plot(self, *args, **kwargs):
...
+        colorbar: bool = True,
+        **kwargs,
+    ):
+        """Compute the Pole Density Function (PDF) of vectors in the
+        stereographic projection.
+
+        Parameters
+        ----------
+        azimuth, polar
+            Azimuth and polar angles in radians.
```
hakonanes, line 188:
```suggestion
            Azimuth and polar angles of the vectors in radians.
```

... or something similar.

orix/plot/stereographic_plot.py
```diff
@@ -167,6 +169,90 @@ def plot(self, *args, **kwargs):
...
+        resolution
+            The angular resolution of the sampling grid in degrees.
+            Default value is 1.
+        sigma
+            The angular resolution of the applied broadening in degrees.
+            Default value is 5.
+        log
+            If True the log(PDF) is calculated. Default is True.
+        colorbar
+            If True a colorabar is shown alongside the PDF plot.
```
hakonanes, line 198:
```suggestion
            If True a colorbar is shown alongside the PDF plot.
```

orix/plot/stereographic_plot.py
```diff
@@ -167,6 +169,90 @@ def plot(self, *args, **kwargs):
...
+        kwargs
+            Keyword arguments passed to
+            :meth:`matplotlib.axes.Axes.pcolormesh`.
+
+        See Also
+        --------
+        matplotlib.axes.Axes.scatter
+        """
+        from orix.sampling.S2_sampling import _sample_S2_equal_area_coordinates
+        from scipy.ndimage import gaussian_filter
```
hakonanes, line 209: I suggest to import `gaussian_filter` along with the other imports. As far as I remember, all imports within functions in orix are there to avoid circular imports.

doc/pole_density_function.ipynb
```diff
@@ -0,0 +1,383 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig = plt.figure(figsize=(2 * h, h))\n",
+    "ax0 = fig.add_subplot(121, projection=\"stereographic\")\n",
+    "ax1 = fig.add_subplot(122, projection=\"3d\")\n",
+    "\n",
+    "v_mesh = sample_S2_equal_area_mesh(resolution=10)\n",
```
hakonanes, line 172: Now that `sample_S2(10, "equal_area")` is available, should this be used instead?

doc/reference.rst
hakonanes: Also, could you add `sample_S2_equal_area_mesh` to the autosummary of the `sampling` module.

orix/sampling/S2_sampling.py
```diff
@@ -59,29 +175,113 @@ def sample_S2_uv_mesh(resolution: float) -> Vector3d:
...
+
+    return Vector3d.from_polar(azimuth=azimuth_prod, polar=polar_prod).unit
+
+
+def sample_S2_cube_mesh(
+    resolution: float, grid_type: str = "spherified_corner"
+) -> Vector3d:
     """Vectors of a cube mesh projected on a unit sphere *S2*.
 
+    Taken from diffsims.
```
hakonanes, line 283: I suggest to remove this note, now that @din14970 himself have ported the functionality.

orix/tests/sampling/test_s2_sampling.py
```diff
@@ -57,6 +92,52 @@ def test_cube_mesh(self, grid_type, resolution, size):
...
+        azi2, polar2, _ = v2.to_polar()
+        assert polar2.ndim == 2
+        assert azi2.shape == polar2.shape
+        assert np.count_nonzero(polar2 == 0) > 1
+        assert np.count_nonzero(polar2 == np.pi) > 1
+
+    def test_equal_area_mesh_raises(self):
+        with pytest.raises(ValueError, match="Hemisphere must be one of "):
+            sampling.sample_S2_equal_area_mesh(10, hemisphere="test")
             _ = S2_sampling.sample_S2_cube_mesh(2, "hexagonal")
```
hakonanes, line 141: This line is never reach, so we can remove it, right?

orix/tests/test_vector3d.py
```diff
@@ -195,6 +196,20 @@ def test_dot_outer(vector, something):
...
+    d_lazy_no_pb = vector.dot_outer(
+        something, lazy=True, progressbar=False, chunk_size=25
+    )
+    assert np.allclose(d_lazy, d_lazy_no_pb)
+
+
+def test_dot_outer_progressbar(vector, something, capsys):
+    d = vector.dot_outer(something, lazy=True, progressbar=True)
+    out, _ = capsys.readouterr()
+    assert "Completed" in out
```
hakonanes, line 211: Nice, did not know of this way to capture output!

orix/tests/test_vector3d.py
```diff
@@ -436,6 +451,112 @@ def test_rmul_notimplemented(self, vector):
...
+        qm2 = [isinstance(c, QuadMesh) for c in fig2.axes[0].collections]
+        assert any(qm2)
+        qmesh2 = fig2.axes[0].collections[qm2.index(True)].get_array().data
+        plt.close(fig2)
+
+        # test mesh not the same, log is different
+        assert not np.allclose(qmesh1, qmesh2)
+
+    def test_pdf_empty_vector(self):
+        v = Vector3d(np.empty((0, 3)))
```
hakonanes, line 548: We have `Vector3d.empty()` for this.

orix/vector/vector3d.py
```diff
@@ -78,46 +83,46 @@ class Vector3d(Object3d):
...
 
     @property
-    def x(self):
+    def x(self) -> np.ndarray:
         """numpy.ndarray : This vector's x data."""
         return self.data[..., 0]
 
     @x.setter
-    def x(self, value):
+    def x(self, value: np.ndarray) -> None:
```
hakonanes, line 91: What is the benefit of `-> None`? If there is none, I suggest to drop giving this particular type hint for every method or function that doesn't return anything.

orix/vector/vector3d.py
```diff
@@ -307,18 +341,41 @@ def dot_outer(self, other):
...
+                with ProgressBar():
+                    da.store(sources=dp, targets=dots)
+            else:
+                da.store(sources=dp, targets=dots)
+        else:
+            dots = np.tensordot(self.data, other.data, axes=(-1, -1))
         return dots
 
-    def cross(self, other):
+    def _dot_outer_dask(self, other: Vector3d, chunk_size: int = 20) -> da.Array:
```
hakonanes, line 356: I suggest to move all private methods to the bottom of the class, even though it is used much further up. I find this cleaner. I'm open to other opinions, of course.

orix/vector/vector3d.py
```diff
@@ -584,84 +651,216 @@ def get_circle(self, opening_angle=np.pi / 2, steps=100):
...
+        resolution
+            The angular resolution of the sampling grid in degrees.
+            Default value is 1.
+        sigma
+            The angular resolution of the applied broadening in degrees.
+            Default value is 5.
+        log
+            If True the log(PDF) is calculated. Default is True.
+        colorbar
+            If True a colorabar is shown alongside the PDF plot.
```
hakonanes, line 691:
```suggestion
            If True a colorbar is shown alongside the PDF plot.
```

orix/vector/vector3d.py
```diff
@@ -584,84 +651,216 @@ def get_circle(self, opening_angle=np.pi / 2, steps=100):
...
+            Whether to return the figure (default is False).
+        kwargs
+            Keyword arguments passed to
+            :meth:`matplotlib.axes.Axes.pcolormesh`.
+
+        Returns
+        -------
+        fig
+            The created figure, returned if `return_figure` is True.
+        """
```
hakonanes, line 734: Add a "See Also" section with `~orix.plot.StereographicPlot.plot_density_function`?

orix/plot/stereographic_plot.py
```diff
@@ -167,6 +169,90 @@ def plot(self, *args, **kwargs):
...
+        colorbar
+            If True a colorabar is shown alongside the PDF plot.
+            Default is True.
+        kwargs
+            Keyword arguments passed to
+            :meth:`matplotlib.axes.Axes.pcolormesh`.
+
+        See Also
+        --------
+        matplotlib.axes.Axes.scatter
```
hakonanes, line 206: Add `orix.vector.Vector3d.plot_density_function`?

hakonanes (changes requested):
Many minor suggestions, mostly of internal nature (syntax etc.). I suggest to merge after these are addressed.

orix/vector/vector3d.py
harripj: No problem with this!

orix/tests/test_vector3d.py
harripj: Thanks!

orix/tests/sampling/test_s2_sampling.py
harripj: Yes, sorry, I think this was a merge error.

orix/plot/stereographic_plot.py
harripj: In fact this has now been updated to allow `Vector3d` argument types also.

orix/vector/vector3d.py
harripj: If we ever consider using type checking in the future, [mypy recommends that functions have a declared return type](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-any-type). I agree that this seems a bit excessive to be honest and from what I can see it is not so clear whether this also applies to setters.

As we are not using type checking currently, I will remove these `None` return types and we can cross that bridge if we come to it.

orix/plot/stereographic_plot.py
hakonanes: For some reason this was mixed up? It still reads `colorabar`.

harripj: Whoops, fixed now.

hakonanes: Btw, not a big problem, but I prefer if the reviewer marks conversations as resolved rather than the submitter. This makes it a bit faster, at least for me, to look through the comments and check that they are adressed.

orix/tests/sampling/test_s2_sampling.py
hakonanes: No problem, just one fewer lines to worry about :)

orix/plot/stereographic_plot.py
harripj: Good point- I will keep this in mind in the future!

orix/vector/vector3d.py
hakonanes: Type checking is indeed something to consider in the future. I've referenced this comment in the relevant issue.

orix/plot/stereographic_plot.py
hakonanes: This hasn't been added yet. Do you consider it unnecessary?

harripj: I'm sorry about this- I had also committed this suggestion but perhaps this (and the mistake above) got lost when I was changing branches.

hakonanes: No problem!

orix/plot/stereographic_plot.py
```diff
@@ -167,6 +170,107 @@ def plot(self, *args, **kwargs):
...
+                    "If one argument is passed it must be an instance of "
+                    + "`orix.vector.Vector3d`."
+                )
+            azimuth, polar, _ = v.to_polar()
+        elif len(args) == 2:
+            azimuth, polar = args
+        else:
+            raise ValueError(
+                "Args must be either `Vector3d` or arrays of azimuth and polar angles."
+            )
```
hakonanes, line 228: Final suggestion here... `StereographicPlot._pretransform_input()` is used in most methods to parse the input arguments with this error message raised if they couldn't be parsed:

```python
                raise ValueError(
                    "Accepts only one (Vector3d) or two (azimuth, polar) input "
                    "arguments"
                )
```

I suggested to use that wording here as well, to be inline with the existing error message.

hakonanes (approved):
I'm very happy with this functionality, and will merge once the checks pass.

Issue #349, opened 2022-05-20
`Black` formatting for Jupyter Notebooks


harripj:
There is a `jupyter` option for `black` that will format notebook cells: see the installation section https://pypi.org/project/black/

This would be pretty easy to implement if people are interested and would lead to consistent notebook formatting. We would perhaps have to add some `# fmt: off` comments, but I think that this is an easy trade off to improve the look of the doc notebooks.

What does everyone think?

harripj:
Great! Will make a PR and we can discuss further.

hakonanes:
Done in #350!

PR #350, opened 2022-05-20
Add `black` automatic formatting to Jupyter notebooks

harripj:
As discussed in #349, we already use `black` for code formatting, and we can extend this to Jupyter notebooks. When installing `black` with `pip`, there is the option `black[jupyter]` (and `black-jupyter` package on `conda`). By default `black` does not format notebooks, but will if `.ipynb` files are passed to it.

In this PR `black-jupyter` is added as a `pre-commit` hook, the optional dependencies, ie. `black[jupyter]`, are added to `setup.py`, a note has been added to the contributing guide on this topic, and the existing notebooks have been formatted.

#### Progress of the PR

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,11 @@ jobs:
       - uses: actions/checkout@v2
       - uses: psf/black@stable
 
+      - name: Check code style of Jupyter notebooks
+        run: |
+          pip install black[jupyter]
+          black doc/*.ipynb
+
   # Make sure all necessary files will be included in a release
   manifest:
     name: check manifest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,11 +10,11 @@ jobs:
       - uses: actions/checkout@v2
       - uses: psf/black@stable
 
-      - name: Check code style of Jupyter notebooks
-        run: |
-          pip install black[jupyter]
-          black doc/*.ipynb
+      - name: Install Black with Jupyter extension
+        run: pip install black[jupyter]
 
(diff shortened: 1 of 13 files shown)


pc494:
Will let @hakonanes review, but can confirm that I think this is a good idea.

harripj:
> I've gone over all the notebooks, and apart from this formatting in the CrystalMap guide

You're right- I will move the comment above to bring the code onto one line.

> I couldn't find an off-the-shelf way to check the code style of notebooks in the [Black documentation](https://black.readthedocs.io/en/stable/integrations/github_actions.html), but we could add one ourselves with something like

Very good idea!

.github/workflows/build.yml
```diff
@@ -10,6 +10,11 @@ jobs:
       - uses: actions/checkout@v2
       - uses: psf/black@stable
 
+      - name: Check code style of Jupyter notebooks
+        run: |
+          pip install black[jupyter]
+          black doc/*.ipynb
+
```
hakonanes, line 17:
```suggestion
      - name: Install Black with Jupyter extension
        run: pip install black[jupyter]

      - name: Check code style of Jupyter notebooks
        run: black doc/*.ipynb
```

hakonanes: Separating the steps increases readability. Sorry for not suggesting this in the first place.

hakonanes (approved):

hakonanes:
Thanks! Will merge once the checks pass.

PR #353, opened 2022-05-23
Add `plot_ipf_color_key()` to `InversePoleFigurePlot`

harripj:
#### Description of the change
As raised by @maclariz in #352.

At the moment it is difficult to display an IPF colour key on a custom figure, which would be very useful behaviour for users when presenting their data. In this PR the `IPFColorKeyTSL.plot()` code is refactored such that the same functionality can be called from `InversePoleFigurePlot` through a new `plot_ipf_color_key()` method.

This feature has also been discussed in #332.

#### Progress of the PR
- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Unit tests with pytest for all lines
- [ ] Clean code style by [running black via pre-commit](https://orix.readthedocs.io/en/latest/contributing.html#code-style)

#### Minimal example of the bug fix or new feature
```python
# old behaviour
from orix.plot.orientation_color_keys import IPFColorKeyTSL
from orix.quaternion import symmetry, Orientation
from orix import plot
from orix.vector import Vector3d

ipfkey = IPFColorKeyTSL(symmetry=symmetry.Oh)
ipfkey.plot()
```
de5d449c-8036-4c52-81b4-3c9bf2475b16

```python
# new behaviour from this PR
from matplotlib import pyplot as plt
fig, ax = plt.subplots(subplot_kw=dict(projection='stereographic'))
o = Orientation.random(10000)
c = ipfkey.orientation2color(o)
ax.scatter(o * Vector3d.zvector(), s=2, c=c)

ax2 = fig.add_axes([0.8, 0.8, 0.2, 0.2], projection='ipf', symmetry=symmetry.Oh)
ax2.plot_ipf_colour_key()
```
a54590b3-e275-4303-b99f-b33f00b088c8

#### For reviewers

- [ ] The PR title is short, concise, and will make sense 1 year later.
- [ ] New functions are imported in corresponding `__init__.py`.
- [ ] New features, API changes, and deprecations are mentioned in the unreleased
      section in `CHANGELOG.rst`.
- [ ] Contributor(s) are listed correctly in `__credits__` in `orix/__init__.py` and in
      `.zenodo.json`.

Issue #352, opened 2022-05-23
Plotting IPF key in matplotlib axes


maclariz:
It is easy to display both maps and IPF keys using the .plot() method.

Maps are also easy to put in matplotlib axes for assembling a nice figure for publication using ax.plot_map(....

It is not so easy to do this for the IPF key as far as I can work out.  At the moment, my work around is to make a figure from the IPF key and then plot that figure in an ax.imshow().  Here is my code:

```python
import io as inout
import cv2

def get_img_from_fig(fig, dpi=180):
    buf = inout.BytesIO()
    fig.savefig(buf, format="png", dpi=dpi)
    buf.seek(0)
    img_arr = np.frombuffer(buf.getvalue(), dtype=np.uint8)
    buf.close()
    img = cv2.imdecode(img_arr, 1)
    img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
    return img

IPFkey = plot.IPFColorKeyTSL(pg_HCP).plot(return_figure=True)

ax.imshow(get_img_from_fig(IPFkey, dpi=600))
```

Surely there is a nicer way to do this

Thanks

Ian

harripj:
Hi Ian,

thanks for raising this and sharing your workaround! I have come across this problem too in the past and completely agree that this is not the best behaviour. We should implement a `plot_color_key()` method or similar so users can add axes wherever they like on their figure and also show the IPF colour key on a separate set of axes, to allow for more customisability.

harripj:
@maclariz I have put something together in #353 which allows the IPF colour key to be plotted in a separate set of axes on the same figure. Is this the type of functionality you had in mind? (Please bear in mind that the PR in #353 is currently a draft!)

maclariz:
@harripj I won't get chance to test now as about to have dinner and then do something totally unrelated to science for the evening.  But maybe get a chance to have a play later tomorrow.

harripj:
No rush at all, have a look when you can. I hope you have a good evening! 😄

hakonanes:
Also see #332, where @emichr asked rhis question and I gave an answer.

hakonanes:
Thanks for raising this issue, Ian. That both you and @emichr raised this issue within two weeks shows that this must be made easier.

I've pointed out a couple of improvements in #332 which will make obtaining the IPF color key and including it in a plot easier and more flexible. @harripj started in this direction in #353. I suggest we continue the discussion in #332.

harripj:
[...]

Issue #332, opened 2022-05-10
User guide notebook detailing use of Matplotlib to produce plots


hakonanes:
@emichr asked how to include for example a `CrystalMap` plot with scalebar and an inverse pole figure in two separate axes but in the same figure. This isn't shown in the documentation, but is easily done using the Matplotlib registered custom "projections" written in orix:

```python
import matplotlib.pyplot as plt

from orix import plot
from orix.crystal_map import CrystalMap
from orix.quaternion import Orientation, symmetry
from orix.vector import Vector3d


plt.rcParams.update({"font.size": 15, "figure.figsize": (10, 5)})

xmap = CrystalMap.empty((20, 30), step_sizes=(1, 1))
xmap.scan_unit = "nm"

g = Orientation.random(50)
g.symmetry = symmetry.Oh

v = Vector3d.zvector()

fig = plt.figure()
ax0 = fig.add_subplot(121, projection="plot_map")
ax0.plot_map(xmap, xmap.x)
ax0.remove_padding()
ax1 = fig.add_subplot(
    122,
    projection="ipf",
    symmetry=g.symmetry,
    direction=v,
    hemisphere="upper"
)
ax1.set_title(str(v.data.squeeze()), fontweight="bold")
ax1.scatter(g)

fig.savefig("/home/hakon/kode/orix_test/crystal_map_ipf_plot_same_figure.png", bbox_inches="tight", pad_inches=0, dpi=150)
```

![crystal_map_ipf_plot_same_figure](https://user-images.githubusercontent.com/12139781/167644684-2dcb20f2-5784-433f-a201-85a94a959a24.png)

I suggest to add a new user guide topic "Plotting" where "Using orix with Matplotlib" can be explained in more detail, going beyond the convenience methods like `CrystalMap.plot()`, `Vector3d.plot()`, and `Orientation.plot()`, which are covered in other user guides.

hakonanes:
> it seems that adding axes for showing color keys (such as IPFColorKeyTSL) is more complicated

[...]

PR #354, opened 2022-05-24
Ax ipf plot

harripj:
As raised by @maclariz in #352.

At the moment it is difficult to display an IPF colour key on a custom figure, which would be very useful behaviour for users when presenting their data. In this PR the `IPFColorKeyTSL.plot()` code is refactored such that the same functionality can be called from `InversePoleFigurePlot` through a new `plot_ipf_color_key()` method.

This feature has also been discussed in #332.

NB. this is a continuation of #353 as I had messed up the commit history.

#### Progress of the PR
- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)

```python
# old behaviour
from orix.plot.orientation_color_keys import IPFColorKeyTSL
from orix.quaternion import symmetry, Orientation
from orix import plot

ipfkey = IPFColorKeyTSL(symmetry=symmetry.Oh)
ipfkey.plot()
```
de5d449c-8036-4c52-81b4-3c9bf2475b16

```python
# new behaviour from this PR
fig, ax = plt.subplots(subplot_kw=dict(projection='stereographic'))
o = Orientation.random(10000)
c = ipfkey.orientation2color(o)
ax.scatter(o * Vector3d.zvector(), s=2, c=c)

ax2 = fig.add_axes([0.8, 0.8, 0.2, 0.2], projection='ipf', symmetry=symmetry.Oh)
ax2.plot_ipf_colour_key()
```
a54590b3-e275-4303-b99f-b33f00b088c8

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/plot/direction_color_keys/direction_color_key_tsl.py
+++ b/orix/plot/direction_color_keys/direction_color_key_tsl.py
@@ -16,6 +16,9 @@
 # You should have received a copy of the GNU General Public License
 # along with orix.  If not, see <http://www.gnu.org/licenses/>.
 
+from typing import Optional, Tuple, Union
+
+from matplotlib.figure import Figure
 import numpy as np
 from scipy.interpolate import griddata
 
@@ -25,6 +28,7 @@ from orix.plot.direction_color_keys._util import (
     rgb_from_polar_coordinates,
 )
 from orix.projections import StereographicProjection
+from orix.quaternion import Symmetry
 from orix.sampling import sample_S2_cube_mesh
 from orix.vector import Vector3d
 
@@ -37,21 +41,21 @@ class DirectionColorKeyTSL(DirectionColorKey):
     This is based on the TSL color key implemented in MTEX.
     """
 
-    def __init__(self, symmetry):
+    def __init__(self, symmetry: Symmetry) -> None:
         """Create an inverse pole figure (IPF) color key to color
         crystal directions according to a Laue symmetry's fundamental
         sector (IPF).
 
         Parameters
         ----------
-        symmetry : orix.quaternion.Symmetry
+        symmetry
             (Laue) symmetry of the crystal. If a non-Laue symmetry
             is given, the Laue symmetry of that symmetry will be used.
         """
         laue_symmetry = symmetry.laue
         super().__init__(laue_symmetry)
 
-    def direction2color(self, direction):
+    def direction2color(self, direction: Vector3d) -> np.ndarray:
         """Return an RGB color per orientation given a Laue symmetry
         and a sample direction.
 
@@ -59,12 +63,12 @@ class DirectionColorKeyTSL(DirectionColorKey):
 
         Parameters
         ----------
-        direction : orix.vector.Vector3d
+        direction
             Directions to color.
 
         Returns
         -------
-        rgb : numpy.ndarray
+        rgb
             Color array of shape `direction.shape` + (3,).
         """
         laue_group = self.symmetry
@@ -73,21 +77,27 @@ class DirectionColorKeyTSL(DirectionColorKey):
         polar = 0.5 + polar / 2
         return rgb_from_polar_coordinates(azimuth, polar)
 
-    def plot(self, return_figure=False):
-        """Plot the inverse pole figure color key.
+    def _create_rgb_grid(
+        self, return_min_max: bool = False
+    ) -> Union[np.ndarray, Tuple[np.ndarray, Tuple[float, float], Tuple[float, float]]]:
+        """Create the 2d colormap used to represent crystal directions.
 
         Parameters
         ----------
-        return_figure : bool, optional
-            Whether to return the figure. Default is False.
+        return_min_max
+            If True the tuples `(min, max)` of the fundamental sector in
+            the stereographic projection for both `x` and `y` are also
+            returned. Default is False
 
         Returns
         -------
-        figure : matplotlib.figure.Figure
-            Color key figure, returned if `return_figure` is True.
+        rgb_grid
+            Colormap values.
+        (x_min, x_max), (y_min, y_max)
+            Tuples `(min, max)` of the fundamental sector in the
+            stereographic projection for both `x` and `y`. Returned if
+            `return_min_max` is True.
         """
-        from orix.plot.inverse_pole_figure_plot import _setup_inverse_pole_figure_plot
-
         laue_group = self.symmetry
         sector = laue_group.fundamental_sector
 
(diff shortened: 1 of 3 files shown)


harripj:
> Hi Ian, to install @harripj's changes in the branch this PR is based on, you should be able to do (see [Python package installation documentation](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-from-vcs)):
> 
> ```shell
> python3 -m pip install -e git+https://github.com/harripj/orix.git@ax-ipf-plot#egg=orix
> ```
> 
> @harripj, I suggest to make a new PR or rebase your branch on the latest commits in `develop`, so only your commits are included in this PR.

Sorry for the confusion all, let's keep the conversation here. I am not quite sure what happened but when rebasing on `upstream/develop` it pulled the changes from #340.

The steps that @hakonanes showed above are still correct for pulling the changes here.

maclariz:
Fails currently:

     20 axb = plt.subplot(gs[0,1], projection='ipf', symmetry=pg_HCP)
---> 21 axb.plot_ipf_colour_key()
     22 
     23 plt.tight_layout()

AttributeError: 'InversePoleFigurePlotSubplot' object has no attribute 'plot_ipf_colour_key'

harripj:
Sorry I should have update the example, the function has been renamed: could you try `axb.plot_ipf_color_key()`without the `u`?

maclariz:
002_IPF_figure_alpha
It works!

Thanks

Ian

maclariz:
Found another problem. 

If I put this in a subplot (in this case made with gridspec), I have no way to control the size of the resulting IPF key and it tends to spill out of the edges with the labels.

Normally, I fix this with graphs by putting them inside a set of inset axes within the gridspec box.  But inset_axes does not support "projection" so I cannot put custom projections in inset_axes.  Another way would be to specify the plot range, but I don't see any handles for doing that in your plot_ipf_color_key, or any way to insert padding around the edges or something like that.  

Maybe this would be ideal, because quantitative evaluation of orientation data often uses multiple plots in a single figure to cover all the data and analysis, and therefore stuffing the key properly inside the axes would be a useful feature.  It is not an instant disaster, however!  I am not publishing this today or this week or even this month, but we are working towards publication of this data and so it would be nice to have this functionality at some point.

Thanks

Ian


002_IPF_figure_beta

harripj:
Hi Ian, you should be able to manually adjust these parameters using the normal matplotlib functions, for example:
```python
fig, ax = plt.subplots(ncols=2, subplot_kw=dict(projection="ipf", symmetry=symmetry.Oh))

for a in ax:
    a.plot_ipf_color_key()

# change axes limits of right hand plot
ax[1].set_xlim(-0.2, 0.6)
ax[1].set_ylim(-0.5, 0.6)

# change individual text parameters for crystal labels on right hand plot
ax[1].texts[1].update({'size': 20})
ax[1].texts[2].set_x(0.1)
```
f0b6cc6c-3867-4087-8b37-7362e6dccca8

Regarding this PR (and orix in general) and how much freedom we build into the orix plotting functions, eg. that is easily accessible using `plot_ipf_color_key()`, my feeling is that it is best to implement a simple function (as is currently the case) and let users finely adjust their plots using the matplotlib syntax (as done here) if they wish to. I think each user will have different specs as to how they want to format their plots and it will complicate things building this into orix when the access is already there using matplotlib. @hakonanes what are your thoughts on this?

maclariz:
Thanks @harripj!
This works.  Kinda odd that it isn't set on a default 0-1 axis, but that's fine.  I suppose you could write a pad function if you wanted, but that isn't that necessary.  Perhaps simply putting a note on the function that you can pad it this way would be enough.

harripj:
> Thanks @harripj!
> This works. 

Great to hear! Hopefully in the near future this feature will be available in a release.

> Kinda odd that it isn't set on a default 0-1 axis, but that's fine.

The units of the plot are from the fundamental zone vertices which have undergone stereographic projection, which is to say that the axes limits will change if the point group is changed.

>  I suppose you could write a pad function if you wanted, but that isn't that necessary. Perhaps simply putting a note on the function that you can pad it this way would be enough.

Good idea, we are always looking to improve the documentation!

hakonanes:
This PR was merged by mistake. I cannot revert it automatically (via the "Revert" button), but hope to be able to after #359 is reverted.

@harripj, are you able to set up new PRs for these?

harripj:
Yes, I will do that now. Sorry about the mess.

hakonanes:
Turns out reverting was very easy, just had to do it in reverse order of when they were merged!

maclariz:
One other issue with this.  This works really well with 622 (HCP) where the pole figure spans the whole 0-1 x-range, and the 6/mmm title is well centred.  This works less well with 432 (BCC/FCC) where the pole figure spans the 0-0.5 x range, and if you restrict the axis size, it them leaves the m3m title over to the left hand side.  Some way of controlling or centring that title is needed.

harripj:
Could you provide an example of this?

harripj:
Thanks. MPL defaults the title to the center of the axes I believe. If you want to adjust it, in your case you could do, for example:
```python
axf.title.set_x(0.4)
```

maclariz:
@harripj This fails.  It is not considered a title by MPL.  You can add a separate title.  I presume this is purely an ax.text() call.

harripj:
This is coded as a `set_title` call: https://github.com/pyxem/orix/blob/2d18672bd3beccf5b9da78e1f22e944d12870c33/orix/plot/inverse_pole_figure_plot.py#L305-L307

> This fails

Hmm, perhaps this is a version issue, the code above works in my environment with MPL version 3.4 and 3.5. Can you check your MPL version?

In general though I think the default behaviour, ie. plotting the title in the middle of the upper axis, is the best behaviour in most cases and should be default. Any custom plots can be specced by the user in each case using MPL functions etc, this way they have the most flexibility.

maclariz:
@harripj I found out that the title is being written to:
set_title('m3m', loc='left')

It can be corrected by writing:
axb = plt.subplot(gs[0,2], projection='ipf', symmetry=pg_FCC)
axb.set_xlim(-0.2,0.6)
axb.plot_ipf_color_key()
a = axb.get_title(loc = 'left')
axb.set_title(a, loc='center')
axb.set_title('', loc='left')

hakonanes:
Good that you found a solution (for now), @maclariz.

To get the most out of yours and @harripj's efforts here, I think it will be very useful if we could summarize your experience of creating the figure shown in https://github.com/pyxem/orix/pull/354#issuecomment-1155226752. This summary can guide us to create useful convenience methods to make plotting easier, like @harripj has done in #361 (2nd version of this PR).

This work should be combined with #332 to create a tutorial (or topic with several tutorials) showing how to use the convenience methods to create these rich figures.

maclariz:
I think you should have:

-  the default title as you have, but allow a user defined title as an alternative
- explicit setting of loc='center' as default, but clear comment that loc='left' or 'right' can be passed, along with x and y values for the title position (if wished to move manually) - this is all things already in set_title, so you are just passing parameters to that method
- default setting of xlim as (min - 0.1, max + 0.1) with the possibility to write xlim_min and xlim_max explicitly.  Again, just passing parameters to an existing MPL method
- possibility to set fontweight for title and other kwargs (and include in documentation)

I think that's it!

Issue #351, opened 2022-05-23
Access pole density function results outside of plot


hakonanes:
This suggestion is a follow up to #340, where we (@harripj and I) discussed introducing functionality to easily access the pole density function (PDF) results (the histogram and bin coordinates, and perhaps some derived values) outside of the plot.

We agreed that a new module `orix.measure` could be introduced for all (or most) quantification functionality in orix.

Any thoughts on this module and approach are very welcome.

harripj:
I think this is a good idea. Currently in #340 the PDF logic is contained within `StereographicPlot.pole_density_function()`, which makes it quite inaccessible outside of the plotting functionality. It would be better move it to a more accessible place as @hakonanes discussed above.

harripj:
Reopened due to previous bad merge #355.

hakonanes:
> We agreed that a new module orix.measure could be introduced for all (or most) quantification functionality in orix.

@pc494, do you have any thoughts on this module? @harripj has introduced this in #362.

PR #355, opened 2022-05-27
Inverse pole density distribution

harripj:
Following on from #340, it is often useful to view the pole density function (PDF) in the fundamental sector of the crystal symmetry, ie. the inverse pole density function (IPDF).

To implement this, the `StereographicPlot.pole_density_function()` has been updated to work with the subclass `InverseStereographicPlot`. When calculating PDF, the axes are checked as to whether they have a `_symmetry` attribute, ie. `IPF`, and if so the calculated `PDF` is folded back into the fundamental sector.

At the moment there are some issues with the implementation, regarding aliasing (during folding procedure) and the axes limits (which I may be due to `QuadMesh`, see plot size below), see the examples below. Grateful for any suggestions, feedback, or help regarding this!

#### Progress of the PR

```python
from orix import plot
from orix.quaternion import symmetry

fig = plt.figure(figsize=(15, 5))

ax0 = fig.add_subplot(131, projection='stereographic')
ax1 = fig.add_subplot(132, projection='ipf', symmetry=symmetry.Oh)
ax2 = fig.add_subplot(133, projection='ipf', symmetry=symmetry.C6)

v = Vector3d((-1, 0, 1))

ax0.pole_density_function(v)
ax1.pole_density_function(v, resolution=0.25)
ax2.pole_density_function(v, resolution=0.25)
```

fa97fbbe-cd32-4437-8c5c-cb0d22f45cce

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


hakonanes:
This will be a very nice feature indeed.

Is it not possible to project vectors into the fundamental sector, and then sample the fundamental sector only in equal area sampling and determine the histogram here? The challenge here is perhaps because of the sector edges? Do we have to determine how much of the bins on the edges are within the sector and "normalize" by this somehow?

harripj:
Yes that's right. I had originally tried this but the sector edges present a problem a) for the broadening as the bin values are 0 outside the sector and b) because the different edges do not sample the same azimuthal and polar range.

For the a) case I had considered reflecting the bins about the sector edge, this seemed to work well but there was still undersampling at the FS edges.

In the current implementation this is not a problem, just the aliasing effect demonstrated in the figures.

Perhaps I will commit this implementation and we could further discuss the pros/cons of both.

harripj:
I have had another look at this and it seems the aliasing problem originated from `sample_equal_area_mesh()`, there should have been an extra data point included with the `polar` sampling coordinates to accommodate the fact that the end point was also being sampled. This has been fixed.

As discussed in #351, I have added the `orix.measure` module (as discussed before with @hakonanes, I think the module name makes sense but it is open for suggestions) to include the code here such that the PDF can be calculated outside of plotting.

To relieve the aliasing effects discussed above, when computing IPDF the initial sampling sampling is performed on a finer grid (factor 2) which is then folded back into the grid at the requested resolution. The result is improved (old on left, current on right):
24e678dc-28b0-4847-b158-cd9a4422d393


and here is the updated initial example for point data:
```python
from orix import plot
from orix.quaternion import symmetry

fig = plt.figure(figsize=(15, 5))

ax0 = fig.add_subplot(131, projection='stereographic')
ax1 = fig.add_subplot(132, projection='ipf', symmetry=symmetry.Oh)
ax2 = fig.add_subplot(133, projection='ipf', symmetry=symmetry.C6)

v = Vector3d([(-2, 0, 1), [2, 2, 1], [1, 1, 2], [0.5, -0.3, 0.2]]).unit

ax0.pole_density_function(v, resolution=1)
ax1.pole_density_function(v, resolution=0.25)
ax2.pole_density_function(v, resolution=0.25)
```
6599d827-e942-4a6a-ab19-85e950bd8345

and random data:
```python
from orix import plot
from orix.quaternion import symmetry

fig = plt.figure(figsize=(15, 5))

ax0 = fig.add_subplot(131, projection='stereographic')
ax1 = fig.add_subplot(132, projection='ipf', symmetry=symmetry.Oh)
ax2 = fig.add_subplot(133, projection='ipf', symmetry=symmetry.C6)

import numpy as np
v = Vector3d(np.random.randn(10_000, 3)).unit

ax0.pole_density_function(v, resolution=1)
ax1.pole_density_function(v, resolution=0.25)
ax2.pole_density_function(v, resolution=0.25)
```
afe2e572-bac1-406b-95c3-a6c76e65ad5f

harripj:
The axes limits of the mesh is now properly set `clip_on=False` so that the figure size is as expected and not huge. The z-ordering of the mesh has been reduced to `0`, ie. like an image, so that the fundamental sector edges are properly displayed over it.

This PR is now ready for review.

hakonanes:
I hope to review this and the other open PR within early next week the latest, sorry for the delay. (I'm busy preparing for an NTNU workshop in two weeks and need new releases of diffsims and kikuchipy with new functionality before that...)

harripj:
No problem, thanks for taking the time!

PR #357, opened 2022-05-31
Add `isort` dependency and pre-commit hook

harripj:
As discussed with @hakonanes and @din14970 in https://github.com/pyxem/orix/pull/334#discussion_r873160646, `isort` is a package that may be used to automatically sort imports. It also hooks into pre-commit to allow for formatting before code submission.

We could consider adding `isort` as a dependency to `orix` to handle import formatting. In `pre-commit-config.yaml` I have added the [standard pre-commit config](https://pycqa.github.io/isort/docs/configuration/pre-commit.html) along with the `black` profile and also with the `"--force-sort-within-sections"` [flag](https://pycqa.github.io/isort/docs/configuration/options.html#sort-relative-in-force-sorted-sections) which groups imports by modules (as we currently do) rather than sorting straight `import`s before `from`-style imports. 

#### Progress of the PR

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -61,9 +61,10 @@ Code style
 
 The code making up orix is formatted closely following the `Style Guide for Python Code
 <https://www.python.org/dev/peps/pep-0008/>`_ with `The Black Code style
-<https://black.readthedocs.io/en/stable/the_black_code_style/index.html>`_. We use
-`pre-commit <https://pre-commit.com>`_ to run ``black`` automatically prior to each
-local commit. Please install it in your environment::
+<https://black.readthedocs.io/en/stable/the_black_code_style/index.html>`_ and
+`isort <https://pycqa.github.io/isort/>`_ to handle module imports. We use
+`pre-commit <https://pre-commit.com>`_ to run ``black`` and ``isort`` automatically
+prior to each local commit. Please install it in your environment::
 
     pre-commit install
 
@@ -74,6 +75,11 @@ Note that ``black`` won't format `docstrings
 <https://www.python.org/dev/peps/pep-0257/>`_. We follow the `numpydoc
 <https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard>`_
 standard.
+
+Note that if imports must be sorted in a certain order, for example to avoid recursion,
+then ``isort`` provides `commands
+<https://pycqa.github.io/isort/docs/configuration/action_comments.html>`_ that may be
+used to prevent sorting.
 
 Comment lines should preferably be limited to 72 characters.
 
--- a/orix/plot/stereographic_plot.py
+++ b/orix/plot/stereographic_plot.py
@@ -33,16 +33,19 @@ import matplotlib.pyplot as plt
 import numpy as np
 from scipy.ndimage import gaussian_filter
 
+# fmt: off
+# isort: off
 from orix.plot._symmetry_marker import (
     TwoFoldMarker,
     ThreeFoldMarker,
     FourFoldMarker,
     SixFoldMarker,
 )
+# isort: on
+# fmt: on
 from orix.projections import InverseStereographicProjection, StereographicProjection
 from orix.vector import Vector3d
 from orix.vector.fundamental_sector import _closed_edges_in_hemisphere
-
 
 ZORDER = dict(text=6, scatter=5, symmetry_marker=4, draw_circle=3, mesh=2)
 
--- a/orix/tests/quaternion/test_orientation.py
+++ b/orix/tests/quaternion/test_orientation.py
@@ -30,9 +30,9 @@ from orix.quaternion.symmetry import (
     D2,
     D3,
     D6,
-    T,
     O,
     Oh,
+    T,
     _groups,
     _proper_groups,
 )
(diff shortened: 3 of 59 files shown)


CONTRIBUTING.rst
```diff
@@ -75,6 +76,11 @@ Note that ``black`` won't format `docstrings
 <https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard>`_
 standard.
 
+Note that if imports must be sorted in a certain order, for example to avoid recursion,
+then ``isort`` provides `commands
+<https://pycqa.github.io/isort/docs/configuration/action_comments.html>`_ that may be
+used to prevent sorting.
+
 Comment lines should preferably be limited to 72 characters.
```
hakonanes, line 85: The paragraph below this line on structuring on package imports is redundant now, right? Or, a note that we're using `isort` could be added as an extra line to the paragraph.

orix/plot/stereographic_plot.py
```diff
@@ -33,17 +33,20 @@
 import numpy as np
 from scipy.ndimage import gaussian_filter
 
+# fmt: off
```
hakonanes, line 36: Is it necessary to also tell Black to not format these lines? Black didn't have to be told before, so why now?

orix/tests/quaternion/test_orientation.py
```diff
@@ -30,9 +30,9 @@
     D2,
     D3,
     D6,
-    T,
     O,
     Oh,
+    T,
```
hakonanes, line 35: Point groups are listed from lowest to highest symmetry, so this list should get a `#isort: skip` flag, right?

hakonanes: Or `off/on`.

orix/tests/sampling/test_cubochoric_sampling.py
```diff
@@ -20,12 +20,12 @@
 import pytest
 
 from orix.quaternion.rotation import Rotation
-from orix.quaternion.symmetry import C1, C2, D2, C4, D4, C3, D3, C6, D6, T, O
+from orix.quaternion.symmetry import C1, C2, C3, C4, C6, D2, D3, D4, D6, O, T
```
hakonanes, line 23: Shouldn't be sorted by isort.

orix/vector/__init__.py
```diff
@@ -25,10 +25,12 @@
 of a spherical region.
 """
 
-# Must be imported first
-from orix.vector.vector3d import check_vector, Vector3d
+# fmt: off
```
hakonanes, line 28: Again, do we have to tell both isort and Black to not sort?

hakonanes (changes requested):
This makes the imports easier and faster to read and frees up review time to be used on implementation and not code style, good addition!

Is it possible to add an `isort` check to the "code style" CI workflow?

orix/plot/stereographic_plot.py
harripj: Annoyingly `black` forces a new line before a comment, which I agree is a good idea but not for these action comments as it actually breaks up imports that may be from the same module. The alternative is:

```
# isort: off
import a.c
import a.b

# isort: on
import a.d, others...
```

Happy with either to be honest, perhaps it is easiest just to accept the new line (ie. as above), what do you think?

hakonanes: Hm, I'd say leave the Black comments in this case. Annoying, but worth it.

harripj:
Should be okay for another look @hakonanes.

The blank line comment behaviour discussed above is still present, despite using [certain parameters](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#custom-configuration). If `isort: skip` is used then imports will actually sort around the skipped line (not what we want). `isort: split` is better but as far as I can work out will add a new line around the split section. So I ended up using `fmt` and `isort` `on/off` around the regions, which gives the correct behaviour.

Note I also updated the `black-jupyter` CI test by adding the `--check` option. With an ill-formatted notebook previously the [test would pass](https://github.com/harripj/orix/runs/6690792272?check_suite_focus=true#step:7:11) but actually reformat the notebook, but now the [test will fail](https://github.com/pyxem/orix/runs/6690830763?check_suite_focus=true#step:7:12) which I think is more inline with what we want to test.

.github/workflows/build.yml
```diff
@@ -9,12 +9,19 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: psf/black@stable
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - uses: isort/isort-action@master
+        with:
+          configuration: --profile black --ff --fss --check-only --diff
```
hakonanes, line 17: Could you write out the options here? `--from-first` instead of `--ff` and `--force-sort-within-sections` instead of `--fss`. So we don't have to look them up to remember what they mean.

hakonanes (approved):
Minor comment which isn't strictly necessary but improves readability.

> So I ended up using fmt and isort on/off around the regions, which gives the correct behaviour.

I think that's fine. Thank you for checking.

> Note I also updated the black-jupyter CI test by adding the --check option

Yes, thank you!

hakonanes:
Thanks! Will merge once checks pass.

.github/workflows/build.yml
harripj: Good idea, in fact the `--ff` flag was incorrect, I have corrected it to `--filter-files`. Thanks!

PR #361, opened 2022-06-03
Add `plot_ipf_color_key()` to `InversePoleFigurePlot` 2

harripj:
Duplicate of #354 after that PR was pre-maturely merged.

Previous PR information below
-------------------------------
As raised by @maclariz in #352.

At the moment it is difficult to display an IPF colour key on a custom figure, which would be very useful behaviour for users when presenting their data. In this PR the `IPFColorKeyTSL.plot()` code is refactored such that the same functionality can be called from `InversePoleFigurePlot` through a new `plot_ipf_color_key()` method.

This feature has also been discussed in #332.

EDIT
I have also converted the RGB IPF image to RGBA and made the invalid points, ie. points outside the FS, transparent. This allows for better custom plot overlays, see the 3rd example below.

NB. this is a continuation of #353 as I had messed up the commit history.

#### Progress of the PR

```python
# old behaviour
from orix.plot.orientation_color_keys import IPFColorKeyTSL
from orix.quaternion import symmetry, Orientation
from orix import plot

ipfkey = IPFColorKeyTSL(symmetry=symmetry.Oh)
ipfkey.plot()
```
de5d449c-8036-4c52-81b4-3c9bf2475b16

```python
# new behaviour from this PR
fig, ax = plt.subplots(subplot_kw=dict(projection='stereographic'))
o = Orientation.random(10000)
c = ipfkey.orientation2color(o)
ax.scatter(o * Vector3d.zvector(), s=2, c=c)

ax2 = fig.add_axes([0.8, 0.8, 0.2, 0.2], projection='ipf', symmetry=symmetry.Oh)
ax2.plot_ipf_color_key()
```
a54590b3-e275-4303-b99f-b33f00b088c8

```python
from orix import data

xmap = data.sdss_ferrite_austenite(allow_download=True)
fig = xmap.plot(return_figure=True)

with plt.rc_context({'font.size': 6}):
    ax = fig.add_axes([0.23, 0.69, 0.15, 0.15], projection='ipf', symmetry=symmetry.Oh)
    ax.plot_ipf_color_key()
    ax.patch.set_facecolor('None')
```

c1fe10a1-2615-4331-9522-9927c1d1467c

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/doc/inverse_pole_figures.ipynb
+++ b/doc/inverse_pole_figures.ipynb
@@ -114,7 +114,7 @@
    "source": [
     "## Plot colour keys\n",
     "\n",
-    "Here we plot the colour keys for the eleven Laue groups."
+    "Here we plot the colour keys for the eleven Laue groups. An example demonstrating how an IPF color key can be added to a custom plot is shown in the [Crystal map notebook](crystal_map.ipynb#Plotting)."
    ]
   },
   {
@@ -213,14 +213,6 @@
     "    ori2.symmetry = ipfkey.symmetry\n",
     "    rgb_z = ipfkey.orientation2color(ori2)\n",
     "    ori2.scatter(\"ipf\", c=rgb_z, direction=ipfkey.direction)"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "d2632e3d",
-   "metadata": {},
-   "source": [
-    "An example demonstrating how an IPF color key can be added to a custom plot is shown in the [Crystal map notebook](crystal_map.ipynb#Plotting)."
    ]
   }
  ],
--- a/orix/plot/inverse_pole_figure_plot.py
+++ b/orix/plot/inverse_pole_figure_plot.py
@@ -313,10 +313,10 @@ class InversePoleFigurePlot(StereographicPlot):
             If ``True`` the Laue group name is displayed as the axes
             title.
 
-        This function maybe used to plot the IPF color key alongside
+        This function may be used to plot the IPF color key alongside
         another plot where the same key was used to color
-        :meth:`orix.quaternion.Orientation` or
-        :meth:`orix.vector.Vector3d` data.
+        :class:`~orix.quaternion.Orientation` or
+        :class:`~orix.vector.Vector3d` data.
         """
         symmetry = self._symmetry
         direction_color_key = DirectionColorKeyTSL(symmetry)
--- a/orix/plot/inverse_pole_figure_plot.py
+++ b/orix/plot/inverse_pole_figure_plot.py
@@ -33,8 +33,8 @@ import numpy as np
 
 from orix.crystal_map import Phase
 from orix.measure import pole_density_function
-from orix.plot.sterographic_plot import ZORDER, StereographicPlot
 from orix.plot.direction_color_keys.direction_color_key_tsl import DirectionColorKeyTSL
+from orix.plot.sterographic_plot import ZORDER, StereographicPlot
 from orix.quaternion import Orientation
 from orix.quaternion.symmetry import C1, Symmetry
 from orix.vector import Miller, Vector3d
--- a/orix/plot/inverse_pole_figure_plot.py
+++ b/orix/plot/inverse_pole_figure_plot.py
@@ -21,7 +21,7 @@
 :class:`~orix.vector.Vector3d`, typically parallel to sample directions,
 rotated by orientations.
 """
-from pyexpat.errors import XML_ERROR_INCOMPLETE_PE
+
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import matplotlib.axes as maxes
@@ -103,7 +103,9 @@ class InversePoleFigurePlot(StereographicPlot):
         return patches[self._has_collection(label="sa_sector", collections=patches)[1]]
 
     def scatter(
-        self, *args: Union[np.ndarray, Orientation, Vector3d], **kwargs: Any
+        self,
+        *args: Union[Tuple[np.ndarray, np.ndarray], Orientation, Vector3d],
+        **kwargs: Any,
     ) -> None:
         """A scatter plot of sample directions rotated by orientations,
         or orientations to rotate sample directions with.
--- a/orix/plot/inverse_pole_figure_plot.py
+++ b/orix/plot/inverse_pole_figure_plot.py
@@ -27,12 +27,13 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 import matplotlib.axes as maxes
 from matplotlib.figure import Figure
 from matplotlib.patches import PathPatch
-import matplotlib.pyplot as plt
 import matplotlib.projections as mprojections
+import matplotlib.pyplot as plt
 import numpy as np
 
 from orix.crystal_map import Phase
-from orix.plot import StereographicPlot
+from orix.measure import pole_density_function
+from orix.plot.sterographic_plot import ZORDER, StereographicPlot
 from orix.plot.direction_color_keys.direction_color_key_tsl import DirectionColorKeyTSL
 from orix.quaternion import Orientation
 from orix.quaternion.symmetry import C1, Symmetry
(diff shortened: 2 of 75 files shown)


harripj:
@hakonanes no problem! This was in response to @maclariz's comments raised in #352 and I have also found it useful for making my own custom plots. Perhaps the syntax could/should be adjusted somewhat, but we can discuss this once you have had a chance to look over it.

orix/plot/inverse_pole_figure_plot.py
```diff
@@ -297,16 +304,45 @@ def _pretransform_input_ipf(self, values):
             v = values[0] * self._direction
         return v.in_fundamental_sector(self._symmetry)
 
+    def plot_ipf_color_key(self) -> None:
```
hakonanes, line 307: I suggest to give the user the option to not show the title (Laue group) via `show_title=True`. The entire block starting with `label_xy =...` can be moved inside a `if show_title` block.

orix/plot/inverse_pole_figure_plot.py
```diff
@@ -297,16 +304,45 @@ def _pretransform_input_ipf(self, values):
...
+                [np.diff(self.get_xlim())[0] / 2, np.max(self.get_ylim())]
+            )
+            is_close = np.isclose(label_xy, expected_xy, atol=0.1).all(axis=1)
+            if any(is_close) and plt.rcParams["axes.titley"] is None:
+                loc = "left"
+
+        self.set_title(symmetry.name, loc=loc, fontweight="bold")
+        self.stereographic_grid(False)
+        self._edge_patch.set_linewidth(1.5)
+        self.imshow(rgba_grid, extent=(x_min, x_max, y_min, y_max), zorder=0)
```
hakonanes, line 334: Passing `extent=x_lim + y_lim` works, I suggest to remove the unpacking of the tuples above and just do that.

orix/plot/inverse_pole_figure_plot.py
```diff
@@ -297,16 +304,45 @@ def _pretransform_input_ipf(self, values):
             v = values[0] * self._direction
         return v.in_fundamental_sector(self._symmetry)
 
+    def plot_ipf_color_key(self) -> None:
+        """Plot an IPF color key code on this axis."""
```
hakonanes, line 308: I suggest to add a brief description below the first sentence description (with line break in between) of the intended use case of this function, i.e. to plot it alongside orientations or vectors colored with this key.

orix/plot/inverse_pole_figure_plot.py
hakonanes, line 307: I don't see the need for `None` here.

orix/plot/inverse_pole_figure_plot.py
hakonanes, line 308: Also, I suggest to allow passing any `font*` keyword argument allowed by `matplotlib.pyplot.text()` to format the title and axes labels with, like `fontsize`. This we can do like so

```python
from matplotlib.font_properties import FontProperties

kwargs = dict(fontsize=12, fontweight="bold")
fp_args = dict()
for key, value in kwargs.items():
    if key.startswith("font"):
        fp_args[key[4:]] = value
if len(fp_args) != 0:
    fp = FontProperties(**fp_args)
    if show_title:
        self.title.set_font_properties(fp)
    for text in self.texts:
        actual_text = text.get_text()
        if actual_text.startswith("[$") and actual_text.endswith("$]"):
            text.set_font_properties(fp)
```

If you agree with this, I can push this change to your branch.

hakonanes (changes requested):
This is a nice addition providing a convenient route to a quite common use case of plotting the IPF color key alongside a color map or pole figure.

I've made some suggestions, mostly for docstrings and the like.

Also, this PR addresses my points in https://github.com/pyxem/orix/issues/332#issuecomment-1125124631 nicely.

orix/plot/inverse_pole_figure_plot.py
harripj: This is just needed for type checking, which we discussed in [#338](https://github.com/pyxem/orix/issues/338).

For the moment, as type hinting is not applied everywhere in the codebase, this is not a problem. But if it gets introduced in the development phase, these return types would need to be added.

hakonanes: OK, then you can leave it.

orix/plot/inverse_pole_figure_plot.py
harripj: My feeling is that this could be handled with a `plt.rc_context` context manager eg:
```python
with plt.rc_context({'font.size': 20, 'axes.titlecolor': 'w'}):
    # plotting code
```

which would save us handling specifics. And if the user wants to adjust specifics then they could work through the MPL API. This could be better demonstrated in a notebook. What do you think?

orix/plot/inverse_pole_figure_plot.py
```diff
@@ -297,16 +304,50 @@ def _pretransform_input_ipf(self, values):
             v = values[0] * self._direction
         return v.in_fundamental_sector(self._symmetry)
 
+    def plot_ipf_color_key(self, show_title=True) -> None:
+        """Plot an IPF color key code on this axis.
+
+        This function maybe used to plot the IPF color key alongside
+        another plot where the same key was used to color
+        :meth:`orix.quaternion.Orientation` or
+        :meth:`orix.vector.Vector3d` data.
```
hakonanes, line 313:
```suggestion
        This function may be used to plot the IPF color key alongside
        another plot where the same key was used to color
        :class:`~orix.quaternion.Orientation` or
        :class:`~orix.vector.Vector3d` data.
```

orix/plot/inverse_pole_figure_plot.py
```diff
@@ -297,16 +304,50 @@ def _pretransform_input_ipf(self, values):
             v = values[0] * self._direction
         return v.in_fundamental_sector(self._symmetry)
 
+    def plot_ipf_color_key(self, show_title=True) -> None:
```
hakonanes, line 307: Can you add a "Parameters" section?

doc/inverse_pole_figures.ipynb
```diff
@@ -214,6 +214,14 @@
     "    rgb_z = ipfkey.orientation2color(ori2)\n",
     "    ori2.scatter(\"ipf\", c=rgb_z, direction=ipfkey.direction)"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d2632e3d",
+   "metadata": {},
+   "source": [
+    "An example demonstrating how an IPF color key can be added to a custom plot is shown in the [Crystal map notebook](crystal_map.ipynb#Plotting)."
```
hakonanes, line 223: I suggest to move this note up to after "Plot colour keys: Here we plot the colour keys for the eleven Laue groups. An example demonstrating...", because that's where I think users might want to look to do this, right?

orix/plot/inverse_pole_figure_plot.py
hakonanes: I agree in general and am OK with leaving this flexibility out of this PR.

I think setting the font size and possibly the weight will be the most desirable option for users of this method. I didn't see the suggested description of the `kwargs` parameter having to change in the future, so I figured this was quite an easy but stable flexibility to provide users. It's at least mentioned here so that we can come back to it in the future!

hakonanes (changes requested):
I've got a couple of more comments, and then I'm happy with this PR and ready to merge!

orix/plot/inverse_pole_figure_plot.py
harripj: Thanks!

doc/inverse_pole_figures.ipynb
harripj: Good idea!

hakonanes (approved):

harripj:
@hakonanes I just passed an eye over this and updated the `show_title` parameter docstring specifying the default value. At the same time tidied up other DS in related files. If you are happy with these changes, I am happy if this PR gets merged! 😄

hakonanes:
The other docstring changes are nice, thank you! Will merge once the checks pass.

The double back tick marks around e.g. \``True\`` is interpreted by Sphinx effectively as code, similar to Markdown's single back tick giving code on GitHub. I've started to use it, similar to NumPy, SciPy etc., to make interpreting docstrings a little faster. I think we should use this more in our docstrings, as you've done in the latest commit :+1:

maclariz:
Just out of interest, how soon until this is released in a full update of orix?  The box with the "official" version still has 0.9.0.post0 as the latest version.

hakonanes:
I'm OK with making a 0.10.0 release next week, as I quite like the pole density function functionality (used to great effect with a thin film dataset recently).

I have some improvements I'd like to make to the docs, separating it into examples (short recipies to common tasks) and tutorials (the existing user guides), like I've done with the [kikuchipy docs](https://kikuchipy.org/en/latest/#learning-resources).

I'll make an issue in preparation for 0.10.0.

Issue #374, opened 2022-08-15
Purpose for  neo_eulerian classes


argerlt:
General new-to-orix question, while thinking about code simplification:

What is the purpose of the neo_eulerian classes in vector/neo_euler.py? All of Orix's math, by design, is done with quaternions. As such, it seems their only purpose would be for to/from statements, or to make plotting/visualization easier. 

However, @hakonanes already wrote all the to/froms in quaternion/_conversions.py, so adding all the to/from statements directly to `Quaternions` would be incredibly easy. Similarly, any axis/angle plots could do the conversions when they are needed, instead of requiring a `Homochoric` or 'Rodriquez` class object. 

Bonus from a simplicity standpoint: Doing this eliminates 4 class objects, 143 lines of code, and (maybe most importantly?) makes sure all the to/from functions use the same conversion code from `_conversions.py`. Would also eliminate the use case of "can we provide support for Homochoric to/from Rodrigues in the Homochoric and Rodrigues classes" and similar such problems, since Quaternion will just always be the de-facto in between.

pc494:
I think the agree with almost everything you've said here. It made sense at one point to have these as a number of separate public classes, but at this point hiding that functionality away within the plotting code seems sensible.

hakonanes:
> wrote all the to/froms in quaternion/_conversions.py

The [conversions](https://github.com/pyxem/orix/blob/develop/orix/quaternion/_conversions.py) taken from Rowenhorst et al. (2015) were added to get [cubochoric sampling of orientations](https://orix.readthedocs.io/en/stable/uniform_sampling_of_orientation_space.html) (taken from Singh and De Graef [2016]) in orix, which we needed in kikuchipy for [EBSD dictionary indexing](https://kikuchipy.org/en/latest/tutorials/pattern_matching.html#Sample-orientation-space). The conversions are private, not ment to be used by a user. However, I would like to make them public in a clean way in the API and without sacrificing speed.

> adding all the to/from statements directly to `Quaternions`

How would you implement this? We can either go via the quaternion, as in `Quaternion.from_axes_angles(ax).to_homochoric()` requiring two steps, or `Quaternion.axes_angles_to_homochoric(ax)` in one step (how to name the methods can be discussed elsewhere). I don't see a drawback in supporting both.

argerlt:
> We can either go via the quaternion, as in Quaternion.from_axes_angles(ax).to_homochoric()
[...]

Issue #373, opened 2022-08-04
Fix inherited methods in the Orientation class


hakonanes:
The `Orientation` class inherits methods and properties from `Rotation`, `Quaternion` and `Object3d`. Currently, these methods and attributes aren't handled properly (e.g. symmetry isn't kept):

`Rotation`

`Quaternion`
* `mean()`

`Object3d`

Please add methods and attributes missing from the lists.

argerlt:
question about this because it came up, what should happen in the following situation?

`
from orix.quaternion import Orientation, symmetry

pg432 = symmetry.O
pg622 = symmetry.get_point_group(180)

o1 = Orientation((1,0,0,0),symmetry=pg432)
o2 = Orientation((1,0,0,0),symmetry=pg622)
r1 = Rotation((1,0,0,0))
r2 = Rotation((1,0,0,0))
m12 = o1-o2
m21 = o2-o1
'

o1 - o2 gives a misorientation with symmetry=(432,662). Makes sense
o1*o1 should give a symmetry of 432, but gives 1. this should be fixed
o1*r1 gives a rotation without symmetry. Makes sense, we are reorienting a rotation, and it agrees with how MTEX works.
m12*r1 also gives a rotation
r1*m12 has the same problem and solution as r1*o2, just use the symmetry of the object on the right. 

what should o1*o2 produce? should it have the symmetry of o1, o2, or a supergroup of both symmetries together? should it throw an error MTEX-style, or just expect users to be aware of the dangers?

Also, same question for m12*o1, o1*m12.

Also not sure how m12*m21 should work. this example isn't the best, but I'm just thinking of the case wehre there are 4 different symmetry objects. 

If you let me know how you want it, I can write this code, it was on my todo list anyway. functions to fix are 'mean' and '__mul__'. are there any others?

hakonanes:
Thank you for listing these use cases, we should provide sensible returns for all of them. I've updated the top comment with your suggestions.

This https://github.com/pyxem/orix/issues/249#issuecomment-1033751210 by @harripj on the (undocumented) misorientation convention in orix might be relevant. We discussed somewhere, and decided I believe, to retire the `o1 - o2` syntax and make `o2 * ~o1` the only supported operation to get the misorientation bringing `o1` into `o2`.

argerlt:
Okay, my two cents:

[...]

Issue #346, opened 2022-05-18
Using scipy's Rotation object


argerlt:
Scipy now has a rotation object, and it might be handy to utilize it in orix

PROS:
- In my (limited) testing, it is significantly faster at calculating misorientations between orientations due to the use of Cython
- you get passive and active rotations for free (try doing `R.from_euler('ZXZ',[euler_angles])` for activeversus `R.from_euler('zxz',[euler_angles])` for passive)
- you get support for mrp, axis-angles, and rotation matrices for free
- Instead of having to write up detailed docstrings for quaternion methods, you can just piggyback off of scipys
- there are a lot of python  classes out there for quaternions, switching to an already popular one will help reduce confusion for new users about conventions
- has built-in left-right multiplication functions to make adding SpecimenSymmetry easier

CONS:
- Scipy uses a different quaternion parameterization `(x, y, z, w)` versus orix and mtex `(w, x, y, z)`.
- This would require a lot of rewrites for something that would have limited immediate payoff other than a big speedup.

Let me know if you are interested in this, I wouldn't mind writing it, but it requires some fundamental changes as opposed to just adding features, so not going to do it unless you feel it would improve your code.

argerlt:
You're right, scipy currently sticks you to 1D arrays of rotations. At one point, I thought about making a child class of scipy Rotations to change that, or contributing directly to Scipy, but never did either.

As for conversion, since scipy does quaternions differently and I never managed to get a reliable converter working, here is the lazy method I use for converting between the two:

`from scipy.spatial.transform import Rotation as R
from orix.quaternion import Rotation as Rotation

scipy_oris = R.from_euler('ZXZ',orix_oris.as_euler())
orix_oris = scipy_oris.as_euler('ZXZ')
`

The conversion itself is slow, but it lets me use scipy's symmetry-aware misorientation calculations, which I'm more familiar with, and is faster than the old orix method (no idea about the new numpy quaterion method)

argerlt:
Okay, I gave this a go with the new numpy quaternion objects and the result honestly surprised me. Here is the quick and dirty speed test I did just now:

`
# quick and dirty speed test for misorientation calculations
import numpy as np
import quaternion
from scipy.spatial.transform import Rotation as R
import time

angles = np.random.rand(1000000, 3)*180

#scipy way
tic = time.time()
s = R.from_euler('ZXZ', angles)
mul1 = s*s.inv()
toc = time.time() - tic
[...]

Issue #345, opened 2022-05-18
making Orientations a child class of Rotations, not Misorientations.


argerlt:
I'd like to write some class method constructors for Misorientations which generates them from paired lists or arrays of orientations, as well as some other misorientation-specific functions.

Right now, Orientations use Misorientations as a base class, which violates [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) and leads to odd edge cases such as calling Orientation.equivalent(ori) and getting a doc string referencing misorientations.  It would really make more sense to make both use rotations as a base class, as opposed to overwriting their shared functions. For my specific case, this would make writing the classmethods a lot more straight forward. 

I was about to reorganize this for my own use anyway, but before i did, I wanted to check if there was a bigger reason I didn't understand/appreciate for misorientations being a base class for orientations. 

Thanks, I'm a big fan of orix.

hakonanes:
Hi @argerlt,

I've seen your [hackathon](https://github.com/mesoOSU/Mart2Aust_Hackathon) fork, and think it's great that you're reaching out to us now, even during the hackaton!

As far as I know, the choice of `Orientation` subclassing `Misorientation` was simply because, at the time `orix` was initially written, it was sufficient to consider orientations as misorientations where the initial "crystal" was the sample reference frame:

https://github.com/pyxem/orix/blob/14639a15a4c66c87891debb8656ef187c3b6e9ea/orix/quaternion/orientation.py#L473-L474

With the initial functionality, this made sense and made for a slightly simpler implementation. The functionality and use cases of mainly `Orientation` have grown quite a lot since then... I think you raising this issue is a sign that we should do the change you suggested. I have no arguments against it.

@pc494, who've been part of the project from the beginning might have some additional corrections to my "historical review".

argerlt:
Thanks, I'll give it a try.

And yeah, as a general FYI, we are trying to recreate a very specific piece of software that exists in MTEX but not in python, and Orix was agreed upon as the best starting point for a pythonic equivalent. As such, most of those forks and subforks contains a LOT of things you won't find useful, but the ideal goal is after this finishes up, myself and Steve Niezgoda will take the  parts that might be useful to the greater Orix crowd and start adding those as pull requests.

hakonanes:
> Thanks, I'll give it a try.

[...]

PR #376, opened 2022-08-18
Add n-dimensional support to private rotation conversion functions

argerlt:

By default, Rotation objects can contain n-dimensional arrays of Rotation data. However, _conversions.py is written to only accept 2D arrays, since the functions use jit for acceleration. This adds wrappers to the jit functions to allow input and output of any dimensional array of Rotations.

Reasons for this:
1) This makes it possible to add to/from functions for Euler, Rodrigues, homochoric, cubochoric, and axis-angle functions into either Rotations or Quaternions. without losing the n-dimensional object support.
2) all the to/from functions will be able to use the jit-accelerated conversion functions copied directly from the Rowenhorst paper. 
3) Orix contains repeated to/from conversion functions, which adds some bloat and gives more room for errors. Wrapping these functions makes it easier to replace this repeated functionality with the jit-accelerated ones.

4)  This update makes it easier to take care of issues 
  #374
  #373 
  #346
  #345  and 
   #254

#### Progress of the PR

The test cases verify the new functions produce identical results to the old 2d versions, but here is some code I used to verify this new code can, in fact, accept n-dimensional arrays of Rotations, and does not slow down the execution of the code by a meaningful amount.

```python
import numpy as np
import time
from orix.quaternion._conversions import ro2ax, ro2ax2d

tocA = 0
tocB = 0
tocC = 0

# Run 10 times to get a good average
for i in np.arange(10):
    rod = np.random.rand(1000000, 4)

    # Convert Rod to axangle the old way using the original function
    tic = time.time()
    out_old = ro2ax2d(rod)
    toc = time.time() - tic
    tocA += toc

    # Do it using the new n-dimensional wrapper
    tic = time.time()
    out_new = ro2ax(rod)
    toc = time.time() - tic
    tocB += toc

    # verify they are identical
    assert np.all(out_new == out_old)

    # Do it using the new n-dimensional wrapper on a weird shaped dataset
    rod_weird = rod.reshape(2, 5, 10, 1, 100, 10, 10, 4)
    tic = time.time()
    out_weird = ro2ax(rod_weird).reshape(rod.shape)
    toc = time.time() - tic
    tocC += toc

    # verify they are identical
    assert np.all(out_old == out_weird)

# check out how long each method took
print("old way:{0:0.04f}".format(tocA))
print("new way:{0:0.04f}".format(tocB))
print("new way weird shape:{0:0.04f}".format(tocC))
```

Heres the output I got:
old way:6.9345
new way:6.9778
new way weird shape:6.9356

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/quaternion/_conversions.py
+++ b/orix/quaternion/_conversions.py
@@ -67,6 +67,34 @@ def get_pyramid_single(xyz):
         return 6
 
 
+@nb.jit("int64[:](float64[:, :])", cache=True, nogil=True, nopython=True)
+def get_pyramid(xyz):
+    """Determine to which out of six pyramids in the cube a 2D array of
+    (x, y, z) coordinates belongs.
+
+    Parameters
+    ----------
+    xyz : numpy.ndarray
+        2D array of n (x, y, z) as 64-bit floats.
+
+    Returns
+    -------
+    pyramid : int
+        1D array of length n.
+        Which pyramid `xyz` belongs to as a 64-bit integer.
+
+    Notes
+    -----
+    This function is optimized with Numba, so care must be taken with
+    array shapes and data types.
+    """
+    n_scalars = xyz.shape[0]
+    pyramids = np.zeros(n_scalars, dtype=np.int64)
+    for i in nb.prange(n_scalars):
+        pyramids[i] = get_pyramid_single(xyz[i])
+    return pyramids
+
+
 @nb.jit("float64[:](float64[:])", cache=True, nogil=True, nopython=True)
 def cu2ho_single(cu):
     """Conversion from a single set of cubochoric coordinates to
@@ -561,4 +589,31 @@ def eu2qu_single(alpha, beta, gamma):
     if qu[0] < 0:
         qu = -qu
 
+    return qu
+
+
+@nb.jit("float64[:, :](float64[:, :])", cache=True, nogil=True, nopython=True)
+def eu2qu(eu):
+    """Conversion from multiple Euler angles (alpha, beta, gamma) to unit
+    quaternions
+
+    Parameters
+    ----------
+    eu : numpy.ndarray
+        2D array of n (alpha, beta, gamma) as 64-bit floats.
+
+    Returns
+    -------
+    qu : numpy.ndarray
+        2D array of n (qo, q1, q2, q3) quaternions as 64-bit floats.
+
+    Notes
+    -----
+    This function is optimized with Numba, so care must be taken with
+    array shapes and data types.
+    """
+    n_vectors = eu.shape[0]
+    qu = np.zeros((n_vectors, 4), dtype=np.float64)
+    for i in nb.prange(n_vectors):
+        qu[i] = eu2qu_single(eu[i, 0], eu[i, 1], eu[i, 2])
     return qu
--- a/orix/quaternion/_conversions.py
+++ b/orix/quaternion/_conversions.py
@@ -16,7 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with orix.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Conversions of orientations between many common representations from
+"""Conversions of rotations between many common representations from
 :cite:`rowenhorst2015consistent`, accelerated with Numba.
 
 This module and documentation is only relevant for orix developers, not
@@ -203,7 +203,7 @@ def cu2ho_2d(cu):
     This function is optimized with Numba, so care must be taken with
     array shapes and data types.
     """
-    ho = np.zeros_like(cu)
+    ho = np.zeros_like(cu, dtype=np.float64)
     for i in nb.prange(cu.shape[0]):
         ho[i] = cu2ho_single(cu[i])
     return ho
@@ -361,7 +361,7 @@ def ax2ro_2d(ax):
     This function is optimized with Numba, so care must be taken with
     array shapes and data types.
     """
-    ro = np.zeros_like(ax)
+    ro = np.zeros_like(ax, dtype=np.float64)
     for i in nb.prange(ax.shape[0]):
         ro[i] = ax2ro_single(ax[i])
     return ro
--- a/orix/quaternion/_conversions.py
+++ b/orix/quaternion/_conversions.py
@@ -80,7 +80,7 @@ def get_pyramid_2d(xyz):
     Returns
     -------
     pyramids : numpy.ndarray
-        1D array of 2hich pyramids ``xyz`` belong to as 64-bit integers.
+        1D array of pyramids as 64-bit integers.
 
     Notes
     -----
@@ -95,11 +95,11 @@ def get_pyramid_2d(xyz):
 
 
 def get_pyramid(xyz):
-    """n-dimensional wrapper for get_pyramid_2d, see the docstring of that
-    function.
+    """n-dimensional wrapper for get_pyramid_2d, see the docstring of
+    that function.
     """
     n_xyz = np.prod(xyz.shape[:-1])
-    xyz2d = xyz.reshape(n_xyz, 3)
+    xyz2d = xyz.astype(np.float64).reshape(n_xyz, 3)
     pyramids = get_pyramid_2d(xyz2d).reshape(n_xyz)
     return pyramids
 
@@ -214,7 +214,7 @@ def cu2ho(cu):
     function.
     """
     n_cu = np.prod(cu.shape[:-1])
-    cu2d = cu.reshape(n_cu, 3)
+    cu2d = cu.astype(np.float64).reshape(n_cu, 3)
     ho = cu2ho_2d(cu2d).reshape(cu.shape)
     return ho
 
@@ -282,8 +282,8 @@ def ho2ax_2d(ho):
 
     Returns
     -------
-    ho : numpy.ndarray
-        2D array of n (x, y, z) as 64-bit floats.
+    ax : numpy.ndarray
+        2D array of n (x, y, z, angle) as 64-bit floats.
 
     Notes
     -----
@@ -302,7 +302,7 @@ def ho2ax(ho):
     function.
     """
     n_ho = np.prod(ho.shape[:-1])
-    ho2d = ho.reshape(n_ho, 3)
+    ho2d = ho.astype(np.float64).reshape(n_ho, 3)
     ho = ho2ax_2d(ho2d).reshape(ho.shape[:-1] + (4,))
     return ho
 
@@ -372,7 +372,7 @@ def ax2ro(ax):
     function.
     """
     n_ax = np.prod(ax.shape[:-1])
-    ax2d = ax.reshape(n_ax, 4)
+    ax2d = ax.astype(np.float64).reshape(n_ax, 4)
     ro = ax2ro_2d(ax2d).reshape(ax.shape)
     return ro
 
@@ -438,7 +438,7 @@ def ro2ax(ro):
     function.
     """
     n_ro = np.prod(ro.shape[:-1])
-    ro2d = ro.reshape(n_ro, 4)
+    ro2d = ro.astype(np.float64).reshape(n_ro, 4)
     ax = ro2ax_2d(ro2d).reshape(ro.shape)
     return ax
 
@@ -503,7 +503,7 @@ def ax2qu(ax):
     function.
     """
     n_ax = np.prod(ax.shape[:-1])
-    ax2d = ax.reshape(n_ax, 4)
+    ax2d = ax.astype(np.float64).reshape(n_ax, 4)
     qu = ax2qu_2d(ax2d).reshape(ax.shape)
     return qu
 
@@ -563,7 +563,7 @@ def ho2ro(ho):
     function.
     """
     n_ho = np.prod(ho.shape[:-1])
-    ho2d = ho.reshape(n_ho, 3)
+    ho2d = ho.astype(np.float64).reshape(n_ho, 3)
     ro = ho2ro_2d(ho2d).reshape(ho.shape[:-1] + (4,))
     return ro
 
@@ -626,7 +626,7 @@ def cu2ro(cu):
     function.
     """
     n_cu = np.prod(cu.shape[:-1])
-    cu2d = cu.reshape(n_cu, 3)
+    cu2d = cu.astype(np.float64).reshape(n_cu, 3)
     ro = cu2ro_2d(cu2d).reshape(cu.shape[:-1] + (4,))
     return ro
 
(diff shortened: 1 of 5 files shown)


hakonanes:
Thanks for this @argerlt! It is a nice generalization of existing functionality.

> By default, Rotation objects can contain n-dimensional arrays of Rotation data. However, _conversions.py is written to only accept 2D arrays, since the functions use jit for acceleration.

The conversions are private, not ment to be used by a user. But, with you mentioning them in #374 and opening this PR, I think it is time to make most of them public as methods to `Quaternion`. I suggest to leave this for another PR.

orix/quaternion/_conversions.py
```diff
@@ -524,8 +607,16 @@ def cu2ro(cu):
...
+def cu2ro(cu):
+    """n-dimensional wrapper for cu2ro2d"""
+    n_cu = np.prod(cu.shape[:-1])
+    cu2d = cu.reshape(n_cu, 3)
+    ro = cu2ro2d(cu2d).reshape(cu.shape[:-1] + (4,))
+    return ro
+
+
+@nb.jit("float64[:](float64[:])", cache=True, nogil=True, nopython=True)
+def eu2qu_single(eu):
```
hakonanes, line 619: The `Parameters` section must be updated.

orix/quaternion/_conversions.py
```diff
@@ -547,18 +638,53 @@ def eu2qu_single(alpha, beta, gamma):
...
+    delta = 0.5 * np.subtract(eu[0], eu[2])
+    c = np.cos(eu[1] / 2)
+    s = np.sin(eu[1] / 2)
 
     qu = np.zeros(4, dtype=np.float64)
-    qu[0] = c * np.cos(sigma)
-    qu[1] = -s * np.cos(delta)
-    qu[2] = -s * np.sin(delta)
-    qu[3] = -c * np.sin(sigma)
+    qu[0] = np.array(c * np.cos(sigma), dtype=np.float64)
```
hakonanes, line 647: Is making these arrays necessary?

orix/quaternion/_conversions.py
```diff
@@ -547,18 +638,53 @@ def eu2qu_single(alpha, beta, gamma):
...
+    """
+    n_vectors = eu.shape[0]
+    qu = np.zeros((n_vectors, 4), dtype=np.float64)
+    for i in nb.prange(n_vectors):
+        qu[i] = eu2qu_single(eu[i])
+    return qu
+
+
+def eu2qu(eu):
+    """n-dimensional wrapper for eu2qu2d"""
```
hakonanes, line 686:
```suggestion
    """N-dimensional wrapper for eu2qu_2d, see the docstring of that
    function.
    """
```

orix/quaternion/_conversions.py
```diff
@@ -250,6 +294,14 @@ def ho2ax(ho):
     return ax
 
 
+def ho2ax(ho):
+    """n-dimensional wrapper for ho2ax2d"""
```
hakonanes, line 298:
```suggestion
    """N-dimensional wrapper for ho2ax_2d, see the docstring of that
    function.
    """
```

orix/quaternion/_conversions.py
```diff
@@ -547,18 +638,53 @@ def eu2qu_single(alpha, beta, gamma):
...
+
+    Parameters
+    ----------
+    eu : numpy.ndarray
+        2D array of n (alpha, beta, gamma) as 64-bit floats.
+
+    Returns
+    -------
+    qu : numpy.ndarray
+        2D array of n (qo, q1, q2, q3) quaternions as 64-bit floats.
```
hakonanes, line 671:
```suggestion
        2D array of n (q0, q1, q2, q3) quaternions as 64-bit floats.
```

orix/tests/quaternion/test_conversions.py
```diff
@@ -133,75 +144,101 @@ def quaternions_conversions():
 class TestRotationConversions:
     def test_rotation_from_euler(self):
         euler = np.array([1, 2, 3])
+        eulers = np.arange(1, 2.5, 0.05).reshape(10, 3)
```
hakonanes, line 147: I suggest to use either `euler` or `eulers` and not both in this test. Is there a reason for these particular values? If not, I suggest to create the angles in a simpler way, like `np.arange(30).reshape(10, 3) / 30`.

hakonanes (changes requested):
Thanks for making these generalizations! Some comments on function naming and docstrings, mostly. Please say so if anything is unclear or you disagree with something.

orix/quaternion/_conversions.py
```diff
@@ -67,6 +67,42 @@ def get_pyramid_single(xyz):
         return 6
 
 
+@nb.jit("int64[:](float64[:, :])", cache=True, nogil=True, nopython=True)
+def get_pyramid2d(xyz):
```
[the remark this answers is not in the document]
argerlt: This occurred to me yesterday actually: would it be better to call these functions `get_pyramid_vector` or `get_pyramid_vectorized`? after all, they are 2D numpy arrays, but not 2D arrays of objects. calling them 2d might be confusing down the line.
Could also leave as-is, or change to some different suggested name.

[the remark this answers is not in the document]
hakonanes: I think the 2D naming is clear, as in the array has two dimensions. These functions will only be available in the public API via `Quaternion` methods anyway, where the shape agnostic version will be used.

orix/quaternion/_conversions.py
```diff
@@ -172,6 +208,14 @@ def cu2ho(cu):
     return ho
 
 
+def cu2ho(cu):
+    """n-dimensional wrapper for cu2ho2d"""
```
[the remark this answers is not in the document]
argerlt: Follow up on this comment and the comments on all functions like it: 
I originally thought about copy/pasting the docstrings from `the _single` and `_2d` functions into the wrappers, but felt that doing so would result in docstrings repeated in triplicate, which felt unnecessary. I could either:

A) do exactly what you are suggesting here and be done
B) move the verbose docstring to the wrappers and write minimalized versions in the `_single` and `_2d` functions (would make the file shorter as a bonus), or
C) just be verbose and make all three functions have full sized docstrings, so help always gives all the info you need.

Thoughts?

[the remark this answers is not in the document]
hakonanes: I'd go with (A). None of the functionality in `_conversions.py` is user facing (included in [the API reference](https://orix.readthedocs.io/en/stable/reference.html)), so we only need to supply docstrings for developers.

hakonanes:
@argerlt, do you intend to use these functions outside of orix? If so, we should include them in the API reference, and find a convenient way to import them.

orix/quaternion/_conversions.py
argerlt: Yes, or at least something similar is.

Without these arrays, I was getting a problem where the right hand sides of lines 647-650 would produce float32 values that would then be inserted into a float64 array, resulting in incorrect values. This was a jit-only problem, because the .py_func versions of the functions execute correctly (float64 in, float64 out).

I tried several variations, which all gave the incorrect results:
```
qu[3] = -c * np.sin(sigma)
qu[3] = -c * np.sin(sigma, dtype = np.float64)
qu[3] = np.multiply(c, np.sin(sigma))
```
etc.
This FEELS weird and wrong, but it was the only method that worked which didn't involve adding extra variables, which seemed to slow down the computation.
If you or anyone else think of a more elegant solution, feel free to change it.

orix/tests/quaternion/test_conversions.py
argerlt: I agree on both points.

argerlt:
> @argerlt, do you intend to use these functions outside of orix? If so, we should include them in the API reference, and find a convenient way to import them.

I didn't plan to. I personally cannot think of any situation where I would want to use these conversions where I wouldn't also just use the Orix Quaternion Class.  

Again, using scipy as a point of reference, I often have to go from active euler to Rodrigues-frank. When doing so, I never access those conversion directly. Instead, I just do:

```
from scipy.spatial.transform import Rotation as R
rod = R.from_euler('ZXZ',[euler_angles_here]).as_mrp()
```
Which I think is conceptually simple and keeps the API clean.

orix/quaternion/_conversions.py
```diff
@@ -67,6 +67,43 @@ def get_pyramid_single(xyz):
...
+
+    Parameters
+    ----------
+    xyz : numpy.ndarray
+        2D array of n (x, y, z) as 64-bit floats.
+
+    Returns
+    -------
+    pyramids : numpy.ndarray
+        1D array of 2hich pyramids ``xyz`` belong to as 64-bit integers.
```
hakonanes, line 83:
```suggestion
        1D array of which pyramids ``xyz`` belong to as 64-bit integers.
```

orix/quaternion/_conversions.py
```diff
@@ -524,16 +621,26 @@ def cu2ro(cu):
...
     """Convert three Euler angles (alpha, beta, gamma) to a unit
     quaternion.
 
     Parameters
     ----------
-    alpha, beta, gamma : float
-        Euler angles in the Bunge convention in radians as 64-bit
-        floats.
+    eu : numpy.ndarray
+        1D array of (alpha, beta, gamma) Euler angles given in radians in the
```
hakonanes, line 642: Keep comments within 72 characters.

orix/quaternion/_conversions.py
hakonanes: Can you do `qu[0] = float(c * np.cos(sigma))`? If not, try `qu[0] = np.float64(c * np.cos(sigma))`.

argerlt:
Alright, assuming you have no further feedback on line 647:
> ```[0] = np.array(c * np.cos(sigma), dtype=np.float64)```

I believe I have addressed all the required changes. let me know if that isn't the case

hakonanes (changes requested):
I agree that a syntax like `Quaternion.from_euler(euler).to_matrix()`etc. is sufficient.

This shapes up rather nicely, thank you very much for addressing all my review comments! I've just got a few of them left.

Just as a heads-up: I prefer to resolve review comments myself, since this makes it easier to decided whether a change has been addressed.

orix/tests/quaternion/test_conversions.py
```diff
@@ -132,76 +143,101 @@ def quaternions_conversions():
 
 class TestRotationConversions:
     def test_rotation_from_euler(self):
-        euler = np.array([1, 2, 3])
```
hakonanes, line 135: One line is not covered anymore (see [coveralls log](https://coveralls.io/builds/51847181/source?filename=orix%2Fquaternion%2F_conversions.py#L669)), I think we should include this rotation in this test.

argerlt: I was confused why this popped up no but not before. Is this because I got rid of `euler` and replaced it with `eulers[0]`?

orix/quaternion/_conversions.py
argerlt: np.float64 passes correctly

argerlt: actually, nevermind, np.float64 doesn't work either. I was loading an old __pycache__ file. I think np.array might be the easiest fix to this.

hakonanes: OK, I'll have a look myself.

argerlt: Huh. Okay, this has be totally confused. 
For some reason, every variation of np.float64, float, etc, breaks. However, it ONLY breaks if the value it is inserting into qu is less than zero. 

I was going to suggest we just approve this and move on, but the underlying implication here is SOMEWHERE, things are getting cast incorrectly, possibly from float64 to float32 and then back, which would translate to a big loss in accuracy. 

I also tried explicitly using numpy multiply, divide, etc, like this:
```
    sigma = np.multiply(0.5, np.add(eu[0], eu[2]))
    delta = np.multiply(0.5, np.subtract(eu[0], eu[2]))
    c = np.cos(np.divide(eu[1], 2))
    s = np.sin(np.divide(eu[1], 2))
    nc = np.multiply(-1, c)
    ns = np.multiply(-1, s)

    qu = np.zeros(4, dtype=np.float64)
    qu[0] = np.multiply(c, np.cos(sigma))
    qu[1] = np.multiply(ns, np.cos(delta))
    qu[2] = np.multiply(ns, np.sin(delta))
    qu[3] = np.multiply(nc, np.sin(sigma))

    if qu[0] < 0:
        qu = -qu

```
but that failed as well.

argerlt:
Most recent push includes all the suggested formatting changes, plus a few I caught on reread.

I also realized that if any of the _conversion.py functions got passed data that wasn't float64, they would fail. Added lines to the wrapper functions to convert inputs to float64, as well as test cases that would pass float32 data to the checks, just to ensure this doesn't pop back up later. 

In reference to the eu2qu in _conversions.py, lines 657-671:
I honestly don't know what is happening here, but if I set atol=1e-40 in test_conversions.py, the test still succeeds, so it seems there is not loss of accuracy with this method, and also no loss of speed. I am ready to just accept this as a "it works and it's correct, so lets move on", but if you can figure out how to write it in a different way, feel free to change it directly.

orix/quaternion/_conversions.py
hakonanes: I agree that it's OK to leave as is and move on.

Out of curiosity, however, which OS and Python version are you testing with? All `pytest -k test_conversions` pass without any changes to the original code on my Ubuntu 22.04 with Python 3.9. Are you running other tests than the ones in `test_convers.py`? If so, we should similar tests to catch the behaviour you're describing.

hakonanes:
I'm including "Austin Gerlt" in `__credits__` in `orix/__init__.py`, and also in `.zenodo.json` with the following metadata:

```json
    {
      "name": "Austin Gerlt",
      "orcid": "0000-0002-2204-2055",
      "affiliation": "The Ohio State University"
    },
```

The ORCID is [here](https://orcid.org/0000-0002-2204-2055) and your affiliation I found [here](https://people.engineering.osu.edu/people/gerlt.1). If you're happy with this, I'll push these changes (and some minor formatting) to your branch before merging.

argerlt:
Yup, works for me. Change what you need.

hakonanes:
Sorry @argerlt, it seems I cannot push commits to your branch `ndim_convert`. Do you allow commits by maintainers, i.e. have the box at the bottom of the PR's right sidebar named "Allow edits and access to secrets by maintainers" ticked?

hakonanes (approved):
Thank you for contributing these nice generalizations, and for addressing all my review comments, @argerlt!

Merging when checks pass.

Could you briefly comment on https://github.com/pyxem/orix/pull/376/files#r951924756?

orix/quaternion/_conversions.py
argerlt: I'm running on a Windows 10 computer with an AMD CPU. specifically from command prompt, When testing, I was running `pytest orix/tests/quaternion/test_conversions.py`

Since you brought this up, I tried it again, same code, same computer, but using Windows Subsystem for Linux (WSL) to run pytest. It passed without needing the np.array() additions! So, I assume this is a windows specific problem with numba and jit, which honestly, makes it even weirder.

hakonanes: Thank you for clarifying. I've encountered surprising effects with Numba in the past, similar to your experience here. This is just a reminder that we should not trust Numba blindly!

PR #392, opened 2022-09-28
Deprecate minimal/brittle "3D" functionality for crystal maps

hakonanes:
#### Description of the change
See #390.

~I have not filtered all the warnings raised by these changes in the tests, because I plan to remove the functionality in the `develop` branch after the changes for 0.10.1 are merged into `develop`, at which point the warnings will be removed as well.~ Ended up doing this so that more important warnings (relating to Matplotlib) are easier to catch.

#### Progress of the PR
- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [x] Unit tests with pytest for all lines
- [x] Clean code style by [running black via pre-commit](https://orix.readthedocs.io/en/latest/contributing.html#code-style)
- [x] Filter warnings raised from internal use
- [ ] Merge when #391 and #393 are approved

#### For reviewers

- [ ] The PR title is short, concise, and will make sense 1 year later.
- [ ] New functions are imported in corresponding `__init__.py`.
- [ ] New features, API changes, and deprecations are mentioned in the unreleased
      section in `CHANGELOG.rst`.
- [ ] Contributor(s) are listed correctly in `__credits__` in `orix/__init__.py` and in
      `.zenodo.json`.

PR #391, opened 2022-09-28
Fix indexing/slicing into sliced CrystalMap

hakonanes:
#### Description of the change
I discovered a bug where indexing/slicing into an already sliced `CrystalMap` assumed the initial shape (`_original_shape`) and not the current shape.

#### Progress of the PR
- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [x] Unit tests with pytest for all lines
- [x] Clean code style by [running black via pre-commit](https://orix.readthedocs.io/en/latest/contributing.html#code-style)
- [ ] Update release version to 0.10.1
- [ ] Merge after #392 and #393 are merged

#### Minimal example of the bug fix or new feature

```python
>>> from orix.crystal_map import CrystalMap
>>> xmap = CrystalMap.empty((4, 5))
>>> xmap.id.reshape(xmap.shape)
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14],
       [15, 16, 17, 18, 19]])
>>> xmap2 = xmap[1:4, :3]
>>> xmap2._original_shape
(4, 5)
>>> xmap2.id.reshape(xmap2.shape)
array([[ 5,  6,  7],
       [10, 11, 12],
       [15, 16, 17]])
```

Getting some IDs of `xmap2` before this fix

```python
>>> xmap2[1, 1].id
array([11])
>>> xmap2[:2, :2].id
array([5, 6])
```

and after this fix
```python
>>> xmap2[1, 1].id
array([13])
>>> xmap2[:2, :2].id
array([ 5,  6, 10, 11])
```

#### For reviewers

- [ ] The PR title is short, concise, and will make sense 1 year later.
- [ ] New functions are imported in corresponding `__init__.py`.
- [ ] New features, API changes, and deprecations are mentioned in the unreleased
      section in `CHANGELOG.rst`.
- [ ] Contributor(s) are listed correctly in `__credits__` in `orix/__init__.py` and in
      `.zenodo.json`.

Issue #390, opened 2022-09-28
Patch release 0.10.1


hakonanes:
I discovered a bug where indexing/slicing into an already sliced `CrystalMap` assumed the initial shape (`_original_shape`) and not the current shape. I'd like to make a patch release 0.10.1.

It's not according to semantic versioning, but I'd like to add deprecation warnings for all references to properties and parameters refering to the z coordinate of a CrystalMap in 0.10.1 as well, with a plan to remove the "functionality" in 0.11. I made a framework for a 3D crystal map when first writing the `CrystalMap` class, but it is brittle, I don't use it myself (and assume nobody does) and it makes adding functionality to `CrystalMap` unnecessarily challenging.

I plan to fix #380 for 0.10.1 as well.

If there aren't any objections to the release and these changes, I'll open separate PRs for each of the three changes, merge them in turn when they are all ready, and then release.

#### Checklist (added by @pc494)

- [x] #391 
- [x] #392 
- [x] #393 
- [x] #394 
- [ ] #395 (post release)

pc494:
I'm fine with "non-semantic" addition of deprecations in patch releases.

hakonanes:
Thank you, Phillip.

All the four open PRs should go into 0.10.1 after they are reviewed, then. I'll clean up the changelog and bump the release version in #391 once the others are merged onto `main`.

pc494:
Feel free to review request me on the other PRs

hakonanes:
Thanks, I’ve done that now.

pc494:
Are you happy to make the release @hakonanes?

hakonanes:
Yes, thank you for the reviews! I just have to bump the version to 0.10.1.

hakonanes:
0.10.1 is now available on PyPI. I'd like to complete the [release steps](https://github.com/pyxem/orix/blob/develop/RELEASE.rst#post-release-action) for 0.10.1 before yanking 0.10.0 from PyPI.

hakonanes:
0.10.1 is now available from `conda-forge` (for Python >= 3.7 only). So, I'll yank 0.10.0 from PyPI.

PR #393, opened 2022-09-28
Allow to pass c/color and s/sizes to StereographicPlot.scatter()

### LINKED ISSUE #380: StereographicPlot.scatter() does not allow passing color parameter
`StereographicPlot.scatter()` throws the following error when passing the `color` parameter:

```python
>>> Vector3d.xvector().scatter(color=(1, 0, 0))  # Red scatter point
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [132], in <cell line: 2>()
      1 v = Vector3d.xvector()
----> 2 v.scatter(color=[1, 0, 0])

File ~/kode/orix/orix/vector/vector3d.py:1123, in Vector3d.scatter(self, projection, figure, axes_labels, vector_labels, hemisphere, reproject, show_hemisphere_label, grid, grid_resolution, figure_kwargs, reproject_scatter_kwargs, text_kwargs, return_figure, **kwargs)
   1121 for i, ax in enumerate(axes):  # Assumes a maximum of two axes
   1122     ax.hemisphere = hemisphere[i]
-> 1123     ax.scatter(self, **kwargs)
   1124     if reproject:
   1125         ax.scatter(v_reprojected, **reproject_scatter_kwargs)

File ~/kode/orix/orix/plot/stereographic_plot.py:212, in StereographicPlot.scatter(self, *args, **kwargs)
    209 if s is not None:
    210     s = _get_array_of_values(value=s, visible=visible)
--> 212 super().scatter(x, y, c=c, s=s, **updated_kwargs)
[...]
ValueError: Supply a 'c' argument or a 'color' kwarg but not both; they differ but their functionalities overlap.
```

The reason can be seen  in this line above `--> 212 super().scatter(x, y, c=c, s=s, **updated_kwargs)`, where `c` is always passed.

I think the solution is to not pass `c` if `color` is among the keyword arguments.

hakonanes:

Fixes #380.

It is only in `StereographicPlot.scatter()` we hard-coded in passing `c` and `s` to `matplotlib.axes.Axes.scatter()`, so it is only there this is handled.

See #390.

#### Progress of the PR
- [ ] Merge when #391 and #392 are approved

The following raised errors before, but now works

```python
>>> from orix import vector
>>> v = vector.Vector3d([[1, 1, 1], [1, 0, 0]])
>>> v.scatter(color=["C0", "C1"])  # Equivalent to passing c=["C0", "C1"]
>>> v.scatter(sizes=[10, 20])  # Equivalent to passing s=[10, 20]
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/plot/stereographic_plot.py
+++ b/orix/plot/stereographic_plot.py
@@ -202,10 +202,20 @@ class StereographicPlot(maxes.Axes):
         if x.size == 0:
             return
 
-        # Color(s) and size(s)
-        c = updated_kwargs.pop("c", "C0")
+        # Color(s)
+        if "color" in updated_kwargs.keys():
+            key_color = "color"
+        else:
+            key_color = "c"
+        c = updated_kwargs.pop(key_color, "C0")
         c = _get_array_of_values(value=c, visible=visible)
-        s = updated_kwargs.pop("s", None)
+
+        # Size(s)
+        if "sizes" in updated_kwargs.keys():
+            key_size = "sizes"
+        else:
+            key_size = "s"
+        s = updated_kwargs.pop(key_size, None)
         if s is not None:
             s = _get_array_of_values(value=s, visible=visible)
 
@@ -333,8 +343,6 @@ class StereographicPlot(maxes.Axes):
             sigma=sigma,
             log=log,
             hemisphere=self.hemisphere,
-            symmetry=None,
-            mrd=True,
             weights=weights,
         )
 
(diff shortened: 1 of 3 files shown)


orix/plot/stereographic_plot.py
```diff
@@ -202,10 +202,20 @@ def scatter(
...
+        if "color" in updated_kwargs.keys():
+            key_color = "color"
+        else:
+            key_color = "c"
+        c = updated_kwargs.pop(key_color, "C0")
         c = _get_array_of_values(value=c, visible=visible)
-        s = updated_kwargs.pop("s", None)
+
+        # Size(s)
+        if "sizes" in updated_kwargs.keys():
```
harripj, line 214: I think people would be more likely to use size than sizes, and is more consistent with MPL.

harripj (approved):
Looks good to me! I have just one comment for consideration.

orix/plot/stereographic_plot.py
hakonanes: I agree, but `sizes` is the parameter supported in `matplotlib.axes.Axes.scatter()`, while `size` is not and throws an `AttributeError`.

harripj: Then consider this point mute and thanks for the information!

hakonanes:
Since #394 is already merged, I'll just merge this as well.

PR #347, opened 2022-05-18
Bump oldest supported Python version to 3.7

harripj:
#### Description of the change
As discussed in #343 Python 3.6 is deprecated, but currently used in CI as the oldest supported version. This is preventing implementation of type hinting (as discussed in #338) into the code base as `annotations` from `__future__` is only available in Python 3.7+.

In this PR the oldest Python version used in CI is bumped to 3.7. A guide for implementing type hints is introduced into the contributing guide.

#### Progress of the PR
- [n/a] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [n/a] Unit tests with pytest for all lines
- [n/a] Clean code style by [running black via pre-commit](https://orix.readthedocs.io/en/latest/contributing.html#code-style)

#### For reviewers

- [x] The PR title is short, concise, and will make sense 1 year later.
- [n/a] New functions are imported in corresponding `__init__.py`.
- [x] New features, API changes, and deprecations are mentioned in the unreleased
      section in `CHANGELOG.rst`.
- [x] Contributor(s) are listed correctly in `__credits__` in `orix/__init__.py` and in
      `.zenodo.json`.

PR #394, opened 2022-09-29
Add `python_requires>=3.7` packaging metadata

sk1p:

As a follow-up to #347, actually prevent installing on incompatible Python versions when using pip.

Note that currently, in a Python 3.6 environment, `pip install orix` will install an incompatible version (0.10.0) of orix. This is caused by the missing `python_requires` metadata - without it, the package is seen as compatible with any Python version.

Merging and releasing this won't magically fix things for 0.10.0 though - without [yanking that release](https://pypi.org/help/#yanked), installing `orix` in an old Python 3.6 environment will continue to pick 0.10.0 and fail at runtime. But adding this and updating once 3.7 support is dropped means the same issue won't come up again.

I'll leave the decision about yanking to the maintainers, of course.

#### Progress of the PR

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -32,6 +32,11 @@
     },
     {
       "name": "Simon Høgås"
+    },
+    {
+      "name": "Alexander Clausen",
+      "orcid": "0000-0002-9555-7455",
+      "affiliation": "Jülich Research Centre, Ernst Ruska Centre"
     }
   ]
 }
--- a/orix/__init__.py
+++ b/orix/__init__.py
@@ -13,4 +13,5 @@ __credits__ = [
     "Niels Cautaerts",
     "Austin Gerlt",
     "Simon Høgås",
+    "Alexander Clausen",
 ]
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,7 @@ setup(
         "Topic :: Scientific/Engineering",
         "Topic :: Scientific/Engineering :: Physics",
     ],
+    python_requires=">=3.7",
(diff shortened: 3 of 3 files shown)


harripj:
@sk1p thanks for raising this and providing a fix. I would be in favour of yanking the `0.10` release once `0.10.1` is released, as python 3.6 is unsupported for this release. What do you think @hakonanes @pc494 ?

hakonanes:
Thank you very much for bringing this to our attention, @sk1p. Yes, we should yank 0.10.0 once 0.10.1 is out.

@sk1p, since this is a bug fix, can you branch of `main` instead and make a PR to that branch instead?

sk1p:
> since this is a bug fix, can you branch of `main` instead and make a PR to that branch instead?

Done! Nowadays possible via the edit button at the top btw.

sk1p:
> Done! Nowadays possible via the edit button at the top btw.

I guess I spoke too soon, as it included some unrelated changes. I'll fix this in a bit.

hakonanes:
If it's OK with you @sk1p I'll push some updates to your branch adding you to the credits (taking relevant Zenodo metadata from the same entry in LiberTEM) and add a changelog entry.

sk1p:
> If it's OK with you @sk1p I'll push some updates to your branch adding you to the credits (taking relevant Zenodo metadata from the same entry in LiberTEM) and add a changelog entry.

Sure, thanks!

Issue #367, opened 2022-06-17
Fix inherited methods in the Miller class


hakonanes:
The `Miller` class inherits methods and properties from `Vector3d` and thus also `Object3d`. Currently, these methods and properties aren't handled properly:

`Vector3d`
* `empty()`
* `from_polar()`
* `get_circle()`
* `get_nearest()` (should raise `NotImplementedError`)
* `rotate()`
* `xvector()`
* `yvector()`
* `zero()`
* `zvector()`
* `perpendicular`

`Object3d`
* `get_random_sample()`
* `squeeze()`
* `stack()`

It's not critical to get this in as a patch release, releasing it in 0.10 is fine.

PR #401, opened 2022-10-09
Add class method `from_align_vectors()` to Quaternion, Misorientation and Orientation

anderscmathisen:

This PR is a result of an ongoing  recent [discussion](https://github.com/pyxem/orix/discussions/400).

I have made functions to wrap `scipy.spacial.transform.Rotation.align_vectors()` to initialize Quaternion, Misorientation or Orientation objects which rotates a set of vectors b into another set a. For the Misorientation and Orientation objects, symmetry arguments are also accepted.

For the Quaternion class, I have also made a function which initializes a Quaternion object from a Scipy Rotation object. 

The implementation of multiple return values in my code is not very elegant, and suggestions for making it better is appreciated.

In the discussion, the conversion from Scipy quaternions to Orix quaternions is discussed. I did look a little into this, and got very confused, but the current conversion in `Quaternion.from_scipy_rotation()` seems to work in the sense that the resulting orientation from the function `Quaternion.from_align_vectors()` indeed transforms vectors from b to a.

#### Progress of the PR
- [X (ish) ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [X] Clean code style by [running black via pre-commit](https://orix.readthedocs.io/en/latest/contributing.html#code-style)

```python

>>> from orix.vector import Vector3d, Miller
>>>
>>> crystal_millers = Miller([[2,-1,0],[0,0,1]])
>>> sample_vectors = Vector3d([[3,1,0],[-1,3,0]])
>>>
>>> ori = Orientation.from_align_vectors(crystal_millers, sample_vectors)
>>> print(crystal_millers)
>>> print(ori * sample_vectors.unit * crystal_millers.length)

```

Output:

```
Miller (2,), point group None, xyz
[[ 2 -1  0]
 [ 0  0  1]]
Vector3d (2,)
[[ 2. -1. -0.]
 [-0. -0.  1.]]
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


orix/quaternion/quaternion.py
```diff
@@ -498,3 +499,66 @@ def _outer_dask(
 
         new_chunks = tuple(chunks1[:-1]) + tuple(chunks2[:-1]) + (-1,)
         return out.rechunk(new_chunks)
+
+    @classmethod
+    def from_scipy_rotation(cls, scipy_rotation: R) -> Quaternion:
```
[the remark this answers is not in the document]
hakonanes: Also, best to rename `x, y, z, w` to `b, c, d, a` in line with other naming of those parameters in orix.

hakonanes (changes requested):
Thank you for taking the time to set up this PR, @anderscmathisen!

This first review is just of the quaternion methods. Will come back to the others once I've found some tests for them to convince myself of their correctness. Then we can write some tests and add some examples to the docstrings.

I find your implementation in good shape. I have many smaller suggestions for changes to keep the code style of the rest of the orix code base ([see the code style guide](https://orix.readthedocs.io/en/stable/contributing.html#code-style)) and some clarifications of the docstrings. Feel free to question my comments!

hakonanes (changes requested):
Thank you for resolving most of my comments. Below are a few more comments.

I'm confident that `Orientation.from_align_vectors()` is OK, based on the following test, which we could simplify and add to the test suite (as multiple smaller unit tests):

```python
from diffpy.structure import Lattice, Structure
import numpy as np
from orix.vector import Miller, Vector3d


phases = {
    "cubic": Phase(point_group="m-3m"),
    "hexagonal": Phase(
        point_group="6/mmm",
        structure=Structure(lattice=Lattice(1, 1, 2, 90, 90, 120))
    ),
    "trigonal": Phase(
        point_group="-3m",
        structure=Structure(lattice=Lattice(4.9, 4.9, 5.4, 90, 90, 120))
    ),
    "tetragonal": Phase(
        point_group="4/mmm",
        structure=Structure(lattice=Lattice(0.5, 0.5, 1, 90, 90, 90))
    ),
    "orthorhombic": Phase(
        point_group="mmm",
        structure=Structure(lattice=Lattice(1, 2, 3, 90, 90, 90))
    ),
    "monoclinic": Phase(
        point_group="2/m",
        structure=Structure(lattice=Lattice(1, 2, 3, 90, 93, 90))
    ),
    "triclinic": Phase(
        point_group="-1",
        structure=Structure(lattice=Lattice(1, 2, 3, 70, 100, 120))
    )
}
phase = phases["triclinic"]

# Set a point group and define a reference orientation (goal)
ori_ref = Orientation.from_axes_angles([1, 1, 1], np.pi / 4, phase.point_group)

# Specify two crystal directions (any will do)
uvw = [[2, 1, 1], [1, 3, 1]]
vec_c = Miller(uvw=uvw, phase=phase)

# Find out where these directions in the reference orientation (crystal)
# point in the sample reference frame
vec_r = Vector3d(~ori_ref * vec_c)

# Plot the reference orientation directions as empty circles
fig = vec_r.scatter(
    ec=["r", "b"],
    s=100,
    fc="none",
    grid=True,
    axes_labels=["X", "Y"],
    return_figure=True,
    hemisphere="both",
)

# Add some randomness to the sample directions (0 error magnitude gives
# exact result)
mag_err = 0
vec_err = Vector3d(np.random.normal(0, mag_err, 3))
vec_r_err = vec_r + vec_err
angle_err = np.rad2deg(vec_r_err.angle_with(vec_r)).mean()
print("Vector deviation: ", angle_err)

# Obtain the rotation which aligns the crystal directions with the
# sample directions (both sample to crystal, r2c, and vice versa)
ori_new_r2c, err = Orientation.from_align_vectors(
    vec_c, vec_r_err, ori_ref.symmetry, return_rmsd=True
)
ori_new_c2r = Orientation.from_align_vectors(vec_r_err, vec_c, ori_ref.symmetry)
print("Error: ", err)

# Assert equality
print(np.allclose(ori_ref.data, ori_new_r2c.data))
print(np.allclose(ori_new_r2c.data, (~ori_new_c2r).data))

# Plot the crystal directions in the new orientation
vec_r2 = Vector3d(~ori_new_r2c * vec_c)
vec_r2.scatter(c=["r", "b"], figure=fig)
```

Output:

```python
Vector deviation:  0.0
Error:  0.0
True
True
```

test_orientation_from_align_vectors

anderscmathisen:
Thank you for the excellent feedback, @hakonanes. Glad to see the test you wrote works:)

Regarding the sorting of the parameters, I have inserted the symmetry argument in method `from_align_vectors()` for `Orientation` and `Misorientation` before the weights argument. The reason for this is that I, and I suspect most people, will use the symmetry argument much more than the weights argument.

orix/quaternion/orientation.py
```diff
@@ -474,6 +529,61 @@ def from_euler(
             ori.symmetry = symmetry
         return ori
 
+    @classmethod
+    def from_align_vectors(
```
hakonanes, line 533: To ensure that this method does what we intend it to do and does not produce unintended results, we should restrict the types of input it takes.

I propose that this method returns an orientation transforming some sample vectors to crystal directions, which is in line with the orientation definition in orix.

Following from this, `b` *must* be crystal directions/plane normals (uvw/hkl) while `a` *must* be sample directions. Thus, `b` must be a `Miller` instance or a `ValueError` is raised. We can then obtain the `symmetry` from `b.point_group`, rendering the `symmetry` parameter redundant, so we should remove it.

hakonanes, line 533: To ensure that this method does what we intend it to do and does not produce unintended results, we should restrict the types of input it takes.

I propose that this method returns an orientation transforming some sample vectors to crystal directions, which is in line with the orientation definition in orix.

Following from this, `b` *must* be crystal directions/plane normals (uvw/hkl) while `a` *must* be sample directions. Thus, `b` must be a `Miller` instance or a `ValueError` is raised. We can then obtain `Orientation.symmetry` from `b.point_group`, rendering the `symmetry` parameter redundant, so we should remove it.

orix/quaternion/orientation.py
```diff
@@ -375,6 +375,61 @@ def get_distance_matrix(
 
         return angles
 
+    @classmethod
+    def from_align_vectors(
```
hakonanes, line 379: To ensure that this method does what we intend it to do and does not produce unintended results, we should restrict the types of input it takes.

I propose that this method returns a misorientation transforming some crystal directions in the first crystal into crystal directions in the other crystal.

Following from this, `b` *must* be crystal directions/plane normals (uvw/hkl) in the first crystal, while `a` *must* be crystal directions/plane normals in the other crystal. Thus, `a` and `b` must be `Miller` instances or a `ValueError` is raised. We can then obtain the `Misorientation.symmetry` instance from `(b.point_group, a.point_group)`, rendering the `symmetry` parameter redundant, so we should remove it.

hakonanes (changes requested):
Thanks again for accomodating my comments.

After having looked into (mis)orientation methods, I think we should narrow down their purpose so that they do what we intend them to do and nothing else. I've made two comments to that effect.

Here's another test I've made to convince myself of the use of `Misorientation.from_align_vectors()`:

```python
from diffpy.structure import Lattice, Structure
import numpy as np
from orix.quaternion import Misorientation, Orientation
from orix.vector import Miller, Vector3d


phases = {
    "cubic": Phase(point_group="m-3m"),
    "hexagonal": Phase(
        point_group="6/mmm",
        structure=Structure(lattice=Lattice(1, 1, 2, 90, 90, 120))
    ),
    "trigonal": Phase(
        point_group="-3m",
        structure=Structure(lattice=Lattice(4.9, 4.9, 5.4, 90, 90, 120))
    ),
    "tetragonal": Phase(
        point_group="4/mmm",
        structure=Structure(lattice=Lattice(0.5, 0.5, 1, 90, 90, 90))
    ),
    "orthorhombic": Phase(
        point_group="mmm",
        structure=Structure(lattice=Lattice(1, 2, 3, 90, 90, 90))
    ),
    "monoclinic": Phase(
        point_group="2/m",
        structure=Structure(lattice=Lattice(1, 2, 3, 90, 93, 90))
    ),
    "triclinic": Phase(
        point_group="-1",
        structure=Structure(lattice=Lattice(1, 2, 3, 70, 100, 120))
    )
}

# Specify two phases, possibly different
phase1 = phases["cubic"]
phase2 = phases["hexagonal"]

# Specify one orientation per crystal (phase)
ori_ref1 = Orientation.from_axes_angles(
    [1, 1, 1], np.pi / 3, symmetry=phase1.point_group
)
ori_ref2 = Orientation.from_axes_angles(
    [1, 3, 2], np.pi / 2, symmetry=phase2.point_group
)

# Get reference orientation (goal). Transformations are composed from
# the right, so: crystal 1 -> sample -> crystal 2
mori_ref = Misorientation(
    ori_ref2 * (~ori_ref1), symmetry=(ori_ref1.symmetry, ori_ref2.symmetry)
)

# Specify two directions in the first crystal
vec_c1 = Miller(uvw=[[1, 1, 0], [0, 0, 1]], phase=phase1)

# Express the same directions with respect to the second crystal
vec_c2 = Miller(xyz=(mori_ref * vec_c1).data, phase=phase2)

# Add some randomness to the second crystal directions (0 error
# magnitude gives exact result)
mag_err = 0
vec_err = Miller(xyz=np.random.normal(0, mag_err, 3), phase=phase2)
vec_c2_err = Miller(xyz=(vec_c2 + vec_err).data, phase=phase2)
angle_err = np.rad2deg(vec_c2_err.angle_with(vec_c2)).mean()
print("Vector deviation: ", angle_err)

# Get the misorientation that aligns the directions in the first crystal
# with those in the second crystal
mori_new, err = Misorientation.from_align_vectors(vec_c2_err, vec_c1, return_rmsd=True)
print("Error: ", err)

# Plot the two directions in the (unrotated) first crystal as open
# circles
fig = vec_c1.scatter(
    ec=["r", "b"],
    s=100,
    fc="none",
    grid=True,
    axes_labels=["e1", "e2"],
    return_figure=True,
    hemisphere="both",
)

# Plot the two directions in the second crystal with respect to the
# first crystal's axes
(~mori_ref * vec_c2_err).scatter(c=["r", "b"], figure=fig)

# Check equality (False if an error is added)
print(np.allclose(mori_new.data, mori_ref.data))
```

Output:

```python
Vector deviation:  0.0
Error:  0.0
True
```

test_misorientation_from_align_vectors

hakonanes:
@anderscmathisen, feel free to question my comments.

From https://github.com/pyxem/orix/discussions/400#discussioncomment-3826033:

> This would be a super helpful function for me as well, as parent/child orientation relationships are often similarly given as vectors to avoid confusion, but MDF and ODF calculations require that same info as misorientations.

@argerlt, does the example I included above (https://github.com/pyxem/orix/pull/401#pullrequestreview-1139428658) look like what you're after?

argerlt:
> @anderscmathisen, feel free to question my comments.
> 
> From [#400 (comment)](https://github.com/pyxem/orix/discussions/400#discussioncomment-3826033):
> 
> > This would be a super helpful function for me as well, as parent/child orientation relationships are often similarly given as vectors to avoid confusion, but MDF and ODF calculations require that same info as misorientations.
> 
> @argerlt, does the example I included above ([#401 (review)](https://github.com/pyxem/orix/pull/401#pullrequestreview-1139428658)) look like what you're after?

I believe so. I was actually coming here to give my suggestion for a from_scipy function, but it looks like the method @anderscmathisen worte passes your tests, so I'm happy!

> @anderscmathisen, feel free to question my comments.
> 
> From [#400 (comment)](https://github.com/pyxem/orix/discussions/400#discussioncomment-3826033):
> 
> > This would be a super helpful function for me as well, as parent/child orientation relationships are often similarly given as vectors to avoid confusion, but MDF and ODF calculations require that same info as misorientations.
> 
> @argerlt, does the example I included above ([#401 (review)](https://github.com/pyxem/orix/pull/401#pullrequestreview-1139428658)) look like what you're after?

Yes it does. I was actually getting ready to submit a PR with my own from_scipy function, but @anderscmathisen 's version is is shorter and still passes all the suggested tests, so I have no feedback.

hakonanes:
> Yes it does.

Thank you for confirming.

orix/quaternion/orientation.py
anderscmathisen: Im not 100% sure that what you write here will be correct, but let me know if I have misunderstood something here. I agree that the method should return an orientation transforming sample vectors to crystal directions. But I think this would mean that `a` should be a Miller instance (the crystal direction) and `b` a general Vector3d instance (the sample direction). The quaternion method returns a transformation, ori,  which can act on `b` so that  `a || ori * b`.

hakonanes: You are correct, sorry for my incorrect statements! They are not in line with my own examples above, even.

anderscmathisen: I have done some more thinking, and I believe that you are correct after all. My bad :).

hakonanes: No no, you are correct, this is also in line with the returned rotation from SciPy, which according to their [Rotation.align_vectors()](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.align_vectors.html) docstring is the

> Best estimate of the rotation that transforms b to a.

anderscmathisen: Great! I have some further problems with checking the types of parameters `a` and `b`. Since `Misorientation` class has the stricter type check (both `a` and `b` are `Miller`), and also is the base class for `Orientation`, the `Orientation.from_align_vectors` is also forced to only accept `Miller` type for `a` and `b`, even though we want it to accept `a` being `Miller` and `b` being a `Vector3d`. Do you have any suggestions here?

hakonanes: This is a good point, and yet another example why both `Orientation` and `Misorientation` should depend on `Rotation` (see https://github.com/pyxem/orix/issues/345 raised by @argerlt).

I believe it can be fixed by both calling `Rotation.from_align_vectors()` instead of `super().from_align_vectors()`?

anderscmathisen: I can commit and push my code if the problem I described doesn't make sense.

anderscmathisen: `Rotation` does not have an implementation of `from_align_vectors()` yet. I could add it, or call `Quaternion.from_align_vector`?

hakonanes: `Rotation` extends `Quaternion`, which has `from_align_vectors()`, hence `Rotation` also has it!

anderscmathisen: True!

hakonanes: > The returned object would then be of type Quaternion, but a hack could be to do something along the lines of

Yes, this must be done, and I wouldn't even call it a hack.

orix/quaternion/orientation.py
```diff
@@ -474,6 +531,65 @@ def from_euler(
...
+        cls,
+        a: Miller,
+        b: Vector3d,
+        weights: Optional[np.ndarray] = None,
+        return_rmsd: bool = False,
+        return_sensitivity: bool = False,
+    ) -> Orientation:
+        """Return an estimated orientation to optimally align two sets of vectors.
```
hakonanes, line 543: Focus this description as well, to something like "Return an estimated orientation to optimally align two sets of vectors in the crystal and sample reference frames"?

orix/quaternion/orientation.py
```diff
@@ -375,6 +375,63 @@ def get_distance_matrix(
...
+
+        Returns
+        -------
+        estimated misorientation
+            Best estimate of the misorientation that transforms b to a.
+        rmsd
+            Returned when ``return_rmsd=True``.
+        sensitivity
+            Returned when ``return_sensitivity=True``.
+        """
```
hakonanes, line 416: I suggest to add one example each to this and the other docstrings, see e.g. https://github.com/pyxem/orix/blob/v0.10.1/orix/quaternion/orientation.py#L527-L564. An example is short, running and self-contained, showing a basic use of the method (see the [contributing guide](https://orix.readthedocs.io/en/stable/contributing.html#run-and-write-tests) for tips). You don't need to import NumPy or Matplotlib, since they are available in all docstring examples as `np` and `plt` already.

orix/quaternion/orientation.py
```diff
@@ -474,6 +531,65 @@ def from_euler(
...
+            a=a,
+            b=b,
+            weights=weights,
+            return_rmsd=return_rmsd,
+            return_sensitivity=return_sensitivity,
+        )
+        out = list(out)
+        out[0] = cls(out[0].data)
+
+        if a.phase:
```
hakonanes, line 588:
```suggestion
        if a.phase.point_group:
```

orix/quaternion/quaternion.py
```diff
@@ -498,3 +499,72 @@ def _outer_dask(
 
         new_chunks = tuple(chunks1[:-1]) + tuple(chunks2[:-1]) + (-1,)
         return out.rechunk(new_chunks)
+
+    @classmethod
+    def from_scipy_rotation(cls, rotation: SciPyRotation) -> Quaternion:
+        """Return quaternion(s) from :class:`scipy.spatial.transform.Rotation`(s).
```
hakonanes, line 505: The final "(s)" apparently makes the Sphinx class directive not work, apparently.

```suggestion
        """Return quaternion(s) from :class:`scipy.spatial.transform.Rotation`.
```

orix/quaternion/quaternion.py
```diff
@@ -498,3 +499,72 @@ def _outer_dask(
...
+
+        Parameters
+        ----------
+        rotation
+            rotation of type :class:`scipy.spatial.transform.Rotation.
```
hakonanes, line 510: No need to repeat the class type when it's already linked to in the type hints (see [here](https://orix--401.org.readthedocs.build/en/401/reference/generated/orix.quaternion.Quaternion.from_scipy_rotation.html)).

```suggestion
            SciPy rotation(s).
```

orix/quaternion/orientation.py
```diff
@@ -375,6 +375,63 @@ def get_distance_matrix(
 
         return angles
 
+    @classmethod
+    def from_align_vectors(
+        cls,
+        a: Miller,
```
hakonanes, line 381: I've reconsidered the naming of these parameters after creating the above examples. While the parameters are in line with SciPy, they, and their interpretation, are not in line with orix' definition of a misorientation. To make their definition as clear as possible, I suggest to rename `a` to `to` and `b` to `from` and rephrase the parameter descriptions to (`a`/`to`) `Directions in the other crystal` and (`b`/`from`) `Directions in the initial crystal`. We can keep the order, as it is in line with how we compose transformations (...* `r3` * `r2` * `r1`). What do you think?

orix/quaternion/orientation.py
```diff
@@ -474,6 +531,65 @@ def from_euler(
             ori.symmetry = symmetry
         return ori
 
+    @classmethod
+    def from_align_vectors(
+        cls,
+        a: Miller,
```
hakonanes, line 537: I suggest the same change of parameter names and docstring descriptions here as in `Misorientation.from_align_vectors()`.

orix/quaternion/quaternion.py
```diff
@@ -498,3 +499,72 @@ def _outer_dask(
...
+            Quaternion(s).
+        """
+        b, c, d, a = rotation.as_quat()
+
+        return cls([a, b, c, d])
+
+    @classmethod
+    def from_align_vectors(
+        cls,
+        a: Vector3d,
```
hakonanes, line 524: I suggest to rename the parameters and rephrase their descriptions as in `Misorientation.from_align_vectors()`.

hakonanes (changes requested):
I've got some comments for parameter naming and descriptions and some minor things for the implementation, otherwise I think it looks good. I suggest to add some docstring examples and tests after my comments are addressed.

orix/quaternion/orientation.py
anderscmathisen: Small problem: `from` is s restricted word in Python and can't be used as argument name. How about `rotate_from` and `rotate_to`?

orix/quaternion/orientation.py
```diff
@@ -375,6 +375,63 @@ def get_distance_matrix(
...
+        estimated misorientation
+            Best estimate of the misorientation that transforms b to a.
+        rmsd
+            Returned when ``return_rmsd=True``.
+        sensitivity
+            Returned when ``return_sensitivity=True``.
+        """
+        if not isinstance(a, Miller) or not isinstance(b, Miller):
+            raise ValueError(
+                f"Arguments a and b must both be of type Miller, but are of type {type(a)} and {type(b)}."
```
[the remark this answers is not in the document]
anderscmathisen: Hmm, I did do the `pre-commit install` according to the contributing guide, and the pre-commit does print something when I make commits. Here is the output of my most recent commit  (somewhat edited because I use a weird font in my terminal). 

```
$ git commit -m "import Miller added to same line as others"
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/anders/.cache/pre-commit/patch1665664778-51385.
black-jupyter............................................................Passed
isort (python)...........................................................Passed
[INFO] Restored changes from /Users/anders/.cache/pre-commit/patch1665664778-51385.
[align_vectors edb1256] import Miller added to same line as others
 1 file changed, 1 insertion(+), 2 deletions(-)
```

I will look some more into why `black` did not do anything about the long codline, but please let me know if you spot any immediate mistakes.

orix/quaternion/orientation.py
anderscmathisen: I temporary went with `to_` and `from_` based on some "[convention](https://stackoverflow.com/questions/37968516/how-do-i-use-a-keyword-as-a-variable-name)" I read  about on stack overflow. These are probably not the best names, and I will try to think of something better.

anderscmathisen:
I have added some examples in the docstrings of the methods. These are probably not the best examples, but probably work fine? I would however have liked to have a better method of displaying the results rather than `np.allclose()` and return value `True`, but the docstring examples test code gave me errors when I used `print`. Any ideas?

The test I added for `Quaternion.from_scipy_rotation` method fails for some reason, even though I think the code should be ok. This is probably related to what @argerlt  wrote in the [discussion](https://github.com/pyxem/orix/discussions/400#discussioncomment-3826033), but Im  not sure. 

The other unit tests are probably a little too focused on checking the type of the return values, idk, these are my first tests. Feel free to give feedback. Perhaps it would have been better to just use the tests you have written in this PR, @hakonanes ?

orix/quaternion/orientation.py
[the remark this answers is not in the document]
hakonanes: That's weird, also weird that it passed our code style check. But it's not important for now, let's just leave it.

orix/quaternion/orientation.py
hakonanes: OK, didn't think of this. Then I suggest to use `initial` and `other`. I've never seen parameter names with underscores in user facing functions in Python before, so I suggest to stay away from those!

orix/quaternion/orientation.py
```diff
@@ -375,6 +375,73 @@ def get_distance_matrix(
...
+        estimated misorientation
+            Best estimate of the ``misorientation`` that transforms ``from_`` to ``to_``.
+            The symmetry of the ``misorientation`` is inferred from the phase of ``to_`` and ``from_``, if given.
+        rmsd
+            Returned when ``return_rmsd=True``.
+        sensitivity
+            Returned when ``return_sensitivity=True``.
+
+        Examples
+        -------
```
hakonanes, line 417: Sphinx is nitpicky, the single missing "-" makes the example render incorrectly ([see docs built from this PR](https://orix--401.org.readthedocs.build/en/401/reference/generated/orix.quaternion.Misorientation.from_align_vectors.html)).

```suggestion
        --------
```

I recommend to check formatting locally before pushing by following our [contributing guide](https://orix.readthedocs.io/en/stable/contributing.html#build-and-write-documentation).

orix/quaternion/orientation.py
```diff
@@ -375,6 +375,73 @@ def get_distance_matrix(
...
+    @classmethod
+    def from_align_vectors(
+        cls,
+        to_: Miller,
+        from_: Miller,
+        weights: Optional[np.ndarray] = None,
+        return_rmsd: bool = False,
+        return_sensitivity: bool = False,
+    ) -> Misorientation:
+        """ "Return an estimated misorientation to optimally align two sets of vectors, one in each crystal."
```
hakonanes, line 387: Remove quotes, and make sure that no docstring lines are longer than 72 characters.

If you haven't already, I recommend adding two visual guides (vertical lines) at 72 and 88 characters in your editor (Under `Settings > Editor > Code Style > Visual guides` in PyCharm).

orix/quaternion/orientation.py
```diff
@@ -375,6 +375,73 @@ def get_distance_matrix(
...
+        rmsd
+            Returned when ``return_rmsd=True``.
+        sensitivity
+            Returned when ``return_sensitivity=True``.
+
+        Examples
+        -------
+        >>> from orix.quaternion import Misorientation
+        >>> from orix.vector import Vector3d, Miller
+        >>> crystal1 = Miller([[1, 0, 0], [0, 1, 0]])
```
hakonanes, line 420: `Miller` requires a `Phase`, so this example fails. You can create a minimal phase with `Phase(point_group="m-3m")`. You can run the docstring examples as explained in our [contributing guide](https://orix.readthedocs.io/en/stable/contributing.html#run-and-write-tests).

Otherwise I'd make the following changes to the example:
* Remove unused `Vector3d`
* Rename `crystal1` to `uvw1`, and pass the vectors as `Miller(uvw=[[1, 0, 0], [0, 1, 0]], phase=Phase(point_group="m-3m")`
* Keep the returned misorientation and show that it rotates the first directions to the second directions by finally calling `mori12 * uvw1`. A much used variable name for a misorientation is `mori`, which would be `mori12` here.
* Should print the misorientation as well to show that the symmetries are retreived from the two sets of crystal directions.

orix/quaternion/orientation.py
```diff
@@ -474,6 +541,74 @@ def from_euler(
...
+        rmsd
+            Returned when ``return_rmsd=True``.
+        sensitivity
+            Returned when ``return_sensitivity=True``.
+
+        Examples
+        -------
+        >>> from orix.quaternion import Orientation
+        >>> from orix.vector import Vector3d, Miller
+        >>> crystal_millers = Miller([[2, -1 ,0], [0, 0, 1]])
```
hakonanes, line 586: Again, pass the phase with a point group, which by printing the orientation demonstrates that the symmetry is obtained from the passed set of crystal directions.

orix/quaternion/quaternion.py
```diff
@@ -498,3 +499,92 @@ def _outer_dask(
...
+        ----------
+        quaternion
+            Quaternion(s).
+
+        Examples
+        -------
+        >>> from orix.quaternion import Quaternion
+        >>> from scipy.spatial.transform import Rotation as SciPyRotation
+        >>> scipy_rot = SciPyRotation.random(random_state = 1)
+        >>> ori = Quaternion.from_scipy_rotation(scipy_rot)
```
hakonanes, line 522: It's not a rotation but a quaternion, call it `quat` instead. Also, pass an explicit rotation to `SciPyRotation` so that we show the different ordering of quaternion elements.

orix/tests/quaternion/test_orientation.py
```diff
@@ -40,7 +40,9 @@
     _groups,
     _proper_groups,
 )
-from orix.vector import AxAngle, Vector3d
+from orix.vector import AxAngle, Vector3d, Miller
```
hakonanes, line 43: Imports are sorted alphabetically (so we find what we're looking for faster), so this should be `AxAngle, Miller, Vector3d`. And, the `Phase` import must go above the import of `AxAnglePlot` etc.

As stated in our [code style](https://orix.readthedocs.io/en/stable/contributing.html#code-style) guide, imports are sorted in three blocks: standard libraries, 3rd party libraries and finally orix imports. Move the diffpy.structure imports above the Matplotlib import.

orix/tests/quaternion/test_orientation.py
```diff
@@ -354,6 +356,36 @@ def test_get_distance_matrix_equal_explicit_calculation(self, symmetry):
...
+            point_group="4", structure=Structure(lattice=TETRAGONAL_LATTICE)
+        )
+        a = Miller([[2, -1, 0], [0, 0, 1]], phase=TETRAGONAL_PHASE)
+        b = Miller([[3, 1, 0], [-1, 3, 0]], phase=TETRAGONAL_PHASE)
+        ori = Misorientation.from_align_vectors(a, b)
+        assert type(ori) == Misorientation
+        assert ori.symmetry == (
+            TETRAGONAL_PHASE.point_group,
+            TETRAGONAL_PHASE.point_group,
+        )
```
hakonanes, line 371: You only use the lattice once, and there is only one phase in this test, so no need to add the crystal system to the name. These variables were upper case in other test files because they were global variables within the files, which is a convention.

```suggestion
        phase = Phase(
            point_group="4", structure=Structure(lattice=Lattice(0.5, 0.5, 1, 90, 90, 90))
        )
        a = Miller([[2, -1, 0], [0, 0, 1]], phase=phase)
        b = Miller([[3, 1, 0], [-1, 3, 0]], phase=phase)
        ori = Misorientation.from_align_vectors(a, b)
        assert type(ori) == Misorientation
        assert ori.symmetry == (phase.point_group,) * 2
```

orix/tests/quaternion/test_orientation.py
```diff
@@ -354,6 +356,36 @@ def test_get_distance_matrix_equal_explicit_calculation(self, symmetry):
...
+        assert type(m) == np.ndarray
+        assert m.shape == (3, 3)
+        out = Misorientation.from_align_vectors(
+            a, b, return_rmsd=True, return_sensitivity=True
+        )
+        assert len(out) == 3
+        a = Vector3d([[2, -1, 0], [0, 0, 1]])
+        with pytest.raises(
+            ValueError,
+            match="Arguments to_ and from_ must both be of type Miller, but are of type <class 'orix.vector.vector3d.Vector3d'> and <class 'orix.vector.miller.Miller'>.",
```
hakonanes, line 385: Put on two lines.

orix/tests/quaternion/test_orientation.py
```diff
@@ -354,6 +356,36 @@ def test_get_distance_matrix_equal_explicit_calculation(self, symmetry):
...
+        b = Miller([[3, 1, 0], [-1, 3, 0]], phase=TETRAGONAL_PHASE)
+        ori = Misorientation.from_align_vectors(a, b)
+        assert type(ori) == Misorientation
+        assert ori.symmetry == (
+            TETRAGONAL_PHASE.point_group,
+            TETRAGONAL_PHASE.point_group,
+        )
+        assert np.allclose(a.unit.data, (ori * b.unit).data)
+        _, e = Misorientation.from_align_vectors(a, b, return_rmsd=True)
+        assert type(e) == np.float64
```
hakonanes, line 374: I recommend to check the value instead.

orix/tests/quaternion/test_orientation.py
```diff
@@ -354,6 +356,36 @@ def test_get_distance_matrix_equal_explicit_calculation(self, symmetry):
...
+        _, m = Misorientation.from_align_vectors(a, b, return_sensitivity=True)
+        assert type(m) == np.ndarray
```
hakonanes, line 376: I recommend to check the value instead. Checking the shape is then done within `np.allclose()`, so the next line is unnecessary.

orix/tests/quaternion/test_orientation.py
```diff
@@ -404,6 +436,34 @@ def test_from_matrix_symmetry(self):
...
+
+        TETRAGONAL_LATTICE = Lattice(0.5, 0.5, 1, 90, 90, 90)
+        TETRAGONAL_PHASE = Phase(
+            point_group="4", structure=Structure(lattice=TETRAGONAL_LATTICE)
+        )
+        a = Miller([[2, -1, 0], [0, 0, 1]], phase=TETRAGONAL_PHASE)
+        b = Vector3d([[3, 1, 0], [-1, 3, 0]])
+        ori = Orientation.from_align_vectors(a, b)
+        assert type(ori) == Orientation
+        assert ori.symmetry.name == "4"
```
hakonanes, line 449: Make dynamic `ori.symmetry.name == phase.point_group.name`.

orix/tests/quaternion/test_quaternion.py
```diff
@@ -282,3 +283,31 @@ def test_outer_dask_wrong_type_raises(self):
...
+        assert m.shape == (3, 3)
+        out = Quaternion.from_align_vectors(
+            a, b, return_rmsd=True, return_sensitivity=True
+        )
+        assert len(out) == 3
+
+    def test_from_scipy_rotation(self):
+        euler = np.array([15, 32, 41]) * np.pi / 180
+        reference_rot = Orientation.from_euler(euler)
+        scipy_rot = SciPyRotation.from_euler("zxz", euler)  # bunge convention
```
hakonanes, line 309: A very good test, just shows how horrible Euler angles are! Two changes are important:
1. Using Euler angles, a SciPy rotation is the inverse of an orix rotation, so we have to invert the orix one
2. SciPy accepts both extrinsic rotations (about a fixed reference frame) and intrinsic rotations (about a reference frame moving with the rotations). Rotations in orix assume the latter, hence we have to capitalize rotation axes.

```suggestion
        reference_rot = Rotation.from_euler(euler, direction="crystal2lab")
        scipy_rot = SciPyRotation.from_euler("ZXZ", euler)  # bunge convention
```

One could argue that we should invert rotations within `Quaternion.from_scipy_rotation()` before returning them because of point 2. above. What do you say, @argerlt and @harripj?

hakonanes: Actually, this test should be moved to `test_rotation.py`, with `Rotation.from_scipy_rotation()` called instead.

hakonanes (changes requested):
Great that you've added docstring examples and tests! I find this PR close to completion.

orix/quaternion/orientation.py
hakonanes: Also, change "one" to "one set", because some people might read it as "one vector in each crystal".

orix/quaternion/orientation.py
```diff
@@ -375,6 +375,73 @@ def get_distance_matrix(
...
+
+        out = super().from_align_vectors(
+            to_=to_,
+            from_=from_,
+            weights=weights,
+            return_rmsd=return_rmsd,
+            return_sensitivity=return_sensitivity,
+        )
+        out = list(out)
+        if hasattr(to_.phase, "point_group") and hasattr(from_.phase, "point_group"):
```
hakonanes, line 440: This will always be true, so it must be

```suggestion
        if to_.phase.point_group and from_.phase.point_group:
```

orix/quaternion/orientation.py
```diff
@@ -474,6 +541,74 @@ def from_euler(
...
+            to_=to_,
+            from_=from_,
+            weights=weights,
+            return_rmsd=return_rmsd,
+            return_sensitivity=return_sensitivity,
+        )
+        out = list(out)
+        out[0] = cls(out[0].data)
+
+        if hasattr(to_.phase, "point_group"):
```
hakonanes, line 607:
```suggestion
        if to_.phase.point_group:
```

hakonanes:
> Perhaps it would have been better to just use the tests you have written in this PR, @hakonanes ?

I think your tests are fine. I'd like to add my examples myself as examples in the doc examples section after this PR is ready, if that's OK with you.

orix/quaternion/orientation.py
anderscmathisen: A miller object can me initialized without a phase argument.

```
m = Miller([[1,0,0]]) #this runs fine
print(m.phase.point_group) #error appears here
```
output
```
AttributeError: 'NoneType' object has no attribute 'point_group'
```

And so we could get errors raised if we use `if to_.phase.point_group:`. But the code above might be a bug with the Miller object?

anderscmathisen:
Yes, that is perfectly fine with me :)

orix/quaternion/orientation.py
hakonanes: You are right. To avoid too many checks, I suggest to just use

```python
try:
    mori = (from_.phase.point_group, to_.phase.point_group)
except (AttributeError, ValueError):
    pass
```

`AttributeError` is raised if the `Miller.phase == None`, and `ValueError` is raised if both `Miller.phase.point_group == None`.

It's not a bug, but this shows that a warning should be raised when passing `Miller([1, 0, 0])`, because then it is assumed that this vector is expressed in terms of the sample reference frame (cartesian coordinates, `xyz`) and not the crystal. This behavior is described in [the documentation](https://orix.readthedocs.io/en/stable/reference/generated/orix.vector.Miller.html). In most cases, if not all, the user should pass the vector as `uvw` or `hkl` etc. Having the possibility to just pass a vector without specifying the vector type or phase is necessary so that methods inherited from `Vector3d` and `Object3d`, where `data` and nothing else is passed on, works. Inherited methods in `Miller` is not handled perfectly yet, and this issue https://github.com/pyxem/orix/issues/367 was raised to improve this.

hakonanes (changes requested):
Tests and examples are good! I have one final question which we must address, and after them I'm happy to make the final changes (add you to the credits list and update the changelog).

hakonanes:
The failing test on Windows with Python 3.9 is handled in #402.

orix/quaternion/quaternion.py
```diff
@@ -498,3 +499,109 @@ def _outer_dask(
...
+        if isinstance(other, Vector3d):
+            vec1 = other.unit.data
+        else:
+            vec1 = np.asarray(other, dtype="float")
+            vec1 /= np.linalg.norm(vec1, axis=-1).reshape(vec1.shape[0], 1)
+        if isinstance(initial, Vector3d):
+            vec2 = initial.unit.data
+        else:
+            vec2 = np.asarray(initial, dtype="float")
+            vec2 /= np.linalg.norm(vec2, axis=-1).reshape(vec1.shape[0], 1)
```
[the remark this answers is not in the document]
anderscmathisen: I agree that the scipy error message is descriptive enough for vectors of different length. 

I also agree that flattening the Vector3d is a way that could work. My only concern is that if the user sends in an array of vectors that is not already flat, then the user might have done so because it expects the function to work in a different way than it actually does(eg initializing arrays of quaternions). It might therefore be better to state in the doctoring that the vectors need to be flat. What do you think?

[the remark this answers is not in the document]
hakonanes: Very good point, I'm happy with requiring "flattened vectors". I would not state that the "vectors have to be of shape `(N, 3)`", as our `Vector3d` would only have shape `(N,)`, since the last dimension is "hidden" in `Vector3d.shape`.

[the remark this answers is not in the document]
anderscmathisen: I have now added this to docstring and pushed. I also used your implementation of accepting `np.ndarray` arguments as it was simpler. I did not implement any error raising if the shape is wrong. I guess the scipy method would do that for us as well:

```ValueError: Expected input `a` to have shape (N, 3), got (2, 2, 3)```

**Review by hakonanes — APPROVED** [review]
I'm happy with this PR now, great work @anderscmathisen!

Before merging, I'd like to wait for @harripj or @argerlt to give their take on my question at the bottom of https://github.com/pyxem/orix/pull/401#discussion_r995462547.

**Comment by hakonanes** [comment]
How would you like your name listed in the credits @anderscmathisen, "Anders Christian Mathisen"? I'll add your name with NTNU as the affiliation in the [Zenodo metadata file](https://github.com/pyxem/orix/blob/develop/.zenodo.json).

**Comment by hakonanes** [comment]
And this is the changelog entries I plan to add to your branch

```restructuredtext
- Creation of ``Quaternion``(s) (or instances of inheriting classes) from SciPy
  ``Rotation``(s).
- Creation of one ``Quaternion`` or ``Rotation`` by aligning sets of vectors in two
  reference frames, one ``Orientation`` by aligning sets of sample vectors and crystal
  vectors, and one ``Misorientation`` by aligning two sets of crystal vectors in two
  differente crystals.
```

harripj:
Thanks for the ping @hakonanes, I will try to have a look a this before tomorrow!

anderscmathisen:
> How would you like your name listed in the credits @anderscmathisen, "Anders Christian Mathisen"? I'll add your name with NTNU as the affiliation in the [Zenodo metadata file](https://github.com/pyxem/orix/blob/develop/.zenodo.json).

Anders Christian Mathisen is fine 👍🏼 The changelog also looks good to me.

orix/tests/quaternion/test_quaternion.py
hakonanes: Do you have time to consider my point 2. above, @harripj? I'll merge this PR after that is settled, with changes made if we go for the inversion.

argerlt: Hi, apologies for radio silence, I've been writing my candidacy document and had time for little else.
Here are my thoughts on both from_scipy and the unit test:
All of scipy is written using active notation and the vector-scalar quaternion used in computer graphics. All of orix is written in passive using scalar vector quaternions from the Rowenhorst paper.  So, I think it makes sense when writing the conversion to assume the input is given in active and needs to be converted to passive.  to make this clear, there should be a value in the function such as:
`Orientation.from_scipy_rotation(cls, input_interpretation='Active')

Secondly though, I the problem with the tests is related to how each program handles the double cover of quaternions across rotation space. Rotations are part of the Lie Group SO3, but quaternions are part of Lie group SU2, which is related homomorphically, and has two quaternions that related to each unique rotation.  Rowenhorst et al. solves this explicitly by giving rules for casting quaternions to the "Northern hemisphere". (ie, q0 is always >0) . However, scipy does not. 
I think this is intentional: using the double cover of quaternions lets people encode inversion operations to their symmetry rotations. For example, (-0.5.0.5,0.5,0.5) and (0.5.0.5,0.5,0.5)  are quaternions that both represent the same 3-fold spin around <111>, but by saying "if q0 is negative, include inversion" expands the symmetry definition in a handy way. However, scipy was designed by a group of people, not a single person, and I think they only address the problems this caused as they come up. thus, problems such as the ones here.

The solution is to not use quaternions, but instead use rotation matrices. rotation matrices are the ONLY representation in either orix or scipy that is actually isomorphic with SO3 (ie, one-to-one and a continuous smooth manifold). Thus, scipy's quat_to_euler formula correctly handles this double cover ambiguity with it's own internal rules. 

This code demonstrates what I am talking about:
```
# passive bunge representation of an Euler angle triplet
passive_bunge = np.array([53, 14, 11])*np.pi/180
# The equivalent active rotation is reversed and the negative (ie, inverted)
active_bunge = passive_bunge[::-1]*(-1)
print("Passive representation:{}".format(passive_bunge))
print("Active representation:{}".format(active_bunge))


# A vector-scalar active quaternion can be made as thus, using capital letters
# to denote extrinsic rotations:
scipy_quat_from_active = R.from_euler('ZXZ',active_bunge)
# or equivalently, created from the passive bunge as thus:
scipy_quat_from_passive = R.from_euler('ZXZ',passive_bunge[::-1]*(-1))
#A scalar-vector passive quaternion can be made as thus:
orix_quat = Orientation.from_euler(passive_bunge)
print("\nscipy quaternion from passive bunge:{}".format(scipy_quat_from_passive.as_quat()))
print("orix quaternion from passive bunge:{}".format(orix_quat.data))


# However, while orix quaternions are uniquely defined,scipy quaternions are not.
# consider this random choice of rotation matrices:
unnorm_rand_rots = np.random.rand(10000,3,3)-0.5
dets = np.linalg.det(unnorm_rand_rots)
unnorm_rand_rots[dets<0] = unnorm_rand_rots[dets<0]*(-1)
passive_rots = unnorm_rand_rots/((np.linalg.det(unnorm_rand_rots).reshape(10000,1,1))**(1/3))
active_rots = np.stack([x.T for x in passive_rots])
# note how when formed into orix quats, the first value is always positive
orix_quats = Orientation.from_matrix(passive_rots)
# but scipy quats obey no such rules
scipy_quats = R.from_matrix(active_rots)
print("\norix minimum value by column:{}".format(np.min(orix_quats.data, axis=0)))
print("scipy minimum value by column:{}".format(np.min(scipy_quats.as_quat(), axis=0)))
# However, both have internally consistent rules for how to convert to
# rotation matrices, the only isomorphic representation of SO3

passive_bunges = (np.random.rand(1000,3) - 0.5)*np.array([np.pi*2,np.pi,np.pi*2])
active_bunges = passive_bunges[...,::-1]*(-1)
scipy_rot = R.from_euler('ZXZ',active_bunges).as_matrix()
orix_rot = Orientation.from_euler(passive_bunges).to_matrix()
print("\nmaximum difference between conventions when converting using")
print("the rotation matrix : {}".format(np.max(scipy_rot-orix_rot)))
```

argerlt:
> I'm happy with this PR now, great work @anderscmathisen!
> 
> Before merging, I'd like to wait for @harripj or @argerlt to give their take on my question at the bottom of [#401 (comment)](https://github.com/pyxem/orix/pull/401#discussion_r995462547).

I left a comment describing my approach to solving the testing problem. If you are interested and this issue is still open October 25th, I can write up a from_scipy function that uses rotations as the pass-between method. Otherwise, @anderscmathisen or @hakonanes , feel free to write your own version, or ignore my comment entirely. 

If you DO want to write it yourself, I believe the method should be something like this:

```
@classmethod
def from_scipy_rotation(cls,assume_active_input=True):
    if assume_active_input == True:
        passive_matrix = np.linalg.inv(scipy_quat.as_matrix())
    else:
        passive_matrix =scipy_quat.as_matrix()
    orix_quat = Orientation.from_matrix(passive_matrix)
```
And the check would be something like:
 ```
# check bunge conversion
passive_bunges = (np.random.rand(1000,3) - 0.5)*np.array([np.pi*2,np.pi,np.pi*2])
active_bunges = passive_bunges[...,::-1]*(-1)
scipy_rot = R.from_euler('ZXZ',active_bunges).as_matrix()
orix_rot = Orientation.from_euler(passive_bunges).to_matrix()
assert(np.max(scipy_rot-orix_rot)<1E-10
 ```
 but obviously much better formatted. Hope this helps!

orix/tests/quaternion/test_quaternion.py
hakonanes: Thank you for giving this some serious thought, @argerlt.

TL;DR: The only change I suggest to the current implementation is to invert rotations returned from `Quaternion.from_scipy_rotation()`.

@pc494 made an argument against exposing the user to considerations of passive/active rotations in https://github.com/pyxem/orix/issues/249#issue-1074368445:

> I think we should make almost all from_euler code (and similar) private so that we can maintain a single internal set of conventions that are consistently applied. As an example, I believe the "lab2crystal" and "crystal2lab" arguments set whether imported data is "active" (static if your transform3d) or "passive" (rotating in transform3d) - this can get extremely confusing very quickly. Basically I don't think we want this exposed to users.

I agree 100%, and I advice against adding a parameter for toggling passive/active rotation interpretation, and aven mentioning it in the docstring. I think the only necessary interpretation is that a transformation is from the lab to the crystal reference frames, and the inverse transformation can be obtained by inverting the rotation. What do you think, @argerlt?

My initial motivation for not inverting rotations internally is because then SciPy's `Rotation.apply(v)` and orix' `Rotation * Vector3d` produce the same vectors. With the implementation in this PR:

```python
>>> import numpy as np
>>> import scipy.spatial.transform.Rotation as SciPyRotation
>>> q1, q2, q3, q0 = [0, 0, np.sin(np.pi/4), np.cos(np.pi/4)]
>>> rot_sp = SciPyRotation([q1, q2, q3, q0])
>>> rot = Rotation.from_scipy_rotation(rot_sp)
>>> v = [1, 2, 3]
>>> rot * Vector3d(v)
Vector3d (1,)
[[-2.  1.  3.]]
>>> rot_sp.apply(v)
[-2.  1.  3.]
>>> (~rot) * Vector3d(v)
Vector3d (1,)
[[ 2. -1.  3.]]
```

But, I guess with SciPy's interpretation of a rotation as active means we should invert internally, as you say @argerlt. The above example can be used in the docstring to clearly show the different interpretations of rotations in the two packages.

argerlt: > I think we should make almost all from_euler code (and similar) private so that we can maintain a single internal set of conventions that are consistently applied. As an example, I believe the "lab2crystal" and "crystal2lab" arguments set whether imported data is "active" (static if your transform3d) or "passive" (rotating in transform3d) - this can get extremely confusing very quickly. Basically I don't think we want this exposed to users.
> 
> I agree 100%, and I advice against adding a parameter for toggling passive/active rotation interpretation, and even mentioning it in the docstring. I think the only necessary interpretation is that a transformation is from the lab to the crystal reference frames, and the inverse transformation can be obtained by inverting the rotation. What do you think, @argerlt?

Okay, I can be on board with that. I see the wisdom, and as long as edge cases like this are dealt with internally in a self-consistent manner, no need to further obfuscate the problem with Active/Passive terminology.

>But, I guess with SciPy's interpretation of a rotation as active means we should invert internally, as you say @argerlt. The above example can be used in the docstring to clearly show the different interpretations of rotations in the two packages.

YES!! Okay, this is actually a way better way to look at this than my long winded explanation. Scipy is active, Orix is passive, invert the rotation in from_scipy_rotations automatically internally, and avoid active/passive language in the docstrings. 

@anderscmathisen  I'm not positive, but I think you MAY still run into issues in your unittests when checking how scipy and Orix interpret random Bunge triplets. If not, ignore this next part because your implementation is faster. If so though, my function was also overkill; just replace quaterion.py lines 531-533 in the current PR:

```
 b, c, d, a = rotation.as_quat()

return cls([a, b, c, d])
```
with
```
return cls.from_matrix(np.linalg.inv(rotation.as_matrix()))
```

anderscmathisen: I too, definitively agree that introducing the passive vs active conventions onto the user is not a good idea. It very quickly becomes very confusing. I am quite confused about the matter myself. 

With regards to your concerns about issues with the current implementaiton of `from_scipy_rotation()`, @argerlt, it seems to me like you are absolutely correct. I ran the following test, which creates a bunch of random euler angles, and basically checks if the `orix` `from_euler` method produces the same result as `Scipy` `from_euler` combined with `from_scipy_rotation`. And it appears that the current implementation is not working. Please let me know if you spot any errors in the test.

```python
from orix.quaternion import Rotation, Quaternion
from scipy.spatial.transform import Rotation as SciPyRotation
import numpy as np

num = 100000
eulers = np.random.random(num*3)
eulers = eulers.reshape((-1,3))
eulers *= np.array([2*np.pi, np.pi, 2*np.pi])

correct_current = 0
correct_matrix = 0
for euler in eulers:

    reference_rot = Rotation.from_euler(euler, direction="crystal2lab")
    scipy_rot = SciPyRotation.from_euler("ZXZ", euler)  
    
    current_implementation_way = Quaternion.from_scipy_rotation(scipy_rot)

    matrix_way = Quaternion(Rotation.from_matrix(scipy_rot.as_matrix()).data)

    if np.allclose(current_implementation_way.data, reference_rot.data):
        correct_current += 1

    if np.allclose(matrix_way.data, reference_rot.data):
        correct_matrix += 1

print(f"With current implementaiton: {correct_current/num * 100}")
print(f"With matrix implementaiton: {correct_matrix/num * 100}")
```
Output
```
With current implementaiton: 24.804000000000002
With matrix implementaiton: 100.0
```

So from this, the matrix way is definitively the correct way. 

 
> @anderscmathisen I'm not positive, but I think you MAY still run into issues in your unittests when checking how scipy and Orix interpret random Bunge triplets. If not, ignore this next part because your implementation is faster. If so though, my function was also overkill; just replace quaterion.py lines 531-533 in the current PR:
> 
> ```
>  b, c, d, a = rotation.as_quat()
> 
> return cls([a, b, c, d])
> ```
> 
> with
> 
> ```
> return cls.from_matrix(np.linalg.inv(rotation.as_matrix()))
> ```

The suggested solution here will however not work, as the `Quaternion` class currently does not have an implementation of `from_matrix`. The `from_matrix` method can not be "borrowed" from the `Rotation` class either. eg:
 ```python
rot = Rotation.from_matrix(np.linalg.inv(rotation.as_matrix()))
return cls(rot.data)
```


 as importing `Rotation` into `quaternion.py` leads to circular import errors.

 One possibility is to move the `from_matrix` class method of `Rotation` to the `Quaternion` class?

Comment (argerlt) [reply orix/tests/quaternion/test_quaternion.py]: > One possibility is to move the from_matrix class method of Rotation to the Quaternion class?

This is  how I would suggest doing it. As part of both #345 and #373 I've suggested we move all the to/from functions to the Quaternion class. I got positive feedback from @hakonanes , but then got busy and never actually followed through with a completed PR, closest I got was [this branch](https://github.com/argerlt/Mart2Aust_Hackathon/tree/to_from_in_quat) which needs to be redone anyway. I bit off more than I could chew in this branch by also trying to eliminate the Neo Eulerian class, which should have just been a separate PR.

I'd say move just from_matrix, then you or I can open an issue to move the other functions  as part of a later PR. Either that, or someone can write a PR to move all of them now, commit it, then rebase this PR from that one. @hakonanes? @harripj @anderscmathisen  thoughts?

Comment (anderscmathisen) [reply orix/tests/quaternion/test_quaternion.py]: Just threw together a new PR for this #406.

Comment (hakonanes) [reply orix/tests/quaternion/test_quaternion.py]: Hm, I do not see the need for orientation matrices. If we invert the SciPy rotation before extracting the quaternion elements in `from_scipy_rotation()`

```python
        b, c, d, a = rotation.inv().as_quat().T
        return cls(np.column_stack([a, b, c, d]))
```

I get this using your example, @anderscmathisen 

```python
import numpy as np
from scipy.spatial.transform import Rotation as SciPyRotation


eulers = Rotation.random((100_000,)).to_euler()

rot_sp = SciPyRotation.from_euler("ZXZ", eulers)
rot_direct = Rotation.from_scipy_rotation(rot_sp)
rot_via_matrix = Rotation.from_matrix(rot_sp.as_matrix())

rot_ref = Rotation.from_euler(eulers)  # lab2crystal
print(rot_ref.angle_with(rot_direct).mean())
# Out: 0.0
print(rot_ref.angle_with(rot_via_matrix).mean())
# Out: 1.5730901506065411
print(rot_ref.angle_with(~rot_via_matrix).mean())
# Out: 0.0
```

We see that by creating orix rotations from SciPy rotations via matrices *without inverting the latter*, as we have now decided that we should, we get a misorientation angle of pi / 2. This angle disappears when inverting the orix reference rotations.

Now that we invert the SciPy rotations, we must also flip the interpretation of the `initial` and `other` reference frames in `Quaternion.from_align_vectors()` by switching the a, b vectors:

```python
        vec1 = initial.unit.data
        vec2 = other.unit.data

        out = SciPyRotation.align_vectors(
            vec1, vec2, weights=weights, return_sensitivity=return_sensitivity
        )
```

With these changes, my previous examples run as before. Of course, some tests fail, which must be updated.

Lastly, moving some of the rotation class methods upstream to the `Quaternion` class is a good idea (I hope to review your new PR soon, @anderscmathisen).

anderscmathisen: From your test here, @hakonanes, I agree that the quaternion conversion works. 

The quaternions made from `from_euler` vs `from_scipy_rotation` via Scipys `from_euler` are, however, not identical in the sense that the elements are not equal, as my test below shows. With the matrix way, they are identical.
But since the angle_with always is zero, I guess the quaternion way is equivalent and that is probably fine?   


```python
import numpy as np
from scipy.spatial.transform import Rotation as SciPyRotation

num = 100000
eulers = Rotation.random((num,)).to_euler()

rot_sp = SciPyRotation.from_euler("ZXZ", eulers)
rot_direct = Rotation.from_scipy_rotation(rot_sp)
rot_via_matrix = Rotation.from_matrix(rot_sp.as_matrix())
rot_ref = Rotation.from_euler(eulers)  # lab2crystal

print(rot_ref.angle_with(rot_direct).mean())
# Out: 0.0
print(rot_ref.angle_with(rot_via_matrix).mean())
# Out: 1.5730901506065411
print(rot_ref.angle_with(~rot_via_matrix).mean())
# Out: 0.0

#Find the persentage of rotations where a, b, c and d are 
#all equal to the reference.

direct = np.isclose(rot_ref.data, rot_direct.data) 
print(np.count_nonzero(np.all(direct, axis=1)) / num * 100)
# Out: 75.047

matrix = np.isclose(rot_ref.data, rot_via_matrix.data)
print(np.count_nonzero(np.all(matrix, axis=1)) / num * 100)
# Out: 0

matrix_inv = np.isclose(rot_ref.data, (~rot_via_matrix).data)
print(np.count_nonzero(np.all(matrix_inv, axis=1)) / num * 100)
# Out: 100
```

harripj:
Sorry for the delay all, and thanks for the interesting discussion! My only comment regarding inverting or not the Scipy rotation is that we document it in `Notes` of the docstring of `from_scipy_rotation()` so that anyone interested can see the interpretation (and then adjust for their use case as required).

As for deciding a default interpretation as discussed, given that SciPy rotations are active, I also think we should invert internally so that a SciPy rotation will be consistent in the orix framework.

orix/tests/quaternion/test_quaternion.py
hakonanes: I don't understand why not using quaternions work, but have given up trying to. I'm happy with using orientation matrices. Let's finish #406 before continuing here.

hakonanes: #406 is in so we can finish this. Please let me know if you need any assistance @anderscmathisen.

anderscmathisen:
I think my last commits here are consistent with out discussions on this PR. The current implementation uses matrix for scipy conversion and also inverse the scipy rotation. I added a short note in the docstring of `from_scipy_rotation`  about the inversion. It's quite short, and  perhaps not very easy to understand for new users?

hakonanes (approved):
After having run through my own examples/tests and sufficiently assured myself of the correctness of the use of `from_align_vectors()`, I'm happy with this PR. Good work, @anderscmathisen! And thanks to others for the discussion of conventions and the correct approach here.

I will make a follow-up PR adding the mentioned examples to the examples gallery. Feel free to do so yourself @anderscmathisen if you have a nice demonstration of the use of `from_align_vectors()` or any other demonstration!

PR #402, opened 2022-10-12
Add row and column indices of points in 2D CrystalMap

hakonanes:
Add `row` and `column` attributes to `CrystalMap` to keep track of map point indices. This was necessary to clearly demonstrate indexing/slicing of crystal maps in a new example (see docs built from this PR).

All other changes are related to silencing of many warnings in the tests, some type hints, and removal of setting of parameters with the default values (unnecessary).

#### Progress of the PR

See example in the docs built from this PR.

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


hakonanes:
@anderscmathisen, I've updated the contributing guide informing that calls like `pip install -e .[doc]` must be replaced by `pip install -e '.[doc]'` with `zsh` shells (Apple's replacement of good ol' `bash` in newer OS'). Thank you for reporting this issue (in person)!

orix/tests/test_crystal_map.py
```diff
@@ -467,6 +468,80 @@ def test_get_twice(self):
...
+        # fmt: off
+        assert np.allclose(
+            xmap2["a"].row,
+            np.array([
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                1,    1, 1, 1,             1,
+                2, 2, 2, 2, 2,             2,
+                3, 3, 3, 3, 3,             3,
+                4, 4, 4, 4, 4, 4, 4, 4, 4, 4,
+            ])
```
harripj, line 498: I like this formatting, very clear 😄

examples/crystal_maps/select_modify_data.py
```diff
@@ -0,0 +1,129 @@
+"""
```
harripj, line 1: Do you think this would better suit being in notebook format, or is there some practicality to keeping it as a file?

harripj (approved):

examples/crystal_maps/select_modify_data.py
harripj: edit: I have just seen in the docs that the notebook is also available, very nice! Is this autogenerated?

orix/tests/test_crystal_map.py
hakonanes: Indeed, otherwise the test might be a bit difficult to understand...

examples/crystal_maps/select_modify_data.py
hakonanes: The examples gallery is generated by the excellent Sphinx-Gallery, which as you noted also generates notebooks after running the scripts (they can also generate MyBinder links, but I haven't tested it yet).

The example is on the longer side, but I don't think it's too long to call it a "short recipe demonstrating a common task", which is the purpose of the examples. I think users might get more out these than searching through a long tutorial for what they need (like the crystal map tutorial).

hakonanes:
Thanks a lot for reviewing, @harripj!

PR #414, opened 2022-12-07
Examples demonstrating (Mis)Orientation.from_align_vectors()

hakonanes:
Added two examples to the examples gallery (from PR comments https://github.com/pyxem/orix/pull/401#pullrequestreview-1138061052 and https://github.com/pyxem/orix/pull/401#pullrequestreview-1139428658):
* Get an orientation from two sets of vectors, one in the sample reference frame and the other in the crystal reference frame. Uses `Orientation.from_align_vectors()`.
* Get a misorientation from two sets of vectors, one in each of the two crystal reference frames. Uses `Misorientation.from_align_vectors()`.

I've also added the possibility to pass `symmetry` to `Orientation.from_scipy_rotation()` and `Misorientation.from_scipy_rotation()`. `Rotation.from_scipy_rotation()` is also updated, so that all these three methods have custom docstrings. I've also made sure these methods return a tuple instead of a list if either of `return_rmsd` or `return_sensitivity` are `True`.

#### Progress of the PR

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -20,11 +20,12 @@
 
 from __future__ import annotations
 
-from typing import Any, Tuple, Union
+from typing import Any, Optional, Tuple, Union
 
 import dask.array as da
 from dask.diagnostics import ProgressBar
 import numpy as np
+from scipy.spatial.transform import Rotation as SciPyRotation
 from scipy.special import hyp0f1
 
 from orix.quaternion import Quaternion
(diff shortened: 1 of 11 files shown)


orix/quaternion/rotation.py
```diff
@@ -20,11 +20,12 @@
 
 from __future__ import annotations
```
pc494, line 21: Does this have a confirmed "no longer future" version we should be aware of?

pc494 (approved):
This looks good to me too, just one short question.

orix/quaternion/rotation.py
hakonanes: Good question, it seems like it is necessary until Python 3.11 is the minimal version when we can use `from typing import Self` and use that instead of the class itself ([SO](https://stackoverflow.com/a/33533514/3228100) and [Python docs](https://docs.python.org/3/library/typing.html#typing.Self)).

pc494: okay, 3.11 being our min version is a long way away, so this is resolved.

PR #437, opened 2023-03-14
Option to use euler fundamental region for mean orientation

anderscmathisen:
#### Description of the change

Added override to the `mean()` function in the `Orientation` class with the option to calculate the mean in the euler fundamental region. Also keeps the symmetry of the `Orientation` for the returned mean. This way, when plotting in an IPF, I think the mean `Orientation` makes more sense. 

#### Progress of the PR
- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [x] Unit tests with pytest for all lines
- [x] Clean code style by [running black via pre-commit](https://orix.readthedocs.io/en/latest/contributing.html#code-style)

#### Minimal example of the bug fix or new feature
```python
import orix.plot
from orix.quaternion import Orientation
from orix.quaternion.symmetry import D6h
from orix.vector import Vector3d
import matplotlib.pyplot as plt
from matplotlib.gridspec import GridSpec

oris = Orientation(
[[ 0.215 ,  0.9696, -0.1001, -0.0602],
 [ 0.2132 , 0.97  , -0.0977, -0.0643],
 [ 0.2154 , 0.9693 ,-0.098 , -0.067 ],
 [ 0.2157 , 0.9692, -0.0989, -0.0659],
 [ 0.2141 , 0.9693 ,-0.1012, -0.0668],
 [ 0.2132 , 0.9696, -0.1001, -0.0659],
 [ 0.2083 , 0.9715 ,-0.0952, -0.0605],
 [ 0.1558 , 0.7913 ,-0.569 , -0.1608],
 [ 0.1495 , 0.79   ,-0.5704, -0.1681],
 [ 0.1537 , 0.7908, -0.5686, -0.1663],
 [ 0.1552 , 0.7918 ,-0.5677, -0.1634],
 [ 0.1518 , 0.792 , -0.5679, -0.165 ],
 [ 0.0484 , 0.3993, -0.8886, -0.2203],
 [ 0.0459 , 0.4007, -0.888,  -0.2209]], symmetry = D6h)

fig = plt.figure(figsize=(9,5))

gs = GridSpec(2,3)

ax_x_not_fundamental = fig.add_subplot(gs[0,0], projection = "ipf", direction = Vector3d.xvector(), symmetry = D6h)
ax_y_not_fundamental = fig.add_subplot(gs[0,1], projection = "ipf", direction = Vector3d.yvector(), symmetry = D6h)
[...]

Issue #434, opened 2023-03-10
Mean of orientations, with symmetry


DorianDepriester:
Hello there!
It seems that the mean() method for orientations actually doesn't take the orientations' symmetry into account. For instance, let's consider the m-3m symmetry and a set of 3 close orientations:

```python
from orix.quaternion import symmetry, Orientation

s = symmetry.Oh # Cubic symmetry
o = Orientation.from_euler([[0,0,0], [1,89,0], [1,1,89]],symmetry=s,degrees=True)
print(o.in_euler_fundamental_region()*180/np.pi) # Just to be sure what we're talking about

# Compute and print the mean orientation
obar = o.mean()
print(obar.in_euler_fundamental_region()*180/np.pi)
```
Thus, the output is:

> [[ 0.  0.  0.]
>  [ 1. 89.  0.]
>  [ 1.  1. 89.]]
> [[15.79096054 30.29090638 15.77845771]]

The result is quite strange. Indeed, the same computation on MTEX
```matlab
CS=crystalSymmetry('m-3m')
mean(orientation.byEuler([0,1,1]*degree, [0,89,1]*degree, [0,0,89]*degree,CS))
```
gives
```
ans = orientation (m-3m -> xyz)
 
  Bunge Euler angles in degree
     phi1        Phi       phi2       Inv.
  90.8333 0.00290889      269.5          0
```

Still, the same with `CS=crystalSymmetry('1')` leads to:

>   Bunge Euler angles in degree
>     phi1     Phi    phi2    Inv.
>   15.791 30.2909 15.7785       0

Am I doing something wrong? I have looked into MTEX's code, and I understand that the mean() method actually relies on the [project2FundamentalRegion](https://github.com/mtex-toolbox/mtex/blob/b9773cb93a12df07ef5aa1db370b985e7bf74120/geometry/%40orientation/mean.m#L52) function. I have tried to implement it in orix, but without any success.

Thank you in advance.

hakonanes:
Hi Dorian,

you have indeed encountered a somewhat surprising result, from a user standpoint. `Orientation.mean()` does not consider symmetry, as you say. I won't quite call it a bug since the method description states that it computes the quaternion mean. But, it is something we should support, and can do quite easily.

Do you get the desired results with this approach?

```python
>>> from orix.quaternion import Orientation, symmetry
>>> s = symmetry.Oh
[...]

Issue #368, opened 2022-06-17
Problem in map_into_symmetry_reduced_zone


maclariz:
Running map_into_symmetry_reduced_zone() on sensible data (single crystal, all similar orientations) yields the following (just a few lines shown for illustration):

```python
[-0.9681  0.0373 -0.0222  0.2467]
[-0.9685  0.0373 -0.0222  0.2452]
[-0.9685  0.0373  0.0222 -0.2451]
[-0.9686  0.0368  0.016  -0.2455]
[-0.9688  0.0373  0.0223 -0.2442]
```
 
 Fixing this manually in a quick for loop gets me to:
 ```python
 [-0.96812383,  0.03732545,  0.02216656, -0.24668136],
 [-0.96849277,  0.03729216,  0.02222252, -0.2452289 ],
 [-0.96851503,  0.03729006,  0.02222605, -0.24514094],
 [-0.96856316,  0.03676689,  0.01601705, -0.24551387],
 [-0.96875969,  0.03726781,  0.02226333, -0.24417231],
````
(which is just a numpy array)
       
reimporting as an Orientation object, it then just transforms back to the first.  Obviously, the constant sign flips in this data on index 2 and 3 in each quaternion play merry hell with any calculations!

harripj:
@maclariz could you provide the original orientation data before calling map_into_reduced_symmetry_zone()?

maclariz:
@harripj Here you go:

[ 0.5101 -0.0107  0.0421 -0.859 ]
 [ 0.5114 -0.0107  0.0421 -0.8582]
 [ 0.2451 -0.0222  0.0373 -0.9685]
 [ 0.2455 -0.016   0.0368 -0.9686]
 [ 0.2442 -0.0223  0.0373 -0.9688]

maclariz:
@harripj m3m (FCC)

maclariz:
The following sorted out my data and resulted in a consistent set of orientations that give consistent misorientation angles:

# Take the data into a pure numpy array
ori5 = ori3.data
# Drop into a setting where index 0 (cos angle) is positive, 1 (prop to x) and 2 (prop to y) are negative
# and index 3 (prop to z) are positive
ori5[:,0] = abs(ori5[:,0])
ori5[:,1] = -abs(ori5[:,1])
ori5[:,2] = -abs(ori5[:,2])
ori5[:,3] = abs(ori5[:,3])
# Transform back into orientation object without calling symmetry
ori5 = Orientation(ori5)
[...]

PR #442, opened 2023-04-06
Change left/right symmetry in reduction of (mis)orientations to fundamental zone

hakonanes:
This PR changes the order of symmetry application in the reduction of misorientations to the misorientation fundamental zone from `sym_A * m_AB * sym_B` to `sym_B * m_AB * sym_A`. The new order is in line with the notion of the misorientation from `o_A` to `o_B` being given by `m_AB = o_B * o_A^-1`.

Note that the same algorithm is used to reduce orientations to the Rodrigues fundamental zone, in which case the two symmetries are the symmetry of the sample reference frame, `sym_A = C1` and the symmetry of the crystal, say `sym_B = Oh`.

This change was motivated by several inconsistent results reported lately. This change fixes all these issues:
* https://github.com/pyxem/orix/issues/368#issuecomment-1163212482
* https://github.com/pyxem/orix/issues/434#issuecomment-1492180192
* #437

I had to update all three clustering tutorials. They give comparable results, although I'm unsure of how to solve the mirroring of the clustered Ti orientations in that tutorial. Any help on spotting an error would be appreciated.

The other major change is naming: `map_into_symmetry_reduced_zone()` -> `reduce()`, inspired by the attractive naming of [`scipy.spatial.transforms.Rotation.reduce()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.reduce.html). I've deprecated the former method in favor of the new one, with removal in v0.13. Arguably not as easy to understand as the former name, but the notion of *reducing* a (mis)orientation to its fundamental zone is there. Using this word allows us also to add `reduce` as an intuitive parameter to various methods if we don't want to apply symmetry operations, instead of having a `use_symmetry` parameter (similar to MTEX' `noSymmetry`). We could then rename `Miller.in_fundamental_region()` to `Miller.reduce()` etc.

Testing this change required some other tools and changes, which are also included in this PR:
* New class method `Vector3d.get_path()` to get vectors along the path between two or more vectors. Useful to plot a bounded region on the stereographic projection, or delineate a path between two vectors.
* Doc example showing how to combine rotations, i.e. from left to right. The example is the one in section 4.2.2 in Rowenhorst et al. (2015).
* Changed time of removal of `convention` parameter in to/from_euler() methods from v1.0 to v0.13, because I don't see v1.0 on the horizon.
* Make stereographic polar grid use Matplotlib's `grid.alpha` config value.
* Restructure some tests for clarity.

#### Progress of the PR
- [x] Scrutiny of the change by others, like testing on their own problems. I've reached out to @maclariz, @DorianDepriester and @anderscmathisen in the above mentioned issues for their feedback.

I suggest to inspect the clustering tutorial notebooks in the docs built from this PR:
* [Clustering tutorials](https://orix--442.org.readthedocs.build/en/442/tutorials/index.html#clustering-of-mis-orientations)
* [Example on combining rotations](https://orix--442.org.readthedocs.build/en/442/examples/rotations/combine_rotations.html)

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


hakonanes:
Considering the change in this PR, the notion of the first and second symmetry in `OrientationRegion.from_symmetry(s1, s2=C1)` changes meaning, I think. I suggest to change parameter names to `start, end` (Python does not allow `from`), and remove the default `C1`. `C1` should really be the default for `start`, but then `end` would also need a default, which isn't possible to define.

An alternative would be to allow one or two symmetries to be passed as `*args` and set `from` to `C1` if only one symmetry was passed. This would effectively maintain the syntax of `OrientationRegion.from_symmetry(D6)` which is currently used some places in the code.

Another alternative is two flip the order `to, from=C1`.

hakonanes:
Talked to @anderscmathisen offline, reduction to the FZ now seems to make sense for his use.

maclariz:
Thanks @hakonanes .  I hope I get some chance to test this soon on the dataset where I had the issues in the past.  What is the exact version number of the new orix version which includes this change?  I will probably install in a test environment initially and not destroy my old install just yet.

hakonanes:
That would be great. This change only lives in my fork so far, which you should be able to install into an environment like so:

```bash
pip install git+https://github.com/hakonanes/orix.git@fix-misorientation-reduce
```

If reviewers agree that his change is necessary, I anticipate that the change will be part of the next minor release v0.12, which should be released shortly after this PR is merged.

maclariz:
@hakonanes I have installed in my testing environment, run it on the old data, and found that the symmetry reduction works more reliably on the Silicon misorientations dataset that was giving problems before.

hakonanes:
Happy to hear that, thank you for doing the tests. Please report back if you still find inconsistencies.

maclariz:
Sorry to spoil the party, but this new order does not work correctly in reduce for 622 symmetry.

Attached is the result of doing cluster analysis using symmetry reduction for HCP Ti using the old method.

Screenshot 2023-04-21 at 11 53 08

maclariz:
And now here is the same workbook run with the new method.  The results from the old method are quantitatively sensible.  The new method is screwy.  So, something about your order of rotations in symmetry reduction in hexagonal is wrong.

Screenshot 2023-04-21 at 11 54 48

hakonanes:
Thank you for sharing this test, @maclariz. I believe this is related to (inconsistent?) alignment of crystal axes with symmetry operations. I've encountered issues with point group -3m myself (chromium nitride in duplex steel) using the changes in this PR.

Given this issue, I suggest to wait with the change in the left/right symmetry of `map_into_symmetry_reduced_zone()` until we are sure of all alignments of crystal axes with symmetry operations. To easily debug these things, I suggest we implement automatic plotting of [symmetry elements in the stereographic projection](https://orix.readthedocs.io/en/stable/examples/stereographic_projection/plot_symmetry_operations.html) for all defined point groups. We can then convince ourselves of the alignments, before continuing with this change.

I'll make another PR for the changes in this PR not strictly related to the left/right symmetry change.

argerlt:
Doing a file-by-file, line-by-line parse though this PR, I believe every feature in here has been added to ORIX as of 0.15.rc1.

I am going to close this PR for the sake of cleanup, but feel free to reopen it if anyone disagrees.

PR #450, opened 2023-05-05
Adding Jit-accelerated to/from functions to quaternion class

argerlt:
This is NOT a finished Pull request, just starting a request so we can have a discussion.

### Short version 
I reorganized `Quaternion` class and made it's "to/from" functionality consistently call from "_conversions.py"

### Long version
I started working on Issue #374 (Purpose for Neo-Eulerian Classes), but it's been 9 months since that issue was started, and since then Neo-Eulerian vectors have been used a lot more on the plotting side of things. As such, removing them is no longer a "quick and easy" code simplification, even if it did make sense.

However, the use of `NeoEuler` in `Quaternion` and all child classes of `Quaternion` was purely for to/from functionality, so I removed those and added jit-accelerated versions of those functions to `_conversions.py`

I also reorganized the functions of `Quaternion` into groups to make them easier to navigate. 

I also removed the `Quaternion.from_neo_euler` command with the thought of later removing the `NeoEuler` class altogether, but I would like someone more familiar with the use of `Rodriguez` and `AxAngle` to weight in before I go down that path. For now, I just replaced all the `Rotation.from_neo_euler(Axangle.from_axis_angle(axis,angle)' calls with the much simpler `Rotation.from_axis_angle(axis,angle)`

Next step would be to either continue removing `NeoEuler`, or to re-add `Quaternion.from_neo_euler` and it's test functions, and leave the rest alone for now.

#### Progress of the PR
- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)

```python
>>> from orix import vector
>>> v = vector.Vector3d([1, 1, 1])
>>> # Your new feature...
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


orix/quaternion/quaternion.py
```diff
@@ -407,25 +431,10 @@ def from_matrix(cls, matrix: Union[np.ndarray, tuple, list]) -> Quaternion:
...
-        q[..., 2] = 0.5 * np.sqrt(np.where(q2_almost < _FLOAT_EPS, 0, q2_almost))
-        q[..., 3] = 0.5 * np.sqrt(np.where(q3_almost < _FLOAT_EPS, 0, q3_almost))
-
-        # Modify component signs if necessary
-        q[..., 1] = np.where(om[..., 2, 1] < om[..., 1, 2], -q[..., 1], q[..., 1])
-        q[..., 2] = np.where(om[..., 0, 2] < om[..., 2, 0], -q[..., 2], q[..., 2])
-        q[..., 3] = np.where(om[..., 1, 0] < om[..., 0, 1], -q[..., 3], q[..., 3])
-
+        # Assert input can be interpreted as an array of (3, 3) arrays
+        assert om.ndim >= 2
```
hakonanes, line 435: `assert` is in orix only used in tests. In code we check and raise an appropriate error (a `ValueError` if the input does not have the required form as per the docstring) with an informative error message. Could you change these checks to this format? Again, you can require 2D by using `np.atleast_2d(om)`. Then we only need the second check, I think.

orix/quaternion/quaternion.py
```diff
@@ -616,6 +687,210 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+
+    def to_homochoric(self) -> Vector3d:
+        """Return the neo-Eulerian Homochoric Vector representation of
+        the normalized quaternions.
+        :cite:`rowenhorst2015consistent`.
+
+        Returns
+        -------
+        homo
+            an orix.vector.Vector3D object containing the axes of rotation,
```
hakonanes, line 870:
```suggestion
            A :class:`Vector3d` containing the axes of rotation,
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +687,210 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+            with lengths equal to [0.75*(angle-sin(angle))]^(1/3).
+
+        Examples
+        --------
+        #TODO
+        Notes
+        --------
```
hakonanes, line 877: Sphinx requires exact number of underlines and letters

```suggestion
        -----
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +687,210 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+        This is often used as a plotting tool, as it produces an isomorphic
+        (though not angle-preserving) mapping from the non-euclidean
+        orientation space into cartesian coordinates. Additionaly, unlike
+        Rodriguez vectors, all rotations map into a finite space, bounded by
```
hakonanes, line 881:
```suggestion
        Rodrigues vectors, all rotations map into a finite space, bounded by
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +687,210 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+        fundamental zones in this representation.
+        """
+        ax = self.axis.unit
+        rod = ax * np.tan(self.angle / 2)
+        return rod
+
+    def to_homochoric(self) -> Vector3d:
+        """Return the neo-Eulerian Homochoric Vector representation of
+        the normalized quaternions.
+        :cite:`rowenhorst2015consistent`.
```
hakonanes, line 865:
```suggestion
        """Return the neo-Eulerian homochoric vector representation of
        the normalized quaternions :cite:`rowenhorst2015consistent`.
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +687,210 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+
+        Returns
+        -------
+        homo
```
hakonanes, line 869:
```suggestion
        vec
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +687,210 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+            an orix.vector.Vector3D object containing the axes of rotation,
+            with lengths equal to [0.75*(angle-sin(angle))]^(1/3).
```
hakonanes, line 871: Good to have the maths in the docstring. I suggest to move it to the notes, though. And formatted with LaTeX, like

```
:math:`0.75\cdot(\theta - \sin(\theta))^{1/3}`
```

To render correctly, the docstring string needs to be a raw string, so we must add a `r"""Return the...`

orix/quaternion/quaternion.py
hakonanes: The same formatting suggestions for this method applies to `to_rodrigues()` as well.

orix/quaternion/quaternion.py
```diff
@@ -135,6 +139,22 @@ def d(self, value: np.ndarray):
         """Set the third vector quaternion component."""
         self.data[..., 3] = value
 
+    @property
```
hakonanes, line 142: Not sure about adding axis and angle of rotation to the quaternion class... I hope @pc494 can give his two cents on this change.

orix/tests/quaternion/test_conversions.py
```diff
@@ -123,6 +126,51 @@ def rodrigues_vectors():
...
 
 
+@pytest.fixture
+def orthogonal_matrices():
+    return np.array(
+        [
+            [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]],
+            [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]],
+            [
+                [-0.95541973, -0.29525098, 0.0],
```
hakonanes, line 136: Do you need so many decimals? You only check four decimals in the tests, anyway. If possible, I suggest to reduce the number of decimals to those needed for the current checks.

orix/tests/quaternion/test_orientation.py
```diff
@@ -496,32 +496,14 @@ def test_from_align_vectors(self):
...
-        axangle = AxAngle.from_axes_angles(axis, angle)
-        o1 = Orientation.from_neo_euler(axangle, Oh)
-        o2 = Orientation.from_axes_angles(axis, angle, Oh)
+        o1 = Orientation.from_axes_angles(axis, angle, Oh)
         assert np.allclose(o1.to_euler(degrees=True), [135, 90, 225])
+        assert o1.symmetry.name == "m-3m"
+        o2 = Orientation.from_axes_angles(axis, np.rad2deg(angle), Oh, degrees=True)
         assert np.allclose(o1.data, o2.data)
-        assert o1.symmetry.name == o2.symmetry.name == "m-3m"
-        assert np.allclose(o1.symmetry.data, o2.symmetry.data)
```
hakonanes, line 521: Please keep these lines, as they confirm that initialization of the orientations are correct.

hakonanes (changes requested):
Thank you for simplifying the conversions between quaternions and Neo-eulerian vectors (axis/angle pairs, Rodrigues vectors and homochoric vectors)! This will make people's workflows shorter and debugging faster.

I've made several comments, to both changes and styling. I anticipate another review after this, as there are some points for discussion.

Please let me know where I can assist.

orix/tests/quaternion/test_conversions.py
argerlt: On a related note, do you know what the source was for the rest of the pytest.fixture data in this file?
I thought I remembered finding a set of files from Rowenhorst/Rollett/Degraef at one point containing these values, but cannot seem to find it back now. If such a source did exist, it would be nice to add a related link and description.

orix/quaternion/quaternion.py
argerlt: moving `.axis` and `.angle` to `Quaternion` made it easier to also move `.to_rodrigues` and `.to_axis_angle`, as well as eventually write their respective `from` statements. 

That said, I'm biased. This is really more about making it easier to change up the quaternion class structure, if we ever chose to do so. Down the road options such as :
   - Make symmetry a child of Quaternion, not Rotation
   - Creating a symmetry-aware Rotation class but restricting it to (C1,C1)
   - Misorientation and Orientation inheriting directly from either `Quaternion`, `Rotation`, or `_TAtion`
   -  Allowing things like `Symmetry.from_Rodrigues()`
  
Are all easier and cleaner if all the to/from functions as well as `.axis` and `.angle` are in `Quaternion`. 

Again, I am not saying we should commit to one of these now, I just want to reduce the mental bandwidth required to make any of these changes down the road.

orix/quaternion/quaternion.py
```diff
@@ -69,6 +70,9 @@ class Quaternion(Object3d):
        v'_z = z(a^2 - b^2 - c^2 + d^2) + 2(y(a \cdot b + c \cdot d) + x(b \cdot d - a \cdot c))
     """
 
+    ########################################
```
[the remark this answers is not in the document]
argerlt: Do you have a good  alternative? I agree they don't match the rest of Orix, but a separator between groups of functions would still be nice for code navigation reasons. 

Could also just remove entirely though, I'm not committed to this, just thought it made it easier to find my place in the code.

orix/quaternion/quaternion.py
```diff
@@ -175,86 +195,85 @@ def __mul__(self, other: Union[Quaternion, Vector3d]):
...
-        q = cls(np.vstack((a, b, c, d)).T)
-        return q
+        These are also referred to as Rodrigues-Frank vectors, and
+        are a neo-eulerian vector representation of orientation space:
 
-    @classmethod
-    def from_neo_euler(cls, neo_euler: "NeoEuler") -> Quaternion:
-        """Create unit quaternion(s) from a neo-euler (vector)
-        representation.
+        Note, while this functionality has been added for completeness,
```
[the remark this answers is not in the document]
argerlt: > less opinionated and more informative

You are not the first person to tell me to be less opinionated in my writing.



How about this? I want to make it clear Rodrigues vectors are useful, but this might still be to wordy/ opinionated>

        Rodrigues vectors are a useful rotation parameterization for certain
        applications ,especially visualization. However, their lengths scale
        with :math:\tan(\theta/2), as does their relative error.
        Additionally, rotations of 180 degrees are equivalent to infinitely
        long vectors. For these reasons, it might be better to use
        axis/angle paris instead, which have a defined values and
        consistent relative errors over all rotations.

orix/quaternion/quaternion.py
```diff
@@ -293,9 +312,31 @@ def from_axes_angles(
...
-        q = cls.from_neo_euler(axangle).unit
-        return q
+        # convert all possible tuple, numpy, and/list combinations into
+        # the numpy array of shape (..,4) expected by
+        # quaternions._conversions.ax2qu()
+        axes = Vector3d(axes).unit
+        if degrees:
+            angles = np.deg2rad(angles)
+        angles = np.array(angles)
+        if angles.shape == ():
```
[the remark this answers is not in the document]
argerlt: This is a way better approach, thanks for suggesting it. 
I previously considered putting these checks as-is in `ax2qu`, but didn't want to make the functions in `_conversions.py` accept non-numpy inputs. Your suggestion is the better middle ground though.

orix/quaternion/quaternion.py
hakonanes: I agree with your choice here, having them as attributes of the quaternion class is fine.

orix/quaternion/quaternion.py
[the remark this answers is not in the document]
hakonanes: > You are not the first person to tell me to be less opinionated in my writing.

Hah. We usually write code for our own use case, but when making it available in a library like orix, there might be users who need to use Rodrigues vectors and not axis/angle pairs for some reason we haven't thought of. Best to not alienate those people, I think.

I think your wording is good, just move the first comma and write "axis/angle pairs instead, which have defined values".

orix/quaternion/quaternion.py
```diff
@@ -175,86 +195,85 @@ def __mul__(self, other: Union[Quaternion, Vector3d]):
...
-            First quaternions.
-        q2
-            Second quaternions.
-        q3
-            Third quaternions.
+    @classmethod
+    def from_rodrigues(
+        cls,
+        axes: Union[np.ndarray, Vector3d, tuple, list],
+        ignore_warnings: bool = False,
```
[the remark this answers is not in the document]
hakonanes: Remember to remove the parameter and update the docstring.

Also, the code will not pass the warnings as they are "raised", not emitted. You should use `warnings.warn("message")` for that. I suggest slight rephrasing of the warnings:

1. Highest angle is greater than 179.999 degrees. Rodrigues vectors cannot paramtrize 2-fold rotations. Consider an alternative class method.
2. Max. estimated error is greater than 0.1%. Rodrigues vectors have increasing associated errors for small angle rotations. Consider an alternative class method.

These are good checks, btw, thanks for adding them!

orix/quaternion/quaternion.py
```diff
@@ -616,6 +674,209 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+        """Return an axis-angle representation of the normalized
+        quaternions.
+        :cite:`rowenhorst2015consistent`.
+
+        Returns
+        -------
+        axis
+            an orix.vector.Vector3D object containing the axes of rotation
+        angle
+            a numpy array representing the angles of rotation
```
hakonanes, line 814: Users can deduce the types from the function signature.

```suggestion
        """Return an axis-angle representation of the normalized
        quaternions :cite:`rowenhorst2015consistent`.

        Returns
        -------
        axis
            The axes of rotation.
        angle
            The angles of rotation.
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +674,209 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+    def to_rodrigues(self) -> Vector3d:
+        r"""Return the neo-Eulerian Rodrigues Vector representation of the
+        normalized quaternions.
+        :cite:`rowenhorst2015consistent`.
+
+        Returns
+        -------
+        rod
+            an orix.vector.Vector3D object containing the axes of rotation,
+            with lengths equal to :math:`\tan(angle/2)` .
```
hakonanes, line 831:
```suggestion
        r"""Return the neo-Eulerian Rodrigues vector representation of the
        normalized quaternions :cite:`rowenhorst2015consistent`.

        Returns
        -------
        rod
            The axes of rotation, with lengths equal to
            :math:`\tan(angle/2)`.
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +674,209 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+    def to_homochoric(self) -> Vector3d:
+        r"""Return the neo-Eulerian homochoric Vector representation of
+        the normalized quaternions. :cite:`rowenhorst2015consistent`.
+
+        Returns
+        -------
+        vec
+            A :class:.Vector3D object containing the axes of rotation,
+            with lengths equal to
+            :math:`0.75\cdot(\theta - \sin(\theta))^{1/3}` .
```
hakonanes, line 858:
```suggestion
        r"""Return the neo-Eulerian homochoric vector representation of
        the normalized quaternions :cite:`rowenhorst2015consistent`.

        Returns
        -------
        vec
            The axes of rotation, with lengths equal to
            :math:`0.75\cdot(\theta - \sin(\theta))^{1/3}`.
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +674,209 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+        --------
+        #TODO
+
+        Notes
+        -----
+        This is often used as a plotting tool, as it produces an isomorphic
+        (though not angle-preserving) mapping from the non-euclidean
+        orientation space into cartesian coordinates. Additionaly, unlike
+        Rodrigues vectors, all rotations map into a finite space, bounded by
+        a sphere of radius pi.
```
hakonanes, line 870:
```suggestion
        This is often used as a plotting tool, as it produces an
        isomorphic (though not angle-preserving) mapping from the
        non-euclidean orientation space into cartesian coordinates.
        Additionally, unlike Rodrigues vectors, all rotations map into a
        finite space, bounded by a sphere of radius :math:`\pi`.
```

orix/quaternion/quaternion.py
```diff
@@ -616,6 +674,209 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+        """
+        ax = self.axis.unit
+        ang = self.angle
+        magnitude = (0.75 * (ang - np.sin(ang))) ** (1 / 3)
+        homo = ax * magnitude
+        return homo
```
hakonanes, line 876:
```suggestion
        vec = ax * magnitude
        return vec
```

orix/tests/quaternion/test_conversions.py
hakonanes: I honestly cannot say. I added these tests in https://github.com/pyxem/orix/commit/477bf4ee33b1d08453f02f67e446625fbbe25d7e, but I don't give any attribution, which I should have done.

orix/quaternion/quaternion.py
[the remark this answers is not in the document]
argerlt: I like this, it takes one line instead of three, but is still visible when quickly scrolling.

orix/quaternion/_conversions.py
```diff
@@ -706,3 +749,101 @@ def eu2qu(eu: np.ndarray) -> np.ndarray:
...
+        qu[2] = 0.5 * np.sqrt(q2_almost)
+
+    if q3_almost < eps:
+        qu[3] = 0
+    elif om[1, 0] < om[0, 1]:
+        qu[3] = -0.5 * np.sqrt(q3_almost)
+    else:
+        qu[3] = 0.5 * np.sqrt(q3_almost)
+
+    return qu
```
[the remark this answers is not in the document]
argerlt: Actually, nevermind. i thought about this more and I think it should be normalized here. It's faster and more consistent with how the rest of `_conversions.py` operates.

Let me know if you disagree though and I can change it back, it's a quick 5 minute switch, and I still could go either way.

orix/quaternion/_conversions.py
```diff
@@ -498,12 +503,55 @@ def ax2qu_2d(ax: np.ndarray) -> np.ndarray:
...
+def ax2qu(axes: np.ndarray, angles: np.ndarray) -> np.ndarray:
     """N-dimensional wrapper for ax2qu_2d, see the docstring of that
-    function.
+    function for further details.
+
+    Parameters
+    ----------
+    axes
+        (...,3) dimensional numpy array of (x, y, z) vectors.
+
```
hakonanes, line 514:
```suggestion
```

orix/quaternion/_conversions.py
```diff
@@ -706,3 +754,282 @@ def eu2qu(eu: np.ndarray) -> np.ndarray:
...
     return qu
+
+
+@nb.jit("float64[:](float64[:, :])", cache=True, nogil=True, nopython=True)
+def om2qu_single(om: np.ndarray) -> np.ndarray:
+    """Convert a single (3,3) rotation matrix into a unit quaternion
+
+    Parameters
+    ----------
+    ma
```
hakonanes, line 765:
```suggestion
    om
```

orix/quaternion/_conversions.py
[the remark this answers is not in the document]
hakonanes: I agree, best to normalize here!

orix/quaternion/_conversions.py
```diff
@@ -30,6 +30,8 @@
 import numba as nb
 import numpy as np
 
+_FLOAT_EPS = np.finfo(float).eps
+_PI = np.pi
```
hakonanes, line 34: Can you confirm a speed-up of caching pi here? I suspect it is over-optimization and that we don't gain much. I suggest to drop it.

orix/quaternion/_conversions.py
```diff
@@ -30,6 +30,8 @@
 import numba as nb
 import numpy as np
 
+_FLOAT_EPS = np.finfo(float).eps
```
hakonanes, line 33: It's safe to drop the underscore since this entire file is private.

orix/quaternion/quaternion.py
```diff
@@ -616,6 +674,134 @@ def identity(cls, shape: Union[int, tuple] = (1,)) -> Quaternion:
...
+        axis
+            The axes of rotation.
+        angle
+            The angles of rotation.
+
+        Examples
+        --------
+        #TODO
+        """
+        return self.axis, self.angle
```
hakonanes, line 747: Remember to account for `degrees`.

orix/tests/quaternion/test_quaternion.py
```diff
@@ -477,16 +477,16 @@ def test_get_rotation_matrix_from_diffpy(self):
...
-        axangle = AxAngle.from_rotation(rotations)
-        quat2 = Quaternion.from_neo_euler(axangle)
-        quat3 = Quaternion.from_axes_angles(axangle.axis.data, axangle.angle)
-        assert np.allclose(quat2.data, quat3.data)
-
-        quat4 = Quaternion.from_axes_angles(
-            axangle.axis.data, np.rad2deg(axangle.angle), degrees=True
-        )
-        assert np.allclose(quat4.data, quat3.data)
+# class TestFromAxesAngles:
```
hakonanes, line 480: Why are these commented out?

hakonanes (changes requested):
Good stuff. Thank you for sorting out most of my comments. There are a few things still left to do in this PR, I've pointed out most of them. Also, some previous comments are still not resolved.

Could you run black version 23.1.0 on the files you've changed to ensure a consist code style? The changes that are needed are reported by our [CI from this PR](https://github.com/pyxem/orix/actions/runs/5182252624/jobs/9338822308?pr=450).

There are some uncovered lines (see the coverage step in our CI). How to write and run tests (also for Numba functions) and check coverage are described in our [contributing guide](https://orix.readthedocs.io/en/stable/dev/running_writing_tests.html). Please let me know if you need any help on this.

orix/quaternion/_conversions.py
argerlt: I'm not sure how to test it's speedup efficiently and I'm fine with dropping it. This was a holdover from a different project I was working on at the same time as ORIX with different preferences.

orix/quaternion/quaternion.py
argerlt: good catch, thanks

orix/tests/quaternion/test_quaternion.py
argerlt: This test used neo_euler, which originally I was intending to remove. That is no longer the case though, so no need to have it commented out.

argerlt:
Okay, I believe this Pull request is now done, barring any further reviewer feedback

There are a handful of things that could still be done
 - ` _conversions.ro2qu` and `_conversions.qu2ro`
 - fully replace `NeoEulerian` with equivalent to/from calls to quaternions
 - replace `Quaternion.axis` and `Quaternion.angle` shortcuts with jit-accelerated versions
 
 However, this PR has been open a very long time, and the above are not pressing matters. If anyone wants to add them, feel free, otherwise I am fine merging this and moving on for the time being.

orix/quaternion/_conversions.py
hakonanes: If I remember correctly, I've tested caching such variables in kikuchipy in some Numba accelerated functions, and found that caching has a minimal impact on speed.

hakonanes:
Thanks for again accomodating my review comments and suggestions, @argerlt. I agree that the PR is in good shape and that functionality is tested and documented. However, there are some style inconsistencies in both syntax and docstrings. Are you OK with me pushing some commits to your branch fixing these?

argerlt:
Sure thing, go ahead.

hakonanes:
I suggest we remove the `Quaternion.from_rodrigues_frank()` method and change `Quaternion.from_rodrigues()` either of these two ways:
1. Allow an optional `omega` parameter for scaling the axis (combined being the Rodrigues-Frank vector). If users have their 4-component vectors in one array, they have to split it to pass it in. Only a minor hassle, I'd say.
2. Allow `ro` to have either 3 (vector) or 4 (vector and scaling) as the final array dimension. Drawback could be that users mistake where the scaling should be (first or last).

@argerlt, what do you think?

I'm still going through the PR and making small adjustments here and there. I especially appreciate that you had a look at 3Drotations and have attempted to harmonize orix with output from this package.

argerlt:
I went back and forth on this issue a lot, and I do think you are right, creating `Quaternion.from_rodrigues_frank()' was the wrong solution. 

I like option one. It causes users inputting a 4-vector to make a conscious decision one way or another, which will probably help avoid errors, and its a method by which users who only know of the 3-vector version won't become unnecessarily confused.

I think option two might end up being a headache to implement, as there will be inevitable edge cases from accepting either a length 3 or 4 vector. 

That said, I think both would work fine.

hakonanes:
It seems to me like the fourth component in the Rodrigues-Frank vector accepted to `Quaternion.from_rodrigues_frank(ro)` (to be removed) is not the angle omega but the scaling factor tan(omega / 2). Is this correct?

hakonanes:
Yes, I see that you interpret omega as being the scaling factor in the docstring of that method. Good.

hakonanes:
@argerlt, if it's fine by you, I'll rebase on `develop` and fix the merge conflict. The docs should then pass.

argerlt:
Works for me. Thanks again, and sorry inlet this sit so long.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

hakonanes:
No problem! I've been off development myself lately.

argerlt:
@hakonanes , code has been rebased and everything seems to be working as expected. let me know if there are any other changes you would like to see.

hakonanes (approved):
This looks good to me now. Thank you for your efforts, @argerlt!

The added from/to methods should make it simpler to work with neo-eulerian rotation representations in orix.

The new tests are also a great to have going forward!

PR #457, opened 2023-08-29
add to_dict method to crystal_map to allow for easy extraction of tho…

SteffenBrinckmann:
add to_dict method to crystal_map to allow for easy extraction of those results that would be printed with __repr__

#### Progress of the PR
- [X] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
       -> most tests were successful; some failed because of some 'url.requests' issue.
      -> copied existing code: hence style should be ok (I don't have black installed and have no experience of using it)

```python
>>> from orix.io import load
>>> datadir = 'data/04/'
>>> fname = 'sdss_ferrite_austenite.ang' #.ang file produced by EMsoft
>>> cm = load(datadir + fname)
>>> metadata = cm.to_dict()
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/crystal_map/crystal_map.py
+++ b/orix/crystal_map/crystal_map.py
@@ -749,6 +749,43 @@ class CrystalMap:
 
         return representation
 
+    def to_dict(self) -> dict[dict[str,str]]:
+        """Return dictionary of the data."""
+        result = {}
+        if self.size == 0:
+            return result
+
+        phases = self.phases_in_data
+        phase_ids = self.phase_id
+
+        # Ensure attributes set to None are treated OK
+        names = ["None" if not name else name for name in phases.names]
+        sg_names = ["None" if not i else i.short_name for i in phases.space_groups]
+        pg_names = ["None" if not i else i.name for i in phases.point_groups]
+        ppg_names = [
+            "None" if not i else i.proper_subgroup.name for i in phases.point_groups
+        ]
+
+        # Determine dictionary of phases
+        unique_phases = np.unique(phase_ids)
+        for i, phase_id in enumerate(unique_phases.astype(int)):
+            p_size = np.where(phase_ids == phase_id)[0].size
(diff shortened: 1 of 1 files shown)


hakonanes:
Thank you for this suggestion, @SteffenBrinckmann.

A public method to extract all contents of a crystal map is a good idea. We already use such a private method in the orix HDF5 writer:

https://github.com/pyxem/orix/blob/2e6f783e07c750f09553bec6fc51f5dfa9abae96/orix/io/plugins/orix_hdf5.py#L248-L298

This function returns *all* contents of a crystal map, including the rotations. It doesn't calculate the phase fractions for you, though. Can you explain what you need this information for, specifically?

I think making the mentioned private function a public crystal map method is the best solution. If you really need the phase fractions as well, we can add this to the dictionary if a boolean keyword is true, for example. What do you think?

SteffenBrinckmann:
hey @hakonanes 
- The fractions are not that difficult to evaluate and does warrant an extra function. 
- I did not know about that function existing, but yes one function makes more sense.
- I would just suggest to not have the function in a writer (it is not about a writer) but it is for interaction, so a more general location would make more sense. (I would have never guessed the function being in the hdf writer).

best, Steffen

CSSFrancis:
I think that @SteffenBrinckmann has a good idea.  If the function is private it would be good to add this function to the CrystalMap object.  Something like a function  `_to_dictionary` would allow it to be saved by hyperspy signals as well which would be useful.  

https://github.com/hyperspy/hyperspy/blob/3b350ff80b68b0928b284e76d3dc0a2b31e6dbf5/hyperspy/misc/utils.py#L524C37-L524C48

@hakonanes @pc494 Thoughts?

hakonanes:
Haven't forgotten about this, but don't have time at the moment. A more thorough discussion about serialization (into dicts, I presume) would be good.

PR #461, opened 2023-10-26
Performance: Speed up of deepcopy for Miller

CSSFrancis:

Speeds up deep copy for `Miller` class.  Current deep copy is slowed down by using the `deepcopy` function. 

This causes issues in `diffsims` when we have to repeatedly copy and rotate some `ReciprocalLatticeVector`

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/vector/miller.py
+++ b/orix/vector/miller.py
@@ -426,7 +426,10 @@ class Miller(Vector3d):
 
     def deepcopy(self) -> Miller:
         """Return a deepcopy of the instance."""
-        return deepcopy(self)
+        data = deepcopy(self.data)  # Otherwise, data is a view
+        copied = self.__class__(xyz=data, phase=self.phase) # deepcopy is slow.
+        copied.coordinate_format = self.coordinate_format
+        return copied
 
     def round(self, max_index: int = 20) -> Miller:
         """Round a set of index triplet (Miller) or quartet


hakonanes:
Thank you for looking into making deepcopying faster.

Your solution by only deepcopying the vectors and not the phase, which is not in line with the current behavior. The phase should not be shared. If deepcopying the phase is a bottleneck we should look into improving the speed of this method instead.

hakonanes:
I realize now that copying only the data is what you want in the case you describe... Are you happy with a parameter `deepcopy(data_only=False)`? If true, your current solution is used. It is false by default, retaining the current behavior.

CSSFrancis:
@hakonanes let me look into this more. I think deepcopying only the data is a good solution but I'll look into the phase deepcopy as well.

hakonanes:
A sidenote: I've considered replacing `deepcopy()` with just `copy()` (and deprecating deepcopy() one minor release). The behavior is the same. But the naming is in line with NumPy's interpretation of a "copy" (not shared memory) and a "view" (shared memory). What do you think?

harripj:
@CSSFrancis @hakonanes thanks for looking into this!

I think the `deepcopy` API should return a deep copy of all relevant data in this case. Whilst this may lead to speedups, I think this could be a footgun for the majority of users who would not expect any memory to be shared when using the API.

Are there other examples in other codebases where `deepcopy`returns partial copies?

If the speed ups are useful in diffsims then I think this PR should be an internal optimisation rather than in orix, such that the copying scope is well-defined and controlled, ie. it is known that `phase` is shared memory.

hakonanes:
Thank you for your input, @harripj. I agree with you on all points. The deepcopy method is meant to be used by end users or in the setup of more demanding methods, not repeatedly inside a loop or similar.

But if we can get any speed up that would be nice.

argerlt:
@CSSFrancis , are you still interested in this or is it okay to close?

argerlt:
No response in 50 days, so I am closing this PR. Feel free to reopen this issue if anyone disagrees.

However, given how stale this PR is, it might make more sense to reopen as a new issue assuming it's still a bottleneck in diffsims.

Issue #480, opened 2024-04-07
Allow passing `Phase` instead of `Symmetry`


hakonanes:
As @CSSFrancis pointed out in https://github.com/pyxem/orix/pull/478#issuecomment-2041084910, it may sometimes be simpler to pass a `Phase` with a point group set, rather than extracting the corresponding `Symmetry`, to functions like the sampling functions. Since a `Phase` only has one `Symmetry`, this should not be a problem.

I'm for this. What do other people think?

harripj:
I think this would be a convenient addition to `sampling.get_sample_reduced_fundamental`. Would this also be applicable for other parts of the API?

hakonanes:
That was my intention, yes,wherever it makes sense.

hakonanes:
I'm not sure of the best implementation. Should we allow passing a `Phase` to functions with a `point_group` parameter? What about to the `space_group` parameter? Or should we replace these parameters with one which can be all these types? I haven't thought this through, but realize it requires some thought 🤔.

harripj:
Would it be enough to improve the docstrings and examples with an example extracting the `point_group` from a `Phase` object?

PR #481, opened 2024-04-07
New Feature: Allow passing phase to sample generation

CSSFrancis:

Adds #480 for sampling generators. 

#### Progress of the PR

```python
a = 5.431
latt = Lattice(a, a, a, 90, 90, 90)
atom_list = []
for coords in [[0, 0, 0], [0.5, 0, 0.5], [0, 0.5, 0.5], [0.5, 0.5, 0]]:
    x, y, z = coords[0], coords[1], coords[2]
    atom_list.append(
        Atom(atype="Si", xyz=[x, y, z], lattice=latt)
    )  # Motif part A
    atom_list.append(
        Atom(atype="Si", xyz=[x + 0.25, y + 0.25, z + 0.25], lattice=latt)
    )  # Motif part B
struct = Structure(atoms=atom_list, lattice=latt)
p = Phase(structure=struct, space_group=227)
rotations = get_sample_reduced_fundamental(resolution=4, phase=phase)

```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/sampling/sample_generators.py
+++ b/orix/sampling/sample_generators.py
@@ -33,6 +33,7 @@ def get_sample_fundamental(
     resolution: Union[int, float] = 2,
     point_group: Optional[Symmetry] = None,
     space_group: Optional[int] = None,
+    phase: Optional[Phase] = None,
     method: str = "cubochoric",
     **kwargs
 ) -> Rotation:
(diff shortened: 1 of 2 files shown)


hakonanes:
Thanks for this, @CSSFrancis.

But, I suggest to wait with this until after the v0.12.0 release. As @harripj asked and I agreed with in #480, we should allow this all places where it makes sense.

Also, I'm not sure of the best implementation. Should we allow passing a `Phase` to the `point_group` parameter? What about to `space_group` parameter? Or should we replace these parameters with one which can be all these types?

CSSFrancis:
@hakonanes That's fair!  We can just make the 0.12.0 release and then figure this out later if you think that is a good idea. 

I was just trying to follow what was there previously.  We should probably have one parameter that checks if it is a point group, a space group or phase or something that checks to make sure only one is defined.

PR #483, opened 2024-04-10
Numpy 2.0.0 Fixes

CSSFrancis:
Running:
```ruff orix/ --select NPY201 --preview```

```
orix/quaternion/symmetry.py:420:40: NPY201 [*] `np.infty` will be removed in NumPy 2.0. Use `numpy.inf` instead.
Found 1 error.
```


pc494 (approved):
Good catch.

CSSFrancis:
@pc494 Just a note I checked both diffsims and pyxem and they appear to be good

PR #491, opened 2024-04-17
Discrepancy between Phase.from_cif and Phase(structure)

dasilvaale:
#### Description of the change
Hello,

there appears to be a discrepancy in the atom xyz positions between creating a Phase using from_cif and assigning a structure directly (Phase(structure)).

This appears to be because lattice alignment is perfomed twice when using from_cif, once in the method itself and once in the Phase.structure setter.

I have removed the alignment in Phase.from_cif, as it will be called in the Phase constructor and added some checks to the test below, which were previously failing with the double lattice alignment.

#### Progress of the PR
- [-] [Docstrings for all functions](https://numpydoc.readthedocs.io/en/latest/example.html)
- [x] Unit tests with pytest for all lines
- [x] Clean code style by [running black via pre-commit](https://orix.readthedocs.io/en/latest/dev/code_style.html)

#### Minimal example of the bug fix or new feature
```python
from tempfile import NamedTemporaryFile

from orix.crystal_map import Phase
from diffpy.structure import loadStructure
import numpy as np

# from test
file_contents = """
#======================================================================

# CRYSTAL DATA

#----------------------------------------------------------------------

data_VESTA_phase_1


_chemical_name_common                  ''
_cell_length_a                         15.50000
_cell_length_b                         4.05000
_cell_length_c                         6.74000
_cell_angle_alpha                      90
_cell_angle_beta                       105.30000
_cell_angle_gamma                      90
_space_group_name_H-M_alt              'C 2/m'
_space_group_IT_number                 12

loop_
_space_group_symop_operation_xyz
   'x, y, z'
   '-x, -y, -z'
   '-x, y, -z'
   'x, -y, z'
   'x+1/2, y+1/2, z'
   '-x+1/2, -y+1/2, -z'
   '-x+1/2, y+1/2, -z'
   'x+1/2, -y+1/2, z'

loop_
   _atom_site_label
   _atom_site_occupancy
   _atom_site_fract_x
   _atom_site_fract_y
   _atom_site_fract_z
   _atom_site_adp_type
   _atom_site_B_iso_or_equiv
   _atom_site_type_symbol
   Mg(1)      1.0     0.000000      0.000000      0.000000     Biso  1.000000 Mg
   Mg(2)      1.0     0.347000      0.000000      0.089000     Biso  1.000000 Mg
   Mg(3)      1.0     0.423000      0.000000      0.652000     Biso  1.000000 Mg
   Si(1)      1.0     0.054000      0.000000      0.649000     Biso  1.000000 Si
[...]

PR #493, opened 2024-04-18
`ax2qu` does not raise anymore if last dimension is `1`

harripj:

`ax2qu` conversion function throws if the arrays are multidimensional and last axis size is 1, even though they are broadcastable.

I appreciate having a redundant axis in an array is not ideal, but sometimes it is easier to work with it to keep the input and output shapes consistent 😄.

I have looked into the [code](https://github.com/pyxem/orix/blob/856491e341efc695da7b483fd108491a425f884f/orix/quaternion/_conversions.py#L535-L582) and there appears to be a lot of shape juggling because the internal functions are numba jitted and ultimately require 2d arrays. These lines in particluar seem to be the issue: https://github.com/pyxem/orix/blob/856491e341efc695da7b483fd108491a425f884f/orix/quaternion/_conversions.py#L557-L561

I believe we should also check here if `axes.shape[:-1] == angles.shape` and this will fix the issue.

#### Progress of the PR
- [-] [Docstrings for all functions](https://numpydoc.readthedocs.io/en/latest/example.html)

```python
>>> import numpy as np

>>> axes = np.random.randn(2, 1, 3)
>>> angles = np.random.randn(*axes.shape[:-1])

>>> Quaternion.from_axes_angles(axes, angles)
ValueError: The dimensions of axes (2, 1) and angles (2,) are incompatible. The dimensions must match or one must be a singular value.
```

With the fix in this PR:
```python
>>> Quaternion.from_axes_angles(axes, angles)
Quaternion (2, 1)
[[[ 0.9999 -0.0058 -0.0025 -0.0078]]

 [[ 0.9386 -0.1434 -0.1071  0.295 ]]]
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


pc494:
This looks good to me, will leave it to @hakonanes to review and merge though.

hakonanes (approved):
We'll spotted!

We can include it in the 0.12.1 release if you change the base branch to `main`. I'll merge #491 after this PR.

harripj:
@hakonanes thanks for reviewing! Yes it would be great to include this in the patch release. I've rebased this PR.

PR #499, opened 2024-05-07
BugFix: Allow get_kwargs for setting kwargs on __mul__

CSSFrancis:
This should fix (part) of the problems upstream with https://github.com/pyxem/diffsims/issues/211 by allowing the function `get_kwargs` to be overwritten in subclasses and change the order/what is passed to the __init__ fuction.

#### Minimal example of the bug fix or new feature (This now works)
```python
from diffsims.crystallography import ReciprocalLatticeVector


phase = Phase(point_group="m-3m")
g = ReciprocalLatticeVector(phase, [1, 1, 1])

R = Rotation.random()
g1 = R * g

# CHanges to RLV
# ---------------- 

    def get_kwargs(self, new_data, *args, **kwargs):
        return dict(phase=self.phase, xyz=new_data)
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -240,7 +240,8 @@ class Quaternion(Object3d):
                 m.coordinate_format = other.coordinate_format
                 return m
             else:
-                return other.__class__(v)
+                kwargs = other.get_kwargs(self)
+                return other.__class__(v, **kwargs)
         return NotImplemented
 
     def __neg__(self) -> Quaternion:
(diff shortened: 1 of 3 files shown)


hakonanes:
Thanks for suggesting a fix to the issue in diffsims. But, I've raised some doubts regarding our approach in the referenced issue https://github.com/pyxem/diffsims/issues/211#issuecomment-2099218511.

CSSFrancis:
@hakonanes I think this should be possible.  It works quite nicely with everything and makes the logic __much__ simpler.  Otherwise we end up having to struggle to track the rotation alongside the RLV from `diffsims` --> `pyxem` and it gets complicated.

Issue #506, opened 2024-07-28
Numpy 2.0.0 problems


pc494:
This has filtered down from upstream (see https://github.com/pyxem/pyxem/issues/1099). `quaternion` on which we depend is currently (and may not for some time be) compatible with `numpy 2.0.0`. 

It's possible we could drag and drop a sister package in place (see discussion https://github.com/moble/quaternion/issues/229) but I'm acutely aware this project is really short of dev resourcing.

The alternative is that we make ourselves `numpy 1.x`  for now.

CSSFrancis:
@pc494 I'm going to try to make the switch this week when I have the time! (Which I should)

hakonanes:
As I mentioned in #507:

> The dependency on numpy-quaternion is NOT important. We only use it for quaternion-quaternion and quaternion-vector multiplication. Our "old" implementation used Numba, and as far as I can remember, it was only twice as slow.
> If numpy-quaternion becomes a hassle, I'd suggest we make it an optional dependency: use it when available, but fall back to Numba if it's not. What do you think?

pc494:
Sorry @hakonanes I had completely forgotten the #507 discussion when I opened this.

hakonanes:
Nono, you opened this long before my post, I just reposted here to say that we're on it (:

hakonanes:
orix 0.13.1 is now available from PyPI, with numpy-quaternion as an optional dependency.

But whaddayaknow, latest diffpy.structure is not compatible with NumPy 2.0: https://github.com/diffpy/diffpy.structure/issues/68.

So orix is still not compatible with NumPy 2.0.

pc494:
patch to 0.13.2 with a minimum version on `diffpy.structure`?

hakonanes:
After they release a new diffpy.structure release compatible with NumPy 2.0? Yes, that sounds like a plan.

hakonanes:
With #525 and an official release of diffpy.structure v3.2.2 (currently an RC out https://pypi.org/project/diffpy.structure/3.2.2rc0/), orix is compatible with NumPy 2.0 (without numpy-quaternion).

hakonanes:
Now that diffpy.structure v3.2.2 is available on PyPI (https://pypi.org/project/diffpy.structure/) and conda-forge (https://anaconda.org/conda-forge/diffpy.structure), we can release orix v0.13.1 which is fully compatible with NumPy v2.

hakonanes:
v0.13.2 I mean

PR #507, opened 2024-07-29
Replace numpy-quaternion with quaternionic.

CSSFrancis:
As per #506.

I'm running into some problems making this change becuase I'm not really sure what the  improper `setter` functions works with the `_data` attribute.

The difference arrises from quaternionic forcing having 4 as the last shape and not automatically casting as the numpy-quaternion package does.

```python

import quaternionic as quaternion2
import quaternion
import numpy as np

quaternion.as_float_array(np.ones((10,4))).shape 
# (10, 4, 4)

quaternion2.array(np.ones((10,4))).shape
# (10,4)
```

@hakonanes any advice for what might need to change as a result?

#### Progress of the PR

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/quaternion/quaternion.py
+++ b/orix/quaternion/quaternion.py
@@ -122,6 +122,10 @@ class Quaternion(Object3d):
 
     dim = 4
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._data = quaternion.array(self.data)
+
     # -------------------------- Properties -------------------------- #
 
     @property
@@ -133,12 +137,12 @@ class Quaternion(Object3d):
         value : numpy.ndarray
             Scalar quaternion component.
         """
-        return self.data[..., 0]
+        return self.data.w
 
     @a.setter
     def a(self, value: np.ndarray):
         """Set the scalar quaternion component."""
-        self.data[..., 0] = value
+        self.data.w = value
 
     @property
     def b(self) -> np.ndarray:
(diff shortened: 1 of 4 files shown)


CSSFrancis:
Okay so we might have to adjust how we do things. 

1. I think that the data for a Quaternion should always be a `quaternion` class from quaternionic. 
2. The rotation class should be Quaternion + a boolean for proper/ improper.  I think the right thing to do would be to make a numpy array class for this but.... Maybe we should just sperate these two arrays into data + improper.

pc494:
It looks like we might get a downstream fix now, which would be excellent.

https://github.com/moble/quaternion/issues/229

hakonanes:
Sorry for answering late here...

The dependency on numpy-quaternion is NOT important. We only use it for quaternion-quaternion and quaternion-vector multiplication. Our "old" implementation used Numba, and as far as I can remember, it was only twice as slow.

If numpy-quaternion becomes a hassle, I'd suggest we make it an optional dependency: use it when available, but fall back to Numba if it's not. What do you think?

hakonanes:
And I should say, I find storing the rotation handedness (proper -> right-handed, improper -> left-handed; or, if you will, the rotation followed by an inversion) as a fifth value 0 or 1 in the `Rotation._data` array is rather elegant and makes the implementation easier.

pc494:
I think this is safe to close.

@CSSFrancis any objection?

Issue #516, opened 2024-09-07
Using setup.py as a command line tool is deprecated


hakonanes:
When installing orix in editable mode, I get the following warning:

```bash
$ pip install -e ".[dev]"
[...]
Installing collected packages: orix
  Attempting uninstall: orix
    Found existing installation: orix 0.13.dev1
    Uninstalling orix-0.13.dev1:
      Successfully uninstalled orix-0.13.dev1
DEPRECATION: Legacy editable install of orix[dev]==0.14.dev0 from file:///Users/hakon/kode/personal/orix (setup.py develop)
is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable
--use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using
--config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can
be found at https://github.com/pypa/pip/issues/11457
  Running setup.py develop for orix
Successfully installed orix-0.14.dev0
```

We should replace setup.py with pyproject.toml.

hakonanes:
Had to correct the title, because setup.py as a *configuration file* for setuptools and other build backends is not deprecated (https://packaging.python.org/en/latest/discussions/setup-py-deprecated/).

PR #517, opened 2024-09-07
Change build backend from setuptools to Hatch

hakonanes:
Addresses #516 and changes build tool from setuptools to [Hatch](https://hatch.pypa.io/latest/).

Changing from setup.py to pyproject.toml and running `python -m build` produces the same wheel as before. The 50 MB-zip with everything is nearly identical as well, even though Hatch doesn't use the MANIFEST.in file. This tells me we don't need it, so I removed it. The `manifix` package (and CI check) is then unnecessary, which is good, because it requires a `setup.py` file anyway.

I removed most of the metadata in `orix/__init__.py`, as the build only needs it in pyproject.toml. The version is left because it is searched by `importlib_metadata.version("orix")`. I also left the credits list, although this isn't used by any tool. Still nice to have there.

Split out coverage dependencies `coverage.py` and `pytest-cov` from `orix[tests]` into `orix[coverage]`.

Will publishing to PyPI go as smoothly as it did? I don't see why not, but cannot say for sure until we try...

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


pyproject.toml
```diff
@@ -0,0 +1,109 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "orix"
+# conf.py assumes orix developers to be first author
+authors = [{name = "orix developers", email = "pyxem.team@gmail.com"}]
```
pc494, line 8: Is the email mandatory here? If not maybe drop it as I'd much rather things came in to the Github

pc494 (approved):
LGTM, one minor admin thing around the email, happy for @hakonanes to decide and merge as he sees fit.

pyproject.toml
hakonanes: Not mandatory, I'll remove it.

Issue #515, opened 2024-09-04
Skip compatibility with Python 3.8 and 3.9?


hakonanes:
Do we loose much by skipping Python 3.8 and 3.9 compatibility? If we support 3.10 and 3.11 (with the plan to support 3.12), that is perhaps enough?

If yes, I'll make the necessary changes.

pc494:
I have long held that we are too generous with backwards Python compatibility. Python 3.10 is October 2021 which seems good enough for me, writing in September 2024.

CSSFrancis:
My reasoning is always if numpy says they can't [support](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) it can we reasonably say that we support it?

hakonanes:
Seems like our dependencies support Python 3.12. So, I suggest to start testing against 3.12 instead of 3.10.

pc494:
Happy with this, especially as py3.12 is now nearly a year old

PR #519, opened 2024-09-08
Make numpy-quaternion and optional dependency

hakonanes:
This PR makes numpy-quaternion an optional dependency. It can be installed using `pip install orix[all]` or `pip install orix numpy-quaternion`.

I've made a private `orix.constants.installed` dictionary with (package, installed bool) available. Whether a dependency is installed or not is then checked with `installed["numpy-quaternion"]`.

Functionality previously done by numpy-quaternion can now be done using Numba's [generalized universal functions](https://numba.readthedocs.io/en/stable/user/vectorize.html#the-guvectorize-decorator), which handles broadcasting in a smart way. Reduces need for us to check shapes.

**How should we handle the optional dependencies on anaconda? I suggest to make an `orix-core` available, without numpy-quaternion. `orix` will still have it. What do people think?**

#### Progress of the PR
- [ ] Improve installation options now that we have optional dependencies

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


hakonanes:
tomllib requires Python 3.11... Will see if we don't need it. Guess we only have to list optional dependencies in `orix.constants.installed`.

hakonanes:
The failing tests for Python 3.8 are expected and will be addressed in a separate PR addressing #515.

hakonanes:
Wanted to benchmark the re-introduced functionality against numpy-quaternion. In an environment with numpy-quaternion available, I think this is a fair benchmarking of the code used:

```python
import numpy as np
import numba as nb
import quaternion

from orix.quaternion import Quaternion

n_v = 10
n_qu = 1000
v = Vector3d(np.random.random(n_v * 3).reshape((n_v, 3)))
Q = Quaternion.random(n_qu)


# Called when numpy-quaternion is not available
@nb.guvectorize("(n),(m)->(m)", cache=True)
def qu_rotate_vec_gufunc(
    qu: np.ndarray, v1: np.ndarray, v2: np.ndarray
) -> None:  # pragma: no cover
    a, b, c, d = qu
    x, y, z = v1
    tx = 2 * (c * z - d * y)
    ty = 2 * (d * x - b * z)
    tz = 2 * (b * y - c * x)
    v2[0] = x + a * tx - d * ty + c * tz
    v2[1] = y + d * tx + a * ty - b * tz
    v2[2] = z - c * tx + b * ty + a * tz


@nb.guvectorize("(n),(n)->(n)", cache=True)
def qu_multiply_gufunc(
    qu1: np.ndarray, qu2: np.ndarray, qu12: np.ndarray
) -> None:  # pragma: no cover
    qu12[0] = qu1[0] * qu2[0] - qu1[1] * qu2[1] - qu1[2] * qu2[2] - qu1[3] * qu2[3]
    qu12[1] = qu1[1] * qu2[0] + qu1[0] * qu2[1] - qu1[3] * qu2[2] + qu1[2] * qu2[3]
    qu12[2] = qu1[2] * qu2[0] + qu1[3] * qu2[1] + qu1[0] * qu2[2] - qu1[1] * qu2[3]
    qu12[3] = qu1[3] * qu2[0] - qu1[2] * qu2[1] + qu1[1] * qu2[2] + qu1[0] * qu2[3]


# numpy-quaternion implementation
%timeit \
    qu = quaternion.from_float_array(Q.data);\
    quaternion.rotate_vectors(qu, v.data)
# 69.8 μs ± 571 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%timeit \
    qu = quaternion.from_float_array(Q.data);\
    qu = quaternion.from_float_array(Q.data);\
    quaternion.as_float_array(qu * qu)
# 4.05 μs ± 35.5 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

# Alternative implementation
#%%
%timeit \
    v2 = np.empty(Q.shape + v.shape + (3,));\
    qu_rotate_vec_gufunc(Q.unit.reshape(-1, 1).data, v.reshape(1, -1).data, v2);\
    v2 = v2.reshape(*Q.shape, *v.shape, 3)
# 92.2 μs ± 988 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%timeit \
    qu12 = np.empty_like(Q.data);\
    qu_multiply_gufunc(Q.data, Q.data, qu12)
# 37.4 μs ± 170 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

So, numpy-quaternion beats orix with Numba 0.60.0 and NumPy 1.26.4 by:
* quaternion-vector outer: 1.3x
* quaternion-quaternion multiplication: 9x

pc494:
Sorry, realise I'm blocking here, plan is to get to it tomorrow/Friday :)

hakonanes:
That would be great, thanks! Even though all tests pass w/o numpy-quaternion, I get different behavior from Vector3d.get_circle() that I have to fix.

hakonanes:
Fixed the issue with Vector3d.get_circle(). Just need to update tests and the changelog.

hakonanes:
Hope to finish testing and changelog updates within Monday. Should be OK to review as is, I don't expect any code changes.

doc/user/installation.rst
```diff
@@ -13,10 +13,18 @@ With pip
 ========
 
 orix is availabe from the Python Package Index (PyPI), and can therefore be installed
-with `pip <https://pip.pypa.io/en/stable>`__. To install, run the following::
+with `pip <https://pip.pypa.io/en/stable>`__.
+To install all of orix' functionality, do::
```
pc494, line 17: extremely boring typo, should be `orix's`

doc/user/installation.rst
```diff
@@ -35,16 +43,23 @@ To install with Anaconda, we recommend you install it in a `conda environment
...
 
-   conda create --name orix-env python=3.9
+   conda create --name orix-env python=3.11
    conda activate orix-env
 
 If you prefer a graphical interface to manage packages and environments, you can install
 the `Anaconda distribution <https://docs.continuum.io/anaconda>`__ instead.
 
-To install::
+To install all of orix' functionality, do::
```
pc494, line 52: maybe not a typo then?

orix/quaternion/quaternion.py
```diff
@@ -1237,3 +1239,59 @@ def _outer_dask(
         new_chunks = tuple(chunks1[:-1]) + tuple(chunks2[:-1]) + (-1,)
 
         return out.rechunk(new_chunks)
+
```
pc494, line 1242: Would this be obvious if I thought about it?

orix/constants.py
```diff
@@ -0,0 +1,19 @@
...
+optional_deps: list[str] = ["numpy-quaternion"]
+installed: dict[str, bool] = {}
+for pkg in optional_deps:
+    try:
+        _ = version(pkg)
+        installed[pkg] = True
+    except ImportError:
+        installed[pkg] = False
+
+# Tolerances
```
pc494, line 15: maybe 1 line here about why we have/want to use two of them (I notice that we do use both)

pc494 (approved):
This LGTM, couple of tiny things for you to look at and action/ignore @hakonanes.

orix/quaternion/quaternion.py
pc494: (as in the stuff below this line)

doc/user/installation.rst
hakonanes: I like typos, means I get to know the language better! I thought words ending with x shouldn't have the s, thank you for pointing this out (:

doc/user/installation.rst
hakonanes: An _intentional_ typo, if you will

orix/quaternion/quaternion.py
hakonanes: Perhaps not. I've added a comment-block to help with the understanding!

orix/constants.py
hakonanes: I've added an explanation. In general, I'd like for us to use the highest precision possible. That isn't always practical, so we have to lower it, but typically should use either of the two. I guess you could call them "high" and "lower" precisions.

doc/user/installation.rst
pc494: The guidance I have found simply says "However, if it sounds better, it is acceptable to just add the apostrophe." so that didn't help, although I would say orix's if I was speaking.

pc494:
Perfect, thanks @hakonanes I'm merging :)

PR #533, opened 2024-12-10
Avoid expensive deepcopy when rotating Miller

viljarjf:

Rotating a Miller instance with a Rotation calls `Miller.__getitem__` to handle improper rotations, regardless of whether any rotations are actually improper.
As  `Miller.__getitem__` involves a deepcopy, avoiding this when not necessary provides a significant speedup, depending on the complexity of the associated Phase.

#### Progress of the PR
- [n/a] [Docstrings for all functions](https://numpydoc.readthedocs.io/en/latest/example.html)

```python

P1 = Phase("P1", 1)
Fm3m = Phase("Fm3m", 225)
P1_vec = Miller.random(P1)
Fm3m_vec = Miller.random(Fm3m)
rot = Rotation.random(2000)


%timeit rot * P1_vec
%timeit rot * Fm3m_vec
```
**Current**
505 μs ± 18.6 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
1.7 ms ± 48.2 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

**New change**
346 μs ± 41.4 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
350 μs ± 16.9 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

Of course, with improper rotations, the runtime is the same:
```python

P1 = Phase("P1", 1)
Fm3m = Phase("Fm3m", 225)
P1_vec = Miller.random(P1)
Fm3m_vec = Miller.random(Fm3m)
rot = Rotation.random(2000) * -1

assert all(rot.improper)

%timeit rot * P1_vec
%timeit rot * Fm3m_vec
```
506 μs ± 26.3 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
1.69 ms ± 44.3 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/quaternion/rotation.py
+++ b/orix/quaternion/rotation.py
@@ -117,8 +117,9 @@ class Rotation(Quaternion):
             return Q
         if isinstance(other, Vector3d):
             v = Quaternion(self) * other
-            improper = (self.improper * np.ones(other.shape)).astype(bool)
-            v[improper] = -v[improper]
+            if np.any(self.improper):
+                improper = (self.improper * np.ones(other.shape)).astype(bool)
+                v[improper] = -v[improper]
             return v
         if isinstance(other, int) or isinstance(other, list):  # abs(1)
             other = np.atleast_1d(other).astype(int)


hakonanes:
Seems simple enough. I'll have a look within tomorrow.

hakonanes:
Looking beyond this PR, I think it may be better to avoid `deepcopy()` altogether and rather add a `copy()` method, like this

```python
def copy(self) -> Self:
    return self.__class__(self)
```

We can update the creation method to allow another instance as the first parameter, and only copy the things we need.

We can deprecate `deepcopy()` in favor of `copy()`.

Benefits:
* More control
* New copy via `Vector3d(Vector3d)` or `Vector3d.copy()`

What do you think, @viljarjf?

hakonanes (approved):

hakonanes:
Let's continue the copy discussion in #535.

hakonanes:
@viljarjf, newly released 0.13.3 has this change in.

Issue #535, opened 2024-12-15
Improve object copying


hakonanes:
Looking beyond #533, I think it may be better to avoid `deepcopy()` altogether and rather add a `copy()` method, like this

```python
def copy(self) -> Self:
    return self.__class__(self)
```

We can update the creation method to allow another instance as the first parameter, and only copy the things we need.

We can deprecate `deepcopy()` in favor of `copy()`.

Benefits:
* More control
* New copy via `Vector3d(Vector3d)` or `Vector3d.copy()`

What do you think, @viljarjf?

_Originally posted by @hakonanes in https://github.com/pyxem/orix/issues/533#issuecomment-2543893484_

@CSSFrancis?

viljarjf:
Sounds like a good idea to me! 

Looks like this is almost entirely supported already, apart from copying the data. Most constructors already correctly handle the first parameter being an instance of itself, but the data is shared rather than copied. This is in line with how regular (not deep) copying is intended to work in python, as per https://docs.python.org/3/library/copy.html, and means your example code already works as it should for most orix objects!

If the constructor instead should deepcopy (which makes more sense, I think), then it is not too much work to add e.g. .copy() to the numpy arrays. We could also get more control over the copy behaviour by overriding the dunder methods for copy and deepcopy, instead of changing any existing code.

viljarjf:
Something like:
```python

class A:
    def __init__(self, arg_1, arg_2, kwarg_1 = 0):
        if isinstance(arg_1, A):
            A.__init__(
                self, 
                arg_1.arg_1.copy(), 
                arg_1.arg_2.copy(),
                arg_1.kwarg_1.copy(),
            )
            return
        self.arg_1 = arg_1
        self.arg_2 = arg_2
        ...
```
feels like a decent design too, and gives control over copying by how we pass the parameters to the recursive constructor call.

PR #544, opened 2025-01-27
Add function to expand asymmetric unit cell of Phase

viljarjf:
Adds `Phase.expand_asymmetric_unit()`, which adds all symmetrically equivalent Atom entries into Phase.structure.
This already exists in diffsims (`ReciprocalLatticeVector.sanitize_phase()`, with a #TODO to upsteam into orix. However, I don't think the implementation in diffsims is entirely correct, seeing as the discrepancy between diffsims' and diffpy's axis alignment convention does not seem to be accounted for.

This expansion seems to be performed automatically when reading from a cif which includes symmetry information. I added a test for this, which includes the cif-file test-cases as raw strings. If there is another preferred way to do this please let me know. I saw the `cif_file`-fixture, but I felt making these single-use files into fixtures was unnecessary.

Additionally, I found `Structure.placeInLattice()`, which does the same as #469 for both coordinates and ADPs.

#### Progress of the PR

```python
>>> phase = Phase(
      structure=Structure(
          atoms = [Atom("Si", xyz=(0, 0, 1))],
          lattice=Lattice(4.04, 4.04, 4.04, 90, 90, 90)
      ),
      space_group=227,
  )
  >>> phase.structure
  [Si   0.000000 0.000000 1.000000 1.0000]
  >>> phase.expand_asymmetric_unit()
  >>> phase.structure
  [Si   0.000000 0.000000 0.000000 1.0000,
   Si   0.000000 0.500000 0.500000 1.0000,
   Si   0.500000 0.500000 0.000000 1.0000,
   Si   0.500000 0.000000 0.500000 1.0000,
   Si   0.750000 0.250000 0.750000 1.0000,
   Si   0.250000 0.250000 0.250000 1.0000,
   Si   0.250000 0.750000 0.750000 1.0000,
   Si   0.750000 0.750000 0.250000 1.0000]
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


CSSFrancis:
@argerlt Any chance you can comment on this?  

Should we be expanding this? Would it be better to have an attribute?  I'm maybe not the best person to comment on this but I think we want to avoid transformations which mess with the symmetry reduced zone.  I think it would be better to have one object.


```python 

@property
def asymetric_unit(self):
    """Return the expanded asymetric unit"""
    pass


@property
def symetric_unit(self):
    """Return the reduced symmetric unit"""
    pass

```

argerlt:
3rd option, I think it would be safer and less confusing to have `Phase.expand_asymmetric_unit` instead return a new object. 

I agree with @CSSFrancis, modifying the space group in place might have some unintended knock-on effects. Additionally though, I believe as-implemented any function outside of `Phase` that expanded the asymmetric unit as part of some calculation would alter the input phase as part of that calculation, which is generally not good practice.  

Adding attributes would work, but I can see that being confusing as well. Some future functions might auto apply `Phase.symmetric_unit` when they shouldn't or vice versa, and staring at a plot of points and a list of xyz coordinates that don't match up could be maddening. 

If instead you generate a new object and return it as the output, users could still do `phase=phase.expand_asymmetric_unit()` , but they could also compare the phases, use both in some future stream, modify one or the other, etc. 

Some other thoughts, take or leave them:
1) It would be nice to add an option in `Phase.__init__` to auto-apply this expansion.
2) I think there is an effort to use `copy` in place of `deepcopy` as part of #535
3) The tests you ran only cover cubic and hexagonal systems. it might be helpful to add something like [monoclinic quarts](https://legacy.materialsproject.org/materials/mp-669458/) to potentially catch future edge cases.

viljarjf:
I changed the implementation to return a new object as suggested. I don't have a use for the reverse operation, so I will leave that for later. I also added the C2/c case you suggested @argerlt.

There is cif-writing support in diffpy, which we could use to independently write cifs for testing. However, this forces P1 symmetry for some reason, regardless of the given symmetry, which leaves having the full cif strings in the tests...

As for having this as an option in the constructor, I am more in favor of doing something like ``p = Phase(...).expand_asymmetric_unit()`` instead of adding options to the constructor.

argerlt:
> As for having this as an option in the constructor, I am more in favor of doing something like `p = Phase(...).expand_asymmetric_unit()` instead of adding options to the constructor.

Seems reasonable. In retrospect, changing the data in Phase at creation goes against my earlier point about silently altering data. You way seems better.

argerlt:
@CSSFrancis, unless you have any objections, I can merge this on Monday.

PR #549, opened 2025-04-11
update random method to gaussian form.

argerlt:

replaces the `while` loop approach currently used in `orix.base.random`  with the more common gaussian sampling used elsewhere. 

functionally identical, but (as shown in #218), is 3 to 5 times faster and removes a `while' loop

#### Progress of the PR

NOTE: While this DOES test every line, I did not write a unit test to prove random variables are uniformly random, as (to my knowledge) such a test does not exist.

I could potentially:
A) calculate the first few spherical harmonics for 10,000 random rotations and vectors, and show the unimodal term dominates
B) calculate the angular distances with `dot_outer` for groups of random objects and show the distributions match (ie, everything is equally far from everything else)

However, I tried both and both are slow (a few seconds each) adding to the already considerable testing time, and I'm not sure they protect against anything. 

```
v = Vector3d.random(10000)
r = Rotation.random(10000)
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ Added
 
 Changed
 -------
+- For speed and space reasons, vectors and quaternions are now randomly
+  generated using a gaussian method as opposed to rejection-based sampling.
 
 Removed
 -------
--- a/orix/_base.py
+++ b/orix/_base.py
@@ -162,7 +162,23 @@ class Object3d:
 
     @classmethod
     def random(cls, shape: Union[int, tuple] = 1) -> Object3d:
-        """Create object with random data.
+        """Create object with uniformly distributed random data.
+
+        For objects that map to the surfaces of n-dimensional unit spheres
+        (unit vectors, quaternions, octonions, etc), randomly sampling values
+        for each axis gives a non-uniform sampling that coalesces near the
+        poles. One proper method is to instead sample an n-dimensional
+        gaussian distribution for each axis, then normalize the result.
+
+        For Vector3d objects, this would produce a uniform sampling of the 2D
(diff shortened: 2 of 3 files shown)


argerlt:
Also, this would close out #218

CSSFrancis:
@argerlt Is this ready for review?  I can take a look at it!

argerlt:
> @argerlt Is this ready for review? I can take a look at it!

yup. Is there  a good additional way to signal it's ready? In the past, I've just only posted completed PR's and tagged @hakonanes , but that feels a bit too demanding of one singular maintainer. If there is a standard way to signal, I can do that instead.

CSSFrancis:
> yup. Is there a good additional way to signal it's ready? In the past, I've just only posted completed PR's and tagged @hakonanes , but that feels a bit too demanding of one singular maintainer. If there is a standard way to signal, I can do that instead.

Hmm there is now :) I added the label "Status:needs-review" which should be a good way to signal that a PR needs review.

CSSFrancis (approved):
I would maybe add the tests.  1-2 seconds isn't terribly slow.  One thing you can do is mark tests as "slow" and add that to the conftest.py file. 

https://github.com/pyxem/pyxem/blob/3d995356388089285e1e5f601e17df635e297f5b/pyxem/tests/conftest.py#L47-L61

```python
@pytest.mark.slow
def test_that_is_slow():
    # long running test
```

Then we can run them with the CI but by default they won't run unless you do 

```pytest tests/ --run-slow```

argerlt:
@CSSFrancis, good to go. Used method B for the test, it went faster than I suspected.

Issue #551, opened 2025-04-18
(low priority) Slow tests


argerlt:
Not a pressing issue, but something someone might enjoy tackling:

My (admittedly slow) laptop takes 462 seconds to run all 2346 unittests for orix. All of these are undoubtedly necessary, but some of them take considerably longer than others. 

Running `pytest --durations=25` gives the following list of the slowest tests.  The top 10 are about 40% of the total testing time, though I don't know if it's plausible to actually make them shorter.

```
=============== slowest 50 durations =============== 
58.93s call     orix/tests/quaternion/test_orientation.py::TestMisorientation::test_get_distance_matrix_progressbar_chunksize
27.18s setup    orix/tests/sampling/test_sampling.py::TestUniformSO3::test_uniform_SO3_sample_regions
22.76s call     orix/tests/quaternion/test_quaternion.py::TestQuaternion::test_outer[shape1]
16.77s call     orix/tests/quaternion/test_orientation.py::TestOrientation::test_get_distance_matrix_lazy_parameters
15.97s call     orix/tests/plot/test_direction_color_keys.py::TestDirectionColorKeyTSL::test_rgba_grid_alpha[symmetry0]
15.14s call     orix/tests/plot/test_direction_color_keys.py::TestDirectionColorKeyTSL::test_rgba_grid_limits[symmetry0-expected_xlim0-expected_ylim0]
15.09s call     orix/tests/plot/test_direction_color_keys.py::TestDirectionColorKeyTSL::test_rgba_grid_shape[symmetry0-expected_shape0]
9.25s call     orix/tests/test_vector3d.py::TestVector3dInversePoleDensityFunction::test_ipdf_plot
8.84s call     orix/tests/plot/test_inverse_pole_figure_plot.py::TestInversePoleFigurePlot::test_inverse_pole_density_function
7.88s call     orix/tests/plot/test_direction_color_keys.py::TestDirectionColorKeyTSL::test_rgba_grid_limits[symmetry1-expected_xlim1-expected_ylim1]
7.87s call     orix/tests/plot/test_direction_color_keys.py::TestDirectionColorKeyTSL::test_rgba_grid_shape[symmetry1-expected_shape1]
6.42s call     orix/tests/plot/test_inverse_pole_figure_plot.py::TestInversePoleFigurePlot::test_plot_ipf_color_key_show_title[sym0-left]
5.05s call     orix/tests/quaternion/test_quaternion.py::TestQuaternion::test_outer_lazy_chunk_size
[...]

PR #553, opened 2025-04-22
Address 542 duplicate phase naming and #551 slow unit tests

argerlt:
When loading a .ang file, ORIX will now either:

1) Do nothing if there is no formula data (this is the old behavior)
2) Rename Phases according to the `formula` value if all formula's are unique (ie "Fe", "Fe3C", .etc.)
3) If there are phases with identical formulas but different point groups, it will rename the phases as "formulat-PointGroup" (ie, `Ti-622` and `Ti-m-3m')

OR 

4) If you pass `'io.loadang('filename' ,autogen_names=False),  it overwrites nothing and uses whatever is already in the file.

@gsparks3, I think this PR addresses your problem of identically names phases. If you could try it out on a few files, that would be appreciated. it works correctly on [this data you posted in issue 543](https://github.com/user-attachments/files/19817756/TwoPhasesSameFormula_OIM8.6.ang.zip) as well as all the default ORIX datasets and unittests, but there might be more edge cases to consider.

@CSSFrancis  As an additional item to also close out #551, I reduced the run time of the 3 largest tests. Past that, it seemed to be a diminishing return and questionable equivalency. 

More importantly though, I updated the documentation to encourage developers to use pytest-xdist for parallel unit testing.  This is what the automated unit-tests have been using since 2022, so it seems safe to add to the development options. these two changes together took my unit testing time from 8 minutes to 40 seconds. 

#### Progress of the PR

(assuming you have unzipped the .ang file discussed above)
```python
>>> from orix import io
>>> ebsd = io.load("TwoPhasesSameFormula_OIM8.6.ang")
>>> ebsd

Out[133]: 
Phase  Orientations    Name  Space group  Point group  Proper point group       Color
    1    63 (78.8%)  Ti-622         None          622                 622  tab:orange
    2    17 (21.2%)  Ti-432         None          432                 432    tab:blue
Properties: unknown1, unknown2, unknown3, unknown4, unknown5, unknown6, unknown7
Scan unit: um
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


orix/io/plugins/ang.py
```diff
@@ -355,10 +387,12 @@ def _get_phases_from_header(header: List[str]) -> dict:
...
 
-    # Use formulas in place of material names if they are all valid
+    # Pop out the formula information, which can optionally be used later to
+    # rename the phses
     formulas = phases.pop("formulas")
-    if len(formulas) == n_phases and all([len(name) for name in formulas]):
-        phases["names"] = formulas
+
+    # if len(formulas) == n_phases and all([len(name) for name in formulas]):
+    #     phases["names"] = formulas
```
CSSFrancis, line 396: This is left over and can be deleted.

CSSFrancis (approved):
@argerlt This looks great.  There is one bit of left over code that might be good to remove.

argerlt:
@CSSFrancis , Added your suggestion and Co-pilot's.   

You could merge this as is, but probably makes sense to do #554 first, then I can change the wording in this test to appropriately match.

My fault for trying to tackle two tasks in one PR.

PR #556, opened 2025-05-15
Add copy constructors

viljarjf:
Partially addresses #535, input very much welcome!
Adds a copy constructor to some classes, i.e. if the first parameter to the constructor is an existing instance, make a copy of it.
This pattern exists in e.g. diffpy.structure already.

As described in the linked issue, we can also then add a copy method like this:
```python
def copy(self) -> Self:
    return self.__class__(self)
```
The control of how deep the copy becomes lies then in the constructor. I left this out for the most part, but e.g. arrays should probably be copied in the constructor call:
```python
class Phase:
    def __init__(self, arg1, ...):
       if isinstance(arg1, Phase):
            return Phase.__init__(
                self,
                arg1.arg1.copy(), # Explicitly copy here, if this behaviour is desired
                ...
            )
```

Currently implemented for:
- Phase
- PhaseList
- CrystalMap
- Quaternion, Rotation, Misorientation, Orientation (already worked by inheriting from Object3d, I just added a test. It might already be tested and I missed it)

#### Progress of the PR

```python
>>> p1 = Phase(...)
>>> p2 = Phase(p1) # p2 is a new copy of p1
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Added
   and Misorientations, albiet with a loss of symmetry information.
 - ``ignore_zero`` option for ``Vector3d.unique`` to allow all-zero elements, which were
   previously discarded. Discarding is still enabled by default.
+- Added copy constructors to ``Phase``, ``PhaseList`` and ``CrystalMap``.
 
 Changed
 -------
--- a/orix/crystal_map/crystal_map.py
+++ b/orix/crystal_map/crystal_map.py
@@ -204,7 +204,7 @@ class CrystalMap:
 
     def __init__(
         self,
-        rotations: Rotation,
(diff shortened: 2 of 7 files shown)


argerlt:
I edited the tests to match the formatting, but everything else looks good.  Also, don't forget to update `CHANGELOG.rst` to reflect the new features.

Since this is related to @hakonanes 's Issue, I suggest we wait to see if he wants to weigh in at all, otherwise once the changelog is updated, I can push this at the start of next week.

argerlt (changes requested):
Changelog.rst needs to be updated with something like `copy constructor added to CrystalMap and Phaselist`

argerlt (approved):

PR #558, opened 2025-05-19
adding ancillary capabilities from PR442

argerlt:
Implements some ancillary capabilities from #442 that were never merged into the develop branch. 
I added @hakonanes as a co-author since I essentially copied his homework on this one. Contains two major changes:
 
1) Renames `map_into_symmetry_reduced_zone()` to `reduce()`,  the value of which @hakonanes  summed up nicely previously:
> The other major change is naming: map_into_symmetry_reduced_zone() -> reduce(), inspired by the attractive naming of [scipy.spatial.transforms.Rotation.reduce()](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.reduce.html). ... 
> ... Using this word allows us also to add reduce as an intuitive parameter to various methods if we don't want to apply symmetry operations, instead of having a use_symmetry parameter (similar to MTEX' noSymmetry). We could then rename `Miller.in_fundamental_region()` to `Miller.reduce()` etc.

2) adds `Rotation.from_path_ends`, which is roughly equivalent to `Vector3d.from_path_ends`.

Still requires unit tests and examples.

# Note, after this is merged, I think the ONLY thing missing from #442 is the proper handling of misorientation symmetry.

#### Progress of the PR

WIP

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


argerlt:
@hakonanes , @CSSFrancis , @viljarjf, anyone got advice on how to unit test `Orientation.from_path_ends`? an easy out would be to just check that the size of the produced paths are the correct length, and the spacing is consistent between points, but that doesn't really test if the paths are in the correct spot. Toy example below to show off the feature in question:

```
from orix.quaternion.symmetry import O
import numpy as np

rot_path = Orientation(
    data=np.array(
        [
            [1, 0, 0, 0],
            [1, 1, 0, 0],
            [1, 0, 1, 0],
            [1, 0, 0, 1],
            [1, 0, -1, 0],
            [1, 0, 0, -1],
            [1, 0, 0, -1],
        ]
    )
)
m3m_path = Orientation(
    data=np.array(
        [
            [1, 0, 0, 0],
            [2, 1, 0, 0],
            [3, 0, 1, 0],
            [4, 0, 0, 1],
            [5, 0, -1, 0],
            [6, 0, 0, -1],
            [7, 0, 0, -1],
            [8, 1, 0, 0],
            [9, 0, 1, 0],
            [10, 0, 0, 1],
            [11, 0, -1, 0],
            [12, 0, 0, -1],
            [13, 0, 0, -1],
        ]
    ),
    symmetry=O,
)

fig, ax = plt.subplots(1, 2)
# plot in homochoric space, no symmetry
Orientation.from_path_ends(rot_path.reduce(), True).reduce().scatter(
    figure=fig, position = [1,2,1], marker=">", c=np.arange(700)
)

# plot in rodrigues space, cubic
Orientation.from_path_ends(m3m_path.reduce(), True).reduce().scatter(
    figure=fig, position = [1,2,2], marker=">", c=np.arange(1300)
)

```
image

orix/quaternion/quaternion.py
```diff
@@ -680,6 +690,58 @@ def from_align_vectors(
...
+            ax, ang = _conversions.qu2ax((~q1 * q2).data)
+            # get 'steps=n' steps along the trip and add them to the journey
+            trip = Quaternion.from_axes_angles(ax, np.linspace(0, ang, steps))
+            path_list.append((q1 * (trip.flatten())).data)
+        path_data = np.concatenate(path_list, axis=0)
+        path = waypoints.__class__(path_data)
+        # copy the symmetry if it exists
+        if hasattr(waypoints, "_symmetry"):
+            path._symmetry = waypoints._symmetry
+        return path
```
hakonanes, line 743: Quaternions do not know anything about symmetries. This method must be overwritten in the orientation and misorientation classes instead.

hakonanes (changes requested):
Good stuff. However, many things need changing!

orix/quaternion/quaternion.py
hakonanes: What I mean is that we should overwrite in (mis)orientation classes to enforce separation of concerns. Quaternions shouldn't know anything about symmetries.

orix/quaternion/quaternion.py
```diff
@@ -680,6 +690,58 @@ def from_align_vectors(
 
         return out[0] if len(out) == 1 else tuple(out)
 
+    @classmethod
+    def from_path_ends(
+        cls, waypoints: Quaternion, close_loop: bool = False, steps: int = 100
```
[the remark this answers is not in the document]
argerlt: Changing to 'points`, but if you can think of a good third option, let me know. 

The rotations this function returns are, in my mind, points tracing out a path in some non-euclidean space, so calling the ends of these paths `points` as well felt confusing. `waypoints` to me sound like sign posts or rock piles along hiking trails, where you walk from one to the next, but that's not super clear to anyone but me.

edges? corners? stops? vertices? none of those feel right either.....

orix/quaternion/quaternion.py
```diff
@@ -680,6 +690,58 @@ def from_align_vectors(
...
+    ) -> Quaternion:
+        """Return Quaternions tracing the shortest path between two or more
+        consecutive waypoints.
+
+        Parameters
+        ----------
+        waypoints : Quaternion
```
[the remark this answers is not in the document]
hakonanes: Type hints from return sections must also be removed.

argerlt:
> I don't have a good suggestion for testing orientation paths.
> 
> However, here's another test which may be useful, just checking that a path from identity to some other one produces the same as a fiber defined by one axes and increasing angles:
> 
> ```python
>     def test_from_path_ends_fiber(self):
>         Q1 = Quaternion.identity()
>         Q2 = Quaternion.from_axes_angles([1, 1, 1], 60, degrees=True)
>         Q12 = Quaternion.stack((Q1, Q2))
>         Q_path1 = Quaternion.from_axes_angles([1, 1, 1], np.arange(59), degrees=True)
>         Q_path2 = Quaternion.from_path_ends(Q12, steps=Q_path1.size)
>         assert np.allclose(Q_path1.dot(Q_path2), 1, atol=1e-3)
> ```

This seems good enough. Also, as stated in the docstring, paths are not symmetry-aware, so this plus the expected size checks, plus the proper symmetry copying seems sufficient.

hakonanes:
I suggest to not mark a PR as a draft after there's been a review.

argerlt:
> I suggest to not mark a PR as a draft after there's been a review.

I was trying to signify I had started fixing this PR, but it wasn't yet to the point where it was worth maintainers taking time to review it.  If there is a better way to signify as much, I can switch. 

I have to swap between computers throughout the week based on other work happening, so pushing partial patches is convenient, and also a way to get unofficial input from others..  @CSSFrancis and I were using labels for a bit, but I was trying out the Draft designation instead since only maintainers can add labels, and I thought draft's might be more inclusive.

orix/quaternion/misorientation.py
```diff
@@ -263,6 +264,37 @@ def from_scipy_rotation(
...
+        Returns
+        -------
+        path
+            quaternions that map a path between the given waypoints.
+        """
+        out = super().from_path_ends(points=points, closed=closed, steps=steps)
+        path = cls(out.data)
+        # copy the symmetry if it exists.
+        if hasattr(points, "_symmetry"):
+            path._symmetry = points._symmetry
```
hakonanes, line 295: What I mean above (https://github.com/pyxem/orix/pull/558/files#r2215490352) is that we should drop the check and just do this:

```suggestion
        path._symmetry = points._symmetry
```

If not (mis)orientations are given, this will raise an attribute error. Users should see this error to learn that they should instead use the equivalent rotation or quaternion method, instead of the call passing silently.

If we want to be more kind to the user, and perhaps save unnecessary computation, we can make a check at the top for type:

```python
if not isinstance(points, Misorientation):
    raise TypeError(f"Points must be a (mis)orientation, not of type {type(points)}")
```

After this is 100% safe to access the symmetry.

argerlt:
@hakonanes , give me 24 hours on this one please. I believe all the code is good to go, but the example I wrote is half-baked and needs some work.

PR #586, opened 2025-09-08
New function to register custom Matplotlib projections

### LINKED ISSUE #585: Register Matplotlib projections calling an explicit function
I suggest we add a function `orix.plot.register_projections()` that registers our custom projections (stereographic, axis-angle, etc.) with Matplotlib. This replaces the side-effect of importing `orix.plot`, which called a file which registered the projections.

As pointed out by @argerlt in https://github.com/pyxem/orix/pull/583, this side-effect became a non-solution when lazy loading was introduced. We shouldn't rely on it, since it's a side-effect.

hakonanes:
I've adde a new function to register all our Matplotlib projections. This replaces registration of the projections via the side-effect of importing the `orix.plot` module.

This is required, as @argerlt points out in #583, because lazy loading of the plot module means importing the module does not cause the projections to be registered. This PR replaces #583.

This fixes #585.

#### Progress of the PR

Previously:
```python
>>> from orix import plot  # Register projections
>>> fig = plt.figure()
>>> ax = fig.add_subplot(projection="stereographic")
```

Now:
```python
>>> import orix.plot as opl
>>> opl.register_projections()
>>> fig = plt.figure()
>>> ax = fig.add_subplot(projection="stereographic")
```

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,11 @@ Added
   :class:`~orix.vector3d.Miller`.
 - Explicit support for Python 3.13.
 - Dependency on `lazy-loader`.
+- Function :func:`~orix.plot.register_projections` to register all our custom
+  projections for use in Matplotlib.
+  An example of a custom projection is the :class:`~orix.plot.StereographicPlot`.
+  This function replaces the previous behavior of relying on a side-effect of importing
+  the :mod:`orix.plot` module, which also registered the projections.
 
 Changed
 -------
--- a/orix/plot/_plot.py
+++ b/orix/plot/_plot.py
@@ -20,9 +20,15 @@
 import matplotlib.projections as mprojections
 
 from .crystal_map_plot import CrystalMapPlot
-from .inverse_pole_figure_plot import InversePoleFigurePlot
 from .rotation_plot import AxAnglePlot, RodriguesPlot
 from .stereographic_plot import StereographicPlot
+
+# Inverse pole figure plot class must be imported below stereographic
+# plot class, since the former imports the latter
+# isort: off
+from .inverse_pole_figure_plot import InversePoleFigurePlot
+
+# isort: on
 
 
 def register_projections() -> None:
--- a/orix/plot/_plot.py
+++ b/orix/plot/_plot.py
@@ -1,0 +1,65 @@
+#
+# Copyright 2018-2025 the orix developers
+#
+# This file is part of orix.
+#
+# orix is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# orix is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with orix. If not, see <http://www.gnu.org/licenses/>.
+#
+
+import matplotlib.projections as mprojections
+
+from .crystal_map_plot import CrystalMapPlot
+from .inverse_pole_figure_plot import InversePoleFigurePlot
+from .rotation_plot import AxAnglePlot, RodriguesPlot
+from .stereographic_plot import StereographicPlot
+
+
+def register_projections() -> None:
+    """Register custom Matplotlib projections.
+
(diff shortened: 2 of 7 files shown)


orix/plot/_plot.py
```diff
@@ -0,0 +1,65 @@
...
+from .inverse_pole_figure_plot import InversePoleFigurePlot
+from .rotation_plot import AxAnglePlot, RodriguesPlot
+from .stereographic_plot import StereographicPlot
+
+
+def register_projections() -> None:
+    """Register custom Matplotlib projections.
+
+    This must be called when using one of our custom projections in
+    Matplotlib.
```
argerlt, line 32:
```suggestion
    """Register custom Matplotlib projections.

    As an alternative to explicitly loading each class, this
    function registers all the custom ORIX projection classes 
    in the current `matplotlib.projections.projection_registry`
    instance, allowing them to be lazily loaded and used as
    needed via the `projections` variable, similar to the default
    `polar` and `3d` default projections.
```

orix/plot/_plot.py
```diff
@@ -0,0 +1,65 @@
...
+    >>> fig = plt.figure()
+
+    May fail
+
+    >>> ax = fig.add_subplot(projection="stereographic")  # doctest: +SKIP
+
+    Works
+
+    >>> register_projections()
+    >>> ax = fig.add_subplot(projection="stereographic")
```
argerlt, line 55:
```suggestion
    >>> import matplotlib.pyplot as plt
    >>> from orix.plot import register_projections
    >>> fig = plt.figure()

    The following will fail, unless class:`~orix.plot.StereographicPlot`
    was previously imported.

    >>> ax = fig.add_subplot(projection="stereographic")  # doctest: +SKIP

    Calling this function explicitly registers every custom projection

    >>> register_projections()
    >>> ax1 = fig.add_subplot(2, 2, 1, projection="stereographic")
    >>> ax2 = fig.add_subplot(2, 2, 2, projection="rodrigues")
    >>> ax3 = fig.add_subplot(2, 2, 3, projection="plot_map")
    >>> ax4 = fig.add_subplot(2, 2, 4, projection="ipf")
```

argerlt (changes requested):
This works for me. Just to make things clean, I'd suggest you add the following block of code to `plot.__init__.py` as opposed to me changing #583, then we just push this PR and close out both #583 and #585

```
"""Extensions of Matplotlib's projections framework for plotting
:class:`~orix.vector.Vector3d`, :class:`~orix.quaternion.Rotation`,
:class:`~orix.quaternion.Orientation`,
:class:`~orix.quaternion.Misorientation`, and
:class:`~orix.crystal_map.CrystalMap`.
"""

import lazy_loader
from orix.plot import _plot

# Imports from stub file (see contributor guide for details)
__getattr__, __dir__, __all__ = lazy_loader.attach_stub(__name__, __file__)

del lazy_loader

# Register orix's custom matplotlib.projection classes
_plot.register_projections()
```

CHANGELOG.rst
```diff
@@ -27,6 +27,11 @@ Added
   :class:`~orix.vector3d.Miller`.
 - Explicit support for Python 3.13.
 - Dependency on `lazy-loader`.
+- Function :func:`~orix.plot.register_projections` to register all our custom
+  projections for use in Matplotlib.
+  An example of a custom projection is the :class:`~orix.plot.StereographicPlot`.
+  This function replaces the previous behavior of relying on a side-effect of importing
+  the :mod:`orix.plot` module, which also registered the projections.
```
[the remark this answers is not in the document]
hakonanes: But, you don't consider registering projections by importing `orix.plot` as a side-effect? I do, and therefore I think this should be kept.

What I'd like to happen is:
1. New users use the `register_projections()` function
2. Existing code relying on this side-effect still works, but we try to encourage (and actively search for, I volunteer for that) using the new function
3. We remove the side-effect, after we're confident that it won't cause big problems

Do you disagree with this plan? If so, what do your forsee as the long-term solution?

orix/plot/_plot.py
hakonanes: I don't think importing `orix.plot.StereographicPlot` but not using it is a behavior we should encourage. This is still a side-effect. Calling the registration function, on the other hand, is explicit and easy to understand.

I suggest to leave the docstring as-is.

argerlt (approved):

hakonanes:
Thank you for the review and merge, @argerlt.

And thank you for referencing that Matplotlib change, I wasn't aware that they had an "apparently unused" import as well!

PR #588, opened 2025-09-19
Perform histogramming for density-plots in equi-angle cubed-sphere co…

### LINKED ISSUE #441: pole_density_function() plots do not appear correct in the center
The center of plots produced by pole_density_function() do not appear to be correct. For example https://orix.readthedocs.io/en/stable/examples/plotting/subplots.html#sphx-glr-examples-plotting-subplots-py. The center of the plot appear distorted and I have similar issues with my pole density plots also.

MACarlsen:
Resolves #441 by performing the histogramming in six separate charts, corresponding to six cube faces, in gnomonic equal-angle coordinates. This avoids the coordinate divergence at the pole, which is unavoidable in any single-chart map of the sphere.

Smoothing is handled by diffusion on the cube surfaces and across edges.

I might interpolate the smoothed histograms onto an longitude-latitude map to avoid down-stream changes. Remains to be seen.

### DIFF
--- a/orix/measure/pole_density_function.py
+++ b/orix/measure/pole_density_function.py
@@ -206,10 +206,6 @@ def pole_density_function(
     ).unit
     v_grid_vertexes = Vector3d.from_polar(azimuth=azimuth_grid, polar=polar_grid).unit
 
-    mask = ~(v_grid <= symmetry.fundamental_sector)
-    v_grid = v_grid.in_fundamental_sector(symmetry)
-    v_grid_vertexes = v_grid_vertexes.in_fundamental_sector(symmetry)
-
     # Interpolation from histograms to plot grid
     grid_face_index, grid_face_coords = _cube_gnom_coordinates(v_grid)
     hist_grid = np.zeros(v_grid.shape)
@@ -228,6 +224,7 @@ def pole_density_function(
     hist = hist_grid
 
     # Mask out points outside funamental region.
+    mask = ~(v_grid <= symmetry.fundamental_sector)
     hist = np.ma.array(hist, mask=mask)
 
     # Transform grdi to mystery coordinates used by plotting routine


MACarlsen:
I'm taking the liberty of computing the histograms at lower-resoltion, if the blurring kernel is chosen much larger than the resolution, and then I interpolate up in the end for plottting. Without that trick performance suffers for small `resoltution` and high `sigma`. (Scales with resolution to the fourth power. Ouch!)

Runtime seems to be dominated by `pcolormesh` now.

image

Just need to write/fix tests.

MACarlsen:
I think this is as far as I can get. I did break some things in order to fix the problem:

The histograms are computed on six separate grids, so to avoid having to touch the plotting code too much in the PR, I am interpolating back to a polar-grid and passing that to the plotting functions.

There are some assumptions in the existing tests, that rely on the plot-grid to be equal area, which it is not in the current version. 

Also, it is not clear from the documentation what the expected behaviour should be in the case ``mrd=False``. My version doesn't return anything sensible with 'mrd=False' at the moment. From the function docstring, I would expect it to be counts per. solid angle while the existing code gives counts per bin.

But since the bins in the equal-area polar grid are not useful, I don't see the value of this. If there is some need to know the counts per arbitrary bin, I would suggest sepparating that functionality from plotting.

But of course this is not my package.
    - Mads

MACarlsen:
1. Feel free! I'm too much of a github beginner to really understand what that means. I'm only touching 3 files, so hopefully it's not so painfull.

2. I just copy-pasted from the readthedocs with minor modifications:

```
import time

from orix import data, plot

### Side quest to get plotting to work ###
from orix.plot import StereographicPlot, InversePoleFigurePlot
import matplotlib.projections as mprojections
mprojections.register_projection(StereographicPlot)
mprojections.register_projection(InversePoleFigurePlot)
### Side quest to get plotting to work ###

xmap = data.sdss_ferrite_austenite(allow_download=True)
print(xmap)

# Extract orientations, O
pg_m3m = xmap.phases[1].point_group.laue
O_fe = xmap["ferrite"].orientations
O_au = xmap["austenite"].orientations

# Some sample direction, v
v = Vector3d([0, 0, 1])
v_title = "Z"

# Rotate sample direction v into every crystal orientation O
t_fe = O_fe * v
t_au = O_au * v

# Set IPDF range
vmin, vmax = (0, 3)

subplot_kw = {"projection": "ipf", "symmetry": pg_m3m, "direction": v}
fig = plt.figure(figsize=(9, 8))

ax0 = fig.add_subplot(221, **subplot_kw)
ax0.scatter(O_fe, alpha=0.05)
_ = ax0.set_title(f"Ferrite, {v_title}")

ax1 = fig.add_subplot(222, **subplot_kw)
ax1.scatter(O_au, alpha=0.05)
_ = ax1.set_title(f"Austenite, {v_title}")

t1 = time.time()
ax2 = fig.add_subplot(223, **subplot_kw)
ax2.pole_density_function(t_fe, vmin=vmin, vmax=vmax)
_ = ax2.set_title(f"Ferrite, {v_title}")
print(time.time()-t1)

t1 = time.time()
ax3 = fig.add_subplot(224, **subplot_kw)
ax3.pole_density_function(t_au, vmin=vmin, vmax=vmax)
_ = ax3.set_title(f"Austenite, {v_title}")
print(time.time()-t1)
plt.show()
```

argerlt:
Ope, apologies. The way permissions are set up, you need to do the merge or rebase yourself.

You can do it in most git GUI's or via command line, but there should also be an option on your github page right below the green download option.
https://github.com/MACarlsen/orix/tree/histograms-in-gnom-cube-coords/develop

argerlt:
hmmm...... Something about this isn't right. if you try to redo this with a different non-cubic symmetry, the results look weird:



```import matplotlib.pyplot as plt

from orix import data, plot
from orix.quaternion.symmetry import C3

plt.close("all")

xmap = data.sdss_ferrite_austenite(allow_download=True)
pg_m3m = xmap.phases[1].point_group.laue
O_au = xmap["austenite"].orientations
O_fe = xmap["ferrite"].orientations

O_au = (pg_m3m.outer(O_au)).flatten()[::13]
O_fe = (pg_m3m.outer(O_au)).flatten()[::13]

O_au.symmetry = C3
O_fe.symmetry = C3

# Orientation colors
ckey_m3m = plot.IPFColorKeyTSL(pg_m3m)
rgb_au = ckey_m3m.orientation2color(O_au)
rgb_fe = ckey_m3m.orientation2color(O_fe)

# Austenite <111> poles in the sample reference frame
t_au = Miller(uvw=[1, 1, 1], phase=xmap.phases["austenite"])
t_au_all = O_au.inv().outer(t_au)

# Ferrite <111> poles in the sample reference frame
t_fe = Miller(uvw=[1, 1, 1], phase=xmap.phases["ferrite"])
t_fe_all = O_fe.inv().outer(t_fe)


fig, ax = plt.subplots(2, 2, subplot_kw={"projection": "ipf", "symmetry": C3})
ax[0, 0].scatter(t_au_all, s=2, c=rgb_au.T)
ax[0, 1].scatter(t_fe_all, s=2, c=rgb_fe.T)
ax[1, 0].pole_density_function(t_au_all)
ax[1, 1].pole_density_function(t_fe_all)


fig.subplots_adjust(hspace=0, wspace=0.1)
plt.tight_layout()
```
image

MACarlsen:
@argerlt
I see what was going on now.

I thought ``in_fundamental_sector()`` projects vectors into the fundamental zone, (int the sense of finding the closest point inside) but it really finds the symmetry equivalent vector inside the zone.

orix/sampling/S2_sampling.py
```diff
@@ -392,6 +392,98 @@ def sample_S2_cube_mesh(
     return Vector3d(np.vstack((bottom, top, east, west, south, north, m_c))).unit
 
 
+def sample_S2_equiangle_cube_mesh_vertexes(
+    resolution: float, grid_type: str = "spherified_corner"
```
argerlt, line 396: I believe `sample_uv_mesh` already provides this capability.

orix/measure/pole_density_function.py
```diff
@@ -102,123 +106,332 @@ def pole_density_function(
...
+
+    # Explicit symmetrization
+    for rotation in symmetry:
+
+        face_index_array, face_coordinates = _cube_gnom_coordinates(rotation * v)
+        face_index_array = face_index_array.ravel()
+        face_coordinate_1 = face_coordinates[0].ravel()
+        face_coordinate_2 = face_coordinates[1].ravel()
+        for face_index in range(6):
+            this_face = face_index_array == face_index
```
argerlt, line 144: Naming like `this face` and `that face` is generally best to avoid, as it should to make sense to anyone looking at this code a year from now.

orix/measure/pole_density_function.py
```diff
@@ -102,123 +106,332 @@ def pole_density_function(
...
+    x, y = x.reshape(v_grid_vertexes.shape), y.reshape(v_grid_vertexes.shape)
 
     if log:
         # +1 to avoid taking the log of 0
         hist = np.log(hist + 1)
 
     return hist, (x, y)
+
+
+def _cube_gnom_coordinates(
```
argerlt, line 241: We should try to replace this function with a considerably more concise one. Additionally, if you can make the face assignments part of a procedural `for` loop, it's generally much more desirable as it's less prone to creating an error later on, and is also much easier for future editors to comprehend.

orix/sampling/S2_sampling.py
```diff
@@ -392,6 +392,98 @@ def sample_S2_cube_mesh(
...
+    data[1, ...] = np.stack([-dim_3, dim_2, dim_1], axis=-1)
+    data[2, ...] = np.stack([dim_2, dim_3, dim_1], axis=-1)
+    data[3, ...] = np.stack([dim_2, -dim_3, dim_1], axis=-1)
+    data[4, ...] = np.stack([dim_2, dim_1, dim_3], axis=-1)
+    data[5, ...] = np.stack([dim_2, dim_1, -dim_3], axis=-1)
+
+    return Vector3d(data=data)
+
+
+def sample_S2_equiangle_cube_mesh_face_centers(
```
argerlt, line 441: As you point out I believe this capability already exists `sample_S2_cube_mesh`. What you are doing here is the equivalent of setting `grid_type="normalized"`
It looks like specifically what you want here is to avoid getting edges and corners, which would be a good optional input to add to the `sample_S2_cube_mesh`. Within the function, you could add an if/then to cut the first and last entry off of the `grid_on_edge` variable.

As a general rule, for a big project like ORIX, it's generally better to reuse as much code as possible to avoid errors down the road. I nothing else, it helps ensure identical calculations are performed identically in different functions throughout ORIX.

For example, some day someone might take your PDF function, compare the calculated centers to vectors calculated with the same resolution via `sample_S2_cube_mesh`, and not understanding why the values are different or in a different order.

argerlt (changes requested):
This functionality is a great upgrade, the code itself just needs some work to make it more readable/concise/reusable. I left some suggestions, feel free to comment if you disagree.

Issue #613, opened 2025-12-12
Recent releases of `orix` break `pyxem` and `kikuchipy`


ericpre:
Most likely orix >=0.14.0

https://github.com/hyperspy/hyperspy-extensions-list/actions/runs/20154109104/job/57852785628

```python
============================= test session starts ==============================
platform linux -- Python 3.12.12, pytest-8.4.2, pluggy-1.6.0
Matplotlib: 3.10.8
Freetype: 2.6.1
rootdir: /home/runner/work/_temp
plugins: rerunfailures-16.1, mpl-0.18.0, xdist-3.8.0, instafail-0.5.0
collected 823 items / 1 error / 5 deselected / 818 selected

==================================== ERRORS ====================================
_______ ERROR collecting tests/test_signals/test_ebsd_master_pattern.py ________
ImportError while importing test module '/home/runner/miniconda3/envs/test/lib/python3.12/site-packages/kikuchipy/tests/test_signals/test_ebsd_master_pattern.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../miniconda3/envs/test/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/test/lib/python3.12/site-packages/kikuchipy/tests/test_signals/test_ebsd_master_pattern.py:779: in <module>
    class TestFitPatternDetectorOrientation:
../../miniconda3/envs/test/lib/python3.12/site-packages/kikuchipy/tests/test_signals/test_ebsd_master_pattern.py:798: in TestFitPatternDetectorOrientation
    simulator = kp.simulations.KikuchiPatternSimulator(ref)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/test/lib/python3.12/site-packages/lazy_loader/__init__.py:82: in __getattr__
    submod = importlib.import_module(submod_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/test/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #614, opened 2025-12-15
v0.14.2 patch release, removing RTD PDF build

### LINKED ISSUE #611: ReadtheDoc PDF errors
In the process of patching 0.14.1, I spent some time trying to understand why ReadtheDocs is failing now. It seems by default the pdf compiler allows 100 latex errors before aborting, and we just recently hit that number.

In this case, the 100th error is due to [crystal_reference_frame.py](https://github.com/pyxem/orix/blob/c19b15fc8e607d096c09fc929f0e8819a2c2df89/examples/crystal_phase/crystal_reference_frame.py#L85), which includes this text:
```
# Given the chosen alignment of basis vectors with the Cartesian reference frame,
# :math:`\mathbf{A}` is defined as
#
# .. math::
#   \begin{equation}
#   \mathbf{A} =
#   \begin{pmatrix}
#   a & 0 & 0\\
#   b\cos\gamma & b\sin\gamma & 0\\
#   c\cos\beta & -c\frac{(\cos\beta\cos\gamma - \cos\alpha)}{\sin\gamma} & \frac{\mathrm{V}}{ab\sin\gamma}
#   \end{pmatrix},
#   \end{equation}
#
# where :math:`V` is the volume of the unit cell.
```

which Sphinx then converts to the following snippet in `orix.tex`:

```
Given the chosen alignment of basis vectors with the Cartesian reference frame,
\(\mathbf{A}\) is defined as
\begin{equation*}
\begin{split}\begin{equation}
\mathbf{A} =
\begin{pmatrix}
a & 0 & 0\\
b\cos\gamma & b\sin\gamma & 0\\
c\cos\beta & -c\frac{(\cos\beta\cos\gamma - \cos\alpha)}{\sin\gamma} & \frac{\mathrm{V}}{ab\sin\gamma}
\end{pmatrix},
\end{equation}\end{split}
\end{equation*}
\sphinxAtStartPar
where \(V\) is the volume of the unit cell.
```

Which throws the following error when built with latexmk

```
l.10411 \end{equation}\end{split}
                                 
! LaTeX Error: \begin{split} on input line 10411 ended by \end{equation}.
See the LaTeX manual or LaTeX Companion for explanation.
```

Which, annoyingly, still renders fine, but counts towards the error threhold.

Image

The suggested solutions seem to be either defining a new latex engine in conf.py, or rewriting  all example files to render more nicely when translated by Sphinx into Latex.

I'm trying a few different engines [here](https://app.readthedocs.org/projects/austin-orix/builds/30649914/), but personally, I think it might be better to just stop building the PDF. ReadTheDocs already generates an offline HTML archive, the PDF is redundant and counts towards our already limited allowed build time.

Thoughts? @CSSFrancis @hakonanes ?

### LINKED ISSUE #616: Obsolete dependency on `pycifrw`
It should be possible to drop the dependence on `pycifrw`:

https://github.com/pyxem/orix/blob/3db9eb7234141c9701cb83799a883d43673a1287/pyproject.toml#L36

`diffpy.structure` is already at `3.3.1`
https://github.com/diffpy/diffpy.structure/releases

argerlt:
I plan on leaving this open for a bit as additional potential downstream pipeline problems are reported, but this includes optional solutions to address #611, #613, and #616

~~1) (optional) make Arrow3D a public facing class~~ (removed, based on discussion in [kikuchipy](https://github.com/pyxem/kikuchipy/pull/765))
2) Remove the pdf build in ReadtheDocs
~~3) remove dependency on pycifrw and version restriction from pytest-rerunfailures, as the bugs necessitating both have been patched.~~

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ All user facing changes to this project are documented in this file. The format 
 on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`__, and this project tries
 its best to adhere to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`__.
 
-2025-12-10 - version 0.14.1
+2025-12-10 - version 0.14.2
 ===========================
 
 Changed
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ All user facing changes to this project are documented in this file. The format 
 on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`__, and this project tries
 its best to adhere to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`__.
 
-2025-12-10 - version 0.14.2
+2026-01-12 - version 0.14.2
 ===========================
 
 Changed
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,17 @@ its best to adhere to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>
 2025-12-10 - version 0.14.1
 ===========================
 
+Changed
+-----
+- The PDF version of the ORIX documentation is no longer automatically built by
+  ReadtheDocs.
+- :class: Arrow3D class is no longer a private class, and is accessible as 
+  ``orix.plot.Arrow3d``.
+
+
+2025-12-10 - version 0.14.1
+===========================
+
 Fixed
 -----
 - Documentation build now passes again.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ All user facing changes to this project are documented in this file. The format 
 on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`__, and this project tries
 its best to adhere to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`__.
 
-2026-01-12 - version 0.14.2
+2026-02-03 - version 0.14.2
 ===========================
 
 Changed
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,8 +13,6 @@ Changed
 -----
 - The PDF version of the ORIX documentation is no longer automatically built by
   ReadtheDocs.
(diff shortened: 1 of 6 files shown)


argerlt:
@hakonanes, once the current kikuchipy patch 0.11.4 is done, assuming nothing new breaks in orix/pyxem/hyperspy/kikuchipy, this patch is probably good to go.

pyproject.toml
```diff
@@ -64,9 +62,7 @@ doc = [
 tests = [
     "numpydoc",
     "pytest                         >= 5.4",
-    # TODO: Remove restriction once a new release fixing
-    # https://github.com/pytest-dev/pytest-rerunfailures/issues/302 is available.
-    "pytest-rerunfailures           < 16.0",
+    "pytest-rerunfailures",
```
hakonanes: We have to say we don't want version 16.0, as described in the [packaging dependency specifier](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#grammar) documentation:

```
    "pytest-rerunfailures           != 16.0",
```

hakonanes (changes requested):
I'm OK with leaving out the PDF. Thanks for making the change!

CHANGELOG.rst
```diff
@@ -6,6 +6,14 @@ All user facing changes to this project are documented in this file. The format
 on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`__, and this project tries
 its best to adhere to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`__.
 
+2025-12-10 - version 0.14.2
```
hakonanes:
```suggestion
2026-01-12 - version 0.14.2
```

hakonanes (changes requested):
Thanks for making the changes! Need to fix the date in the changelog, otherwise good!

hakonanes (approved):
Thanks for making these changes, @argerlt!

hakonanes:
We forgot to bump the version, as per our release guide. I'll make a new PR.

PR #631, opened 2026-02-23
Update __init__.py

argerlt:
Tagged release workflow only changes if __init__/.py changes. This triggers that change by fixing the author order


hakonanes (approved):
Good catch again. We need to be able to manually run the build creating a tagged release draft. My change enables this:
- https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow
- https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch

And BTW, the contributor list we use in source is this one: https://github.com/pyxem/orix/graphs/contributors?selectedMetric=additions.

argerlt:
There is some new workflow problem:
https://github.com/pyxem/orix/actions/runs/22326150715
Since the change is only on the main branch, GitHub doesn't let anyone run the workflow manually from the actions tab, so I had to use the github CLI, which might explain the weird failure:

```gh workflow run "Perhaps make a tagged release" --ref main```

One possible option is to make a PR on the default "develop" branch that adds the same line as above, allowing for manual dispatches.

hakonanes:
Hm. I've cherry-picked the workflow change in https://github.com/pyxem/orix/pull/632. Let's hope we can trigger it manually after that is merged.

PR #634, opened 2026-02-25
Post release 0.14.2 update

hakonanes:
This PR brings changes from main into develop.

I've also updated the release guide with steps to manually create a release draft when automatic creation fails.


hakonanes:
The builds fail because they cannot upload to coveralls.io, which currently returns an "Origin DNS error": https://coveralls.io/. I suggest to ignore this for now and see whether it gets fixed later.

argerlt (approved):

argerlt:
agreed.

Issue #531, opened 2024-11-08
Grain segmentation and KAM map tool


argerlt:
Currently working on my thesis and I don't have the time to properly add this feature to orix, but as part of an unrelated project, I needed to segment a grain map, so I wrote two different functions for doing so.

 If other members have input, I would be interested in hearing it, and if someone has interest in adding the feature themselves, they are more than welcome to adopt this in whatever way they see fit. Otherwise, I'll tackle it early 2025.

SHORT VERSION: 
I wrote a flood-fill method based off of skimage's pixel_graph, but didn't like that it took 10 seconds and didn't return bounding polygons. Then I wrote a second faster voronoi diagram-based function that returns the bounds, but it requires a package like `shapely` to calculate the polygons, and also cannot handle connectivity's other than a "plus" sign. 
I think the voronoi method is better, but the flood fill is far easier to implement. examples of both are shown below using publicly available data, as well as a Kernel Average Misorientation (KAM) map, which I get for free as part of the segmentation.

Comparison of segmentations
![image](https://github.com/user-attachments/assets/d7253030-ca7d-4b14-8319-69d273ccaebb)

KAM map
![image](https://github.com/user-attachments/assets/05682606-3a23-4e06-bbc5-ae1a65227fcb)

Link to the data which is available via Globus:
https://materialsdatafacility.org/detail/8f586539-f397-4af1-9c52-3c2e0da796c8-1.0
data shown here is located at `Processed data/EBSD data 4 - SHI, distortion corrected/GES_HEDM1_ebsdPCSHIFT_310_regi.ang`

```
# -*- coding: utf-8 -*-
"""
Created on Fri Nov  8 11:46:41 2024

@author: agerlt
"""

# scientific python stuff
import numpy as np
from scipy import stats, spatial, sparse
from skimage import measure
from skimage.util._map_array import map_array
from skimage.morphology._util import _raveled_offsets_and_distances
import sparse  # THis hsould probably be replaced with scipy.sparse...

# Orix stuff
from orix import io
from orix.plot import IPFColorKeyTSL
from orix.quaternion import Orientation, Rotation, OrientationRegion, symmetry
from orix.vector import Homochoric, AxAngle, Rodrigues, Vector3d
from orix.crystal_map import crystal_map

# plotting stuff
import matplotlib.pyplot as plt
from matplotlib.patches import Polygon
from matplotlib import cm
import copy

# pseudo- grain boundry stuff
import shapely
from shapely import MultiPolygon
[...]

Issue #573, opened 2025-07-11
New styling of imports


hakonanes:
I suggest a new "styling" of imports in the documentation (and test suite). Instead of importing functions and classes, we import modules.

Current:
```python
from orix.quaternion import Rotation

Rotation.random()
```

New:
```python
import orix.quaternion as oqu

oqu.Rotation.random()
```

If we're good at marking files as private (orix/quaternion/_misorientation.py etc.), users can look up of which functions and classes a module provides by tabbing with the cursor at `oqu.` etc.

Each module will then be importing accordingly:
```python
import orix.data as oda
import orix.crystal_map as ocm
import orix.io as oio
import orix.measure as ome
import orix.plot as opl
import orix.projections as opr
import orix.quaternion as oqu
import orix.sampling as osa
import orix.vector as ove
```

hakonanes:
Any thoughts, @argerlt, @CSSFrancis, or @viljarjf?

argerlt:
I've look at this a few times and am never sure how to feel. Every time I begin to write a response, I type out an example and change my mind. I also tried rewriting a few examples in this style, most felt better, some felt more convoluted.

Pros:
- Orix has some crazy long inmport sections. This would solve that.
- If you think of `from orix.quaternion import Rotation` as akin to `from  scipy.linalg import eig`, it is wild that we import bare classes and functions. Seem almost irresponsible from a reading comprehension point.
- I'd never have the frustrating experience of mixing up `scipy.spatial.transform.Rotation` with `orix.quaternion.Rotation` agian.

Cons:
- It's new and going to take time to get used to.
- I think I might prefer:
```
import orix as ox
import numpy as np
r =ox.Rotation.from_axes_angles(ox.Vector3d.zvector(), np.pi/4)
```


Though I suppose what your are suggesting wouldn't prevent this

CSSFrancis:
@hakonanes I personally dont ever really import modules. I think it becomes difficult to track all of the module short names and it's not hard to just make a long list at the top.

I also agree with @argerlt, but to maybe put a finer point to his suggestion.  My problem is the names of the modules.

Mostly the quaternion and crystal_map modules are confusing.  A module name shouldn't mirror a class name as there are other classes in that module. 

`from orix.crystal_map import Phase ` confused me for a for a  long while. And looking through the modules now I don't really have the first guess as to where the symmetries and point groups would be imported. 

Quaternion -> transform 
Crystal_map-> ???

hakonanes:
[...]

PR #642, opened 2026-03-23
Neighbor based analysing tool

Tagebh:

I am currently working on hybrid indexing as part of my master at NTNU, and I also wrote a project thesis on hybrid indexing in the autumn of last year, where I used the hybrid indexing tutorial made på Håkon as a starting point. As part of this work I have made a Kernel Average Misorientation (KAM) tool based on the code from @argerlt in issue #531. 

#### Progress of the PR

I have structured the code so it has the potential to be used for many things involving neighbors (not only KAM). 

The Neighbors() function performs the general preperations to do neighbor-based calculations on a xmap or any 2d array really, and can be used on it's own to do whatever you want with neighbors. 

The neighbor_misorientation() and KAM_calc() then uses Neighbors() to make a KAM map. 

I have also included a slight variation of KAM which I am calling Number of Same Neighbors, which also uses the results from Neighbors() and neighbor_misorientation() but does a slightly different calculation. A user defines what degree of misorientation that counts as a "different" orientation, and then it counts how many neighbors has the same orientation as a central point, where each point in the dataset acts as a central point once.

```python
import numpy as np

from orix.crystal_map import CrystalMap

from skimage.morphology._util import _raveled_offsets_and_distances
from skimage.util._map_array import map_array

def Neighbors(data,foot,mask=None):  
    """Performs the general preperations to do neighbor-based calculations on CrystalMaps in a vectorized form
        
    Args:
        data: 2d array. Can be a CrystalMap object
        mask: 2D array with same shape as CrystalMap indicating what points are valid or not eg. Non-indexed points
        foot: The kernel that defines which neighbors to look at. Can be both binary and not
        
    Returns: (maybe should be returned as an object instead?)
        indices: Array where the index for each valid point is repeated as many times as it has valid neighbors
        neighbor_indices: Array with the corosponding valid neighbors
        foot_values: contains how each neighbor in neighbor_indices is weighted defined by a given footprine/kernel
        num_neigbors_valid: the number of neighbors each valid point has
        nodes_valid: The index for valid points with at least one valid neighbor
        nodes_no_valid_neighbors: The index for valid points with no valid neighbors
        nodes_not_valid: The index for non-valid points 
        
    Raises:
        Have to add ValueErrors
    """
    if mask is None:
        if isinstance(data, CrystalMap):
            mask = data.is_indexed.reshape(data.shape)
        else:
            mask = np.full(data.shape, True) #All pointd are valid
    nodes = np.flatnonzero(mask)
    
    pad = np.max(foot.shape)//2
    padded_mask = np.pad(mask, pad, mode='constant', constant_values=False) 
    padded_nodes = np.flatnonzero(padded_mask)

    #The offset are given in 1D for the neighbors defined in 2D
    neighbor_offsets, dist = _raveled_offsets_and_distances(padded_mask.shape, footprint=foot)

    padded_neighbors = padded_nodes[:, np.newaxis] + neighbor_offsets
    neighbors = map_array(padded_neighbors, padded_nodes, nodes) #finds the indeces (in a non-padded format) for the neighbors belonging to valid points
    neighbors_mask = padded_mask.reshape(-1)[padded_neighbors] #sets which of those neigbors are valid
    
    num_neighbors = np.sum(neighbors_mask, axis=1)#the number of neighbors each valid point has
    indices = np.repeat(nodes, num_neighbors) #Array where the index for each valid point is repeated as many times as it has valid neighbors
    neighbor_indices = neighbors[neighbors_mask] #Array with the corosponding valid neighbors

    #Support for non-binary footprints 
    foot_offsets, dist = _raveled_offsets_and_distances(foot.shape, footprint=foot)
    foot_nodes = foot_offsets+foot.size//2
    foot_values = foot.reshape(-1)[np.repeat([foot_nodes],mask.sum(),axis=0)[neighbors_mask]]#contains how each neighbor in neighbor_indices is weighted defined by a given footprine

    valid_neighbor_mask = num_neighbors !=0
    nodes_valid = nodes[valid_neighbor_mask] #The index for valid points with at least one valid neighbor
    nodes_no_valid_neighbors = nodes[~valid_neighbor_mask] #The index for valid points with no valid neighbors
    nodes_invalid = np.flatnonzero(~mask) #The index for non-valid points (same as the mask input, not really neccesary)
    
    num_neighbors_valid = num_neighbors[valid_neighbor_mask]
    
    return(indices, neighbor_indices, foot_values, num_neighbors_valid, nodes_valid, nodes_no_valid_neighbors, nodes_invalid)


def neighbor_misorientation(xmap, indices, neighbor_indices, degree=False, crystall_symmetry=None):
    """Calculates the misorientation angles between a central point and its neighbors
    Args:
        xmap: CrystalMap object
        indices: Array where the index for each valid point is repeated as many times as it has valid neighbors
        neighbor_indices: Array with the corosponding valid neighbors
        degree (bool): True-return results in degree. False-return results in radiens
        crystall_symmetry: used to differentiate between rotation and orientation
    Returns:
        d: misorientation angles given in radiens
        
    Raises:
        Have to add ValueErrors
    """
    if crystall_symmetry is None:
        crystall_symmetry = xmap.phases[0].point_group #This can pehaps lead to problems, maybe make a check for multiple phases
    
    O_central_points = Orientation(xmap.rotations[indices], crystall_symmetry)
    O_neighbors = Orientation(xmap.rotations[neighbor_indices], crystall_symmetry)
    mis_ori = O_central_points.angle_with(O_neighbors)

    if degree:
        mis_ori = mis_ori*180/np.pi
    
    return mis_ori

def KAM_calc(xmap, mis_ori, non_index_value, no_neighbors_value, num_neighbors_valid, nodes_valid, nodes_no_valid_neighbors, nodes_invalid, foot_values=None):
    
    """Makes the KAM map
    Args:
        xmap: CrystalMap object (or any 2d array actually)
        mis_ori: Misorientation angles given in radiens or degrees (from neighbor_misorientation())
        non_index_value (int or float): Kam value given to a non-indexed points in the xmap
        no_neigbors_value (int or float): Kam value given to an indexed point without a single indexed neighbor

        These are from Neighbors()
        num_neigbors_valid: the number of neighbors each valid point has
        nodes_valid: The index for valid points with at least one valid neighbor
        nodes_no_valid_neighbors: The index for valid points with no valid neighbors
        nodes_not_valid: The index for non-valid points 
        foot_values: The weight of different neighbors if non-binary foot is used
        
    Returns:
        kam_map_im: 2D array with same shape as xmap with all the KAM values
        
    Raises:
        Have to add ValueErrors
    """
    if foot_values is None:
        cumulative_miso = np.add.reduceat(mis_ori, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]]) 
    else:
        cumulative_miso = np.add.reduceat(mis_ori*foot_values, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]])
    
    kam_map = np.full(xmap.size, np.nan, dtype=np.float32)
    kam_map[nodes_valid] = cumulative_miso/num_neighbors_valid
    kam_map[nodes_no_valid_neighbors] = no_neighbors_value
    kam_map[nodes_invalid] = non_index_value
    kam_map_im = kam_map.reshape(xmap.shape)
    
    return kam_map_im

def NSN_calc(xmap, mis_ori, non_index_value, no_neighbors_value, lim, num_neighbors_valid, nodes_valid, nodes_no_valid_neighbors, nodes_invalid, foot_values=None):
    """Makes a Number of same neighbors (NSN) map. Each point in the xmap gets assigned the value equal to the number similar 
       oriented neighbors. Where similar is defined by a user set limit 
    Args:
        xmap: CrystalMap object (or any 2d array actually)
        mis_ori: Misorientation angles given in radiens or degrees (from neighbor_misorientation())
        non_index_value (int or float): NSN value given to a non-indexed points in the xmap
        no_neigbors_value (int or float): NSN value given to an indexed point without a single indexed neighbor
        lim (int or float): The limit for when two neighboring point are defined to have different/same orientations

        These are from Neighbors()
        num_neigbors_valid: the number of neighbors each valid point has
        nodes_valid: The index for valid points with at least one valid neighbor
        nodes_no_valid_neighbors: The index for valid points with no valid neighbors
        nodes_not_valid: The index for non-valid points
        foot_values: The weight of different neighbors if non-binary foot is used
        
    Returns:
        NDN_map_im: 2D array with same shape as xmap with all the NDN values
        
    Raises:
        Have to add ValueErrors
    """

    same_neighbors = mis_ori < lim
    if foot_values is None:
        NSN = np.add.reduceat(same_neighbors, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]]) 
    else:
        NSN = np.add.reduceat(same_neighbors*foot_values, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]])
    
    NSN_map = np.full(xmap.size, np.nan, dtype=np.float32)
    NSN_map[nodes_valid] = NSN 
    NSN_map[nodes_no_valid_neighbors] = no_neighbors_value
    NSN_map[nodes_invalid] = non_index_value
        
    NSN_map_im = NSN_map.reshape(xmap.shape)
    
    return NSN_map_im
```

Here is the functions as .py file 
[_neighbors.py](https://github.com/user-attachments/files/26185394/_neighbors.py)

The reulting KAM and NDN (Number of different neigbors (NDN) = total neighbors - NSN) map using a 3x3 binary kernel (8 neighbors with equal weighting) could look something like this (using one of the nickel datasets already in kikuchipy):

image

I have also made a small jupyter notebook that I used to test the functions  
[test_neighbors.ipynb](https://github.com/user-attachments/files/26185566/test_neighbors.ipynb)

Here is an example from that notebook where the amount of equal colored neighbors is counted in a small dummy dataset
```python

import _neighbors as Neighbors #import my library

#make a dummy dataset
test_mask = np.array([[True,True,True,False,True,True,True,],
                      [True,False,True,True,True,True,True],
                      [True,True,True,False,False,False,True,],
                      [True,True,True,False,True,False,False],
                      [True,True,True,False,False,False,True,]])

test_data = np.array([[1,2,1,0,2,1,1],
                      [1,0,2,1,2,2,1],
                      [2,2,2,0,0,0,2],
                      [2,2,2,0,1,0,0],
                      [2,2,2,0,0,0,2]])

#define the footprint and run the preperation function neighbors()
x=3
foot = np.ones([x,x])
indices, neighbor_indices, foot_values, num_neighbors_valid, nodes_valid, nodes_no_valid_neighbors, nodes_not_valid=Neighbors.Neighbors(data = test_data,
                                                                                                                                        foot = foot,
                                                                                                                                        mask = test_mask)
#calculate how many neighbors how the same value/color as the central point
central_points = test_data.reshape(-1)[indices]
neighbors = test_data.reshape(-1)[neighbor_indices]

l = []
for i in range(len(ori)):
    if central_points[i] == neighbors[i]:
        l.append(1)
    else:
        l.append(0)
same_N = np.array(l)

np.r_[0, np.cumsum(num_neighbors_valid)[:-1]]
sum_same_N = np.add.reduceat(same_N*foot_values, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]])

im = np.full(test_data.size, np.nan)
im[nodes_valid] = sum_same_N
im[nodes_no_valid_neighbors] = -1
im[nodes_not_valid] = -2

im = im.reshape(test_data.shape)

#display the footprint used, the dataset and the number of same neighbors

print('Footprint:')
print()
print(foot)

fig, ax = plt.subplots()
rows = test_data.shape[0]
cols = test_data.shape[1]
ax.imshow(test_data, extent=[0,cols, rows, 0], origin='lower')
ax.set_xlim(0, cols)
ax.set_ylim(0, rows)
ax.set_xticks(np.arange(0, cols + 1))
ax.set_yticks(np.arange(0, rows + 1))
ax.set_xticklabels([])
ax.set_yticklabels([])
ax.grid(which='major', color='k', linestyle='-', linewidth=2)

for i in range(rows):
    for j in range(cols):
        ax.text(0.5 + j, 0.5 + i, im[rows - 1 - i, j], ha='center', va='center', color='red', fontsize=12)
        
ax.set_title('Number of same neighbors')

plt.show()

print('purple: data value is 0 (invalid)')
print('teal: data value is 1')
print('yellow: data value is 2')
print()
print('-2: invalid point (value set by user)')
print('-1: valid point without a valid neighbor (value set by user) (can be both green and yellow)')
```
The output with a 3x3 binary footprint
image

The output when the footprint is 3x3 and not binary (0.5 in the corners)
image

I do not have any experience with git hub, so I do not really know how to impliment it, but I hope this seems usefull and that someone wants to pick it up.

Kind regards
Tage

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.

### DIFF
--- a/orix/crystal_map/_neighbors.py
+++ b/orix/crystal_map/_neighbors.py
@@ -1,0 +1,164 @@
+import numpy as np
+
+from orix.crystal_map import CrystalMap
+from orix.quaternion import Orientation
+
+from skimage.morphology._util import _raveled_offsets_and_distances
+from skimage.util._map_array import map_array
+
+def Neighbors(data,foot,mask=None):  
+    """Performs the general preperations to do neighbor-based calculations on CrystalMaps in a vectorized form
+        
+    Args:
+        data: 2d array. Can be a CrystalMap object
+        mask: 2D array with same shape as CrystalMap indicating what points are valid or not eg. Non-indexed points
+        foot: The kernel that defines which neighbors to look at. Can be both binary and not
+        
+    Returns: (maybe should be returned as an object instead?)
+        indices: Array where the index for each valid point is repeated as many times as it has valid neighbors
+        neighbor_indices: Array with the corosponding valid neighbors
+        foot_values: contains how each neighbor in neighbor_indices is weighted defined by a given footprine/kernel
+        num_neigbors_valid: the number of neighbors each valid point has
+        nodes_valid: The index for valid points with at least one valid neighbor
+        nodes_no_valid_neighbors: The index for valid points with no valid neighbors
+        nodes_not_valid: The index for non-valid points 
+        
+    Raises:
+        Have to add ValueErrors
+    """
+    if mask is None:
+        if isinstance(data, CrystalMap):
+            mask = data.is_indexed.reshape(data.shape)
+        else:
+            mask = np.full(data.shape, True) #All pointd are valid
+    nodes = np.flatnonzero(mask)
+    
+    pad = np.max(foot.shape)//2
+    padded_mask = np.pad(mask, pad, mode='constant', constant_values=False) 
+    padded_nodes = np.flatnonzero(padded_mask)
+
+    #The offset are given in 1D for the neighbors defined in 2D
+    neighbor_offsets, dist = _raveled_offsets_and_distances(padded_mask.shape, footprint=foot)
+
+    padded_neighbors = padded_nodes[:, np.newaxis] + neighbor_offsets
+    neighbors = map_array(padded_neighbors, padded_nodes, nodes) #finds the indeces (in a non-padded format) for the neighbors belonging to valid points
+    neighbors_mask = padded_mask.reshape(-1)[padded_neighbors] #sets which of those neigbors are valid
+    
+    num_neighbors = np.sum(neighbors_mask, axis=1)#the number of neighbors each valid point has
+    indices = np.repeat(nodes, num_neighbors) #Array where the index for each valid point is repeated as many times as it has valid neighbors
+    neighbor_indices = neighbors[neighbors_mask] #Array with the corosponding valid neighbors
+
+    #Support for non-binary footprints 
+    foot_offsets, dist = _raveled_offsets_and_distances(foot.shape, footprint=foot)
+    foot_nodes = foot_offsets+foot.size//2
+    foot_values = foot.reshape(-1)[np.repeat([foot_nodes],mask.sum(),axis=0)[neighbors_mask]]#contains how each neighbor in neighbor_indices is weighted defined by a given footprine
+
+    valid_neighbor_mask = num_neighbors !=0
+    nodes_valid = nodes[valid_neighbor_mask] #The index for valid points with at least one valid neighbor
+    nodes_no_valid_neighbors = nodes[~valid_neighbor_mask] #The index for valid points with no valid neighbors
+    nodes_invalid = np.flatnonzero(~mask) #The index for non-valid points (same as the mask input, not really neccesary)
+    
+    num_neighbors_valid = num_neighbors[valid_neighbor_mask]
+    
+    return(indices, neighbor_indices, foot_values, num_neighbors_valid, nodes_valid, nodes_no_valid_neighbors, nodes_invalid)
+
+
+def neighbor_misorientation(xmap, indices, neighbor_indices, degree=False, crystall_symmetry=None):
+    """Calculates the misorientation angles between a central point and its neighbors
+    Args:
+        xmap: CrystalMap object
+        indices: Array where the index for each valid point is repeated as many times as it has valid neighbors
+        neighbor_indices: Array with the corosponding valid neighbors
+        degree (bool): True-return results in degree. False-return results in radiens
+        crystall_symmetry: used to differentiate between rotation and orientation
+    Returns:
+        d: misorientation angles given in radiens
+        
+    Raises:
+        Have to add ValueErrors
+    """
+    if crystall_symmetry is None:
+        crystall_symmetry = xmap.phases[0].point_group #This can pehaps lead to problems, maybe make a check for multiple phases
+    
+    O_central_points = Orientation(xmap.rotations[indices], crystall_symmetry)
+    O_neighbors = Orientation(xmap.rotations[neighbor_indices], crystall_symmetry)
+    mis_ori = O_central_points.angle_with(O_neighbors)
+
+    if degree:
+        mis_ori = mis_ori*180/np.pi
+    
+    return mis_ori
+
+def KAM_calc(xmap, mis_ori, non_index_value, no_neighbors_value, num_neighbors_valid, nodes_valid, nodes_no_valid_neighbors, nodes_invalid, foot_values=None):
+    
+    """Makes the KAM map
+    Args:
+        xmap: CrystalMap object (or any 2d array actually)
+        mis_ori: Misorientation angles given in radiens or degrees (from neighbor_misorientation())
+        non_index_value (int or float): Kam value given to a non-indexed points in the xmap
+        no_neigbors_value (int or float): Kam value given to an indexed point without a single indexed neighbor
+
+        These are from Neighbors()
+        num_neigbors_valid: the number of neighbors each valid point has
+        nodes_valid: The index for valid points with at least one valid neighbor
+        nodes_no_valid_neighbors: The index for valid points with no valid neighbors
+        nodes_not_valid: The index for non-valid points 
+        foot_values: The weight of different neighbors if non-binary foot is used
+        
+    Returns:
+        kam_map_im: 2D array with same shape as xmap with all the KAM values
+        
+    Raises:
+        Have to add ValueErrors
+    """
+    if foot_values is None:
+        cumulative_miso = np.add.reduceat(mis_ori, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]]) 
+    else:
+        cumulative_miso = np.add.reduceat(mis_ori*foot_values, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]])
+    
+    kam_map = np.full(xmap.size, np.nan, dtype=np.float32)
+    kam_map[nodes_valid] = cumulative_miso/num_neighbors_valid
+    kam_map[nodes_no_valid_neighbors] = no_neighbors_value
+    kam_map[nodes_invalid] = non_index_value
+    kam_map_im = kam_map.reshape(xmap.shape)
+    
+    return kam_map_im
+
+def NSN_calc(xmap, mis_ori, non_index_value, no_neighbors_value, lim, num_neighbors_valid, nodes_valid, nodes_no_valid_neighbors, nodes_invalid, foot_values=None):
+    """Makes a Number of same neighbors (NSN) map. Each point in the xmap gets assigned the value equal to the number similar 
+       oriented neighbors. Where similar is defined by a user set limit 
+    Args:
+        xmap: CrystalMap object (or any 2d array actually)
+        mis_ori: Misorientation angles given in radiens or degrees (from neighbor_misorientation())
+        non_index_value (int or float): NSN value given to a non-indexed points in the xmap
+        no_neigbors_value (int or float): NSN value given to an indexed point without a single indexed neighbor
+        lim (int or float): The limit for when two neighboring point are defined to have different/same orientations
+
+        These are from Neighbors()
+        num_neigbors_valid: the number of neighbors each valid point has
+        nodes_valid: The index for valid points with at least one valid neighbor
+        nodes_no_valid_neighbors: The index for valid points with no valid neighbors
+        nodes_not_valid: The index for non-valid points
+        foot_values: The weight of different neighbors if non-binary foot is used
+        
+    Returns:
+        NDN_map_im: 2D array with same shape as xmap with all the NDN values
+        
+    Raises:
+        Have to add ValueErrors
+    """
+
+    same_neighbors = mis_ori < lim
+    if foot_values is None:
+        NSN = np.add.reduceat(same_neighbors, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]]) 
+    else:
+        NSN = np.add.reduceat(same_neighbors*foot_values, np.r_[0, np.cumsum(num_neighbors_valid)[:-1]])
+    
+    NSN_map = np.full(xmap.size, np.nan, dtype=np.float32)
+    NSN_map[nodes_valid] = NSN 
+    NSN_map[nodes_no_valid_neighbors] = no_neighbors_value
+    NSN_map[nodes_invalid] = non_index_value
+        
+    NSN_map_im = NSN_map.reshape(xmap.shape)
+    
+    return NSN_map_im
--- a/orix/crystal_map/_neighbors.py
+++ b/orix/crystal_map/_neighbors.py
@@ -1,19 +1,19 @@
 import numpy as np
+from skimage.morphology._util import _raveled_offsets_and_distances
+from skimage.util._map_array import map_array
 
 from orix.crystal_map import CrystalMap
 from orix.quaternion import Orientation
 
-from skimage.morphology._util import _raveled_offsets_and_distances
-from skimage.util._map_array import map_array
 
-def Neighbors(data,foot,mask=None):  
+def Neighbors(data, foot, mask=None):
     """Performs the general preperations to do neighbor-based calculations on CrystalMaps in a vectorized form
-        
+
     Args:
         data: 2d array. Can be a CrystalMap object
         mask: 2D array with same shape as CrystalMap indicating what points are valid or not eg. Non-indexed points
         foot: The kernel that defines which neighbors to look at. Can be both binary and not
-        
+
     Returns: (maybe should be returned as an object instead?)
         indices: Array where the index for each valid point is repeated as many times as it has valid neighbors
         neighbor_indices: Array with the corosponding valid neighbors
@@ -21,8 +21,8 @@ def Neighbors(data,foot,mask=None):
         num_neigbors_valid: the number of neighbors each valid point has
         nodes_valid: The index for valid points with at least one valid neighbor
         nodes_no_valid_neighbors: The index for valid points with no valid neighbors
-        nodes_not_valid: The index for non-valid points 
-        
+        nodes_not_valid: The index for non-valid points
+
     Raises:
         Have to add ValueErrors
     """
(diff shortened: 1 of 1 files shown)


hakonanes:
Thank you @Tagebh, the functionality in your comment will be very useful to have in orix!

When adding your own commits, can you use the develop branch as the target branch here in the orix repo?

Tagebh:
Hey @hakonanes 

I changed the PR so it commits to the developer branch, and I added the code to orix/crystal_map/_neighbors.py. I also made the PR ready for review

hakonanes:
Great, thank you!

Just to confirm, is it fine with you if I push to your branch? I plan to push stuff like tests, documentation, and perhaps some restructuring of your file. Alternatively, I can make a PR to the develop branch in your fork, so you can go through my suggested changes before being visible here.

argerlt:
@hakonanes and @Tagebh , do either of you care if I try tackling this actually? Same comment as @hakonanes, making pushes directly to the original branch. 

@hakonanes, if you've already started, ignore this comment, there are plenty of other things for me to do in ORIX.

Also, general question: how do we feel about this pattern of keeping all the neighborhood functions in a `_neighbors.py` file? I think i like it, but just trying to think of potential pitfalls.

Tagebh:
@argerlt, as for me you are more than welcome to tackle this problem if you want to. I don't really know if @hakonanes has started or not, so it's probably best to wait for him to answer as well. 

I feel like that it's natural to add new neighborhood functions into `_neighbors.py`

argerlt:
@Tagebh, sorry this took so long, thank you for your patience.

Some starting notes:
1) To help improve readability and consistency, ORIX tries to stick as close as possible to the [numpydoc style guide](https://numpydoc.readthedocs.io/en/latest/format.html). In terms of this PR, this means switching docstrings from this:
```
    """Performs the general preperations to do neighbor-based calculations on CrystalMaps in a vectorized form
        
    Args:
        data: 2d array. Can be a CrystalMap object
        mask: 2D array with same shape as CrystalMap indicating what points are valid or not eg. Non-indexed points
        foot: The kernel that defines which neighbors to look at. Can be both binary and not
        
    Returns: (maybe should be returned as an object instead?)
        indices: Array where the index for each valid point is repeated as many times as it has valid neighbors
        neighbor_indices: Array with the corosponding valid neighbors
        foot_values: contains how each neighbor in neighbor_indices is weighted defined by a given footprine/kernel
        num_neigbors_valid: the number of neighbors each valid point has
        nodes_valid: The index for valid points with at least one valid neighbor
        nodes_no_valid_neighbors: The index for valid points with no valid neighbors
        nodes_not_valid: The index for non-valid points 
        
    Raises:
        Have to add ValueErrors
    """
```
to something like this:
```
def Neighbors(data: np.ndarray,
                          foot:np.ndarray,
                          mask:np.ndarray |None=None):  
    """Calculates the arrays necessary for neighbor-based calculations on CrystalMaps in a vectorized form
        
    Parameters
    ---------
        data
            either a 2D array or a crystal map.
        foot
            The kernel defining the local neighborhood.
        mask
            Boolean array of the same shape as `data`, indicating which points are included in the
            neigborhood lookup.
        
    Returns
    -------
        indices
            Array where the index for each valid point is repeated as many times as it has valid neighbors.
        neighbor_indices
            Array with the corosponding valid neighbors
    """
```

2) Again for sanity purposes, we try to stick with [the PEP 8 naming guidelines](https://peps.python.org/pep-0008/), specifically that funcitons should be lowercase, and functions intended for private use should have underscores before them. in this case, `Neigbors` should be `_neighbors`, `KAM_calc` should be `kam_calc`, etc.

3) I know my bad code is the root source of this error, but we try to avoid adding new dependencies when we can and especially dependencies on private functions like `skimage.morphology._util._raveled_offsets_and_distances` and `skimage.util._map_array.map_array` (note the underscores indicating private functions and modules)

4) ORIX as a whole also has a strictly enforced style rules to help ensure people a year from now can still read your code.  Check the [ORIX style guide](https://orix.readthedocs.io/en/stable/dev/code_style.html) as well as the guide for [setting up a development environment](https://orix.readthedocs.io/en/stable/dev/setting_up_development_installation.html) as a starting point, but ill also add you can cheese the system a bit by autofixing your PR by adding the following comment:

```
pre-commit.ci autofix
```

argerlt:
^ If you click on pre-commit.ci's commit above, you can see what rules it enforced.

orix/crystal_map/_neighbors.py
```diff
@@ -0,0 +1,226 @@
+import numpy as np
+from skimage.morphology._util import _raveled_offsets_and_distances
+from skimage.util._map_array import map_array
+
+from orix.crystal_map import CrystalMap
+from orix.quaternion import Orientation
```
argerlt, line 22:
```suggestion
import orix.quaternion as oqu
```

argerlt: Not an official change yet, but the current practice is to import modules as described [here](https://github.com/pyxem/orix/issues/573) to avoid confusion

orix/crystal_map/_neighbors.py
```diff
@@ -0,0 +1,226 @@
...
+
```
argerlt, line 23: I believe the following function will replace '_raveled_offsets_and_distances`
```suggestion

def _raveled_offsets(array_shape: tuple,
                     kernel: np.ndarray):
    """Compute neighboring pixel offsets in raveled coordinate space.
    
    This is a simlification of the "_raveled_offsets_and_distances"
    function in scikit-image.morphology._util, version 0.26.0. 
    
    Parameters
    ----------    
    array_shape
        the shape of the array the kernel is being applied to.
        equivalent to `array.shape`.

    kernel
        the 2D array representing the neighborhood. this is normally
        a 2D Von Neumann neighborhood such as [[0,1,0],[1,0,1],[0,1,0]],
        but could also be a larger array of weighted values. all
        non-zero entries will produce an offset value.

    Returns
    -------
    raveled_offsets
        an array of offsets for each non-zero entry in the kernel.
    """
    center = tuple(s // 2 for s in kernel.shape)
    offsets = np.stack(
        [(idx - c) for idx, c in zip(np.nonzero(kernel), center)], axis=-1
    )
    ravel_factors = array_shape[1:] + (1,)
    raveled_offsets = (offsets * ravel_factors).sum(axis=1)
    return np.sort(raveled_offsets)
```

orix/crystal_map/_neighbors.py
```diff
@@ -0,0 +1,226 @@
...
+
+def Neighbors(data, foot, mask=None):
```
argerlt: Few comments on this function:

1) Avoid capitalizing functions

2) In general, to improve usability and reduce maintenance down-the-road, we want to require as few inputs and return as few outputs as necessary in a function. I believe (though feel free to disagree, the minimum you need for this function is:
_find_neighbors(mask, kernel)--> [neighbors, indices]

Everything else you output can be quickly calculated if and when you need if from just the nxm raveled neighbors list, I would also go a step farther and make the indices an optional output since they can be calculated in one line from the mask.

orix/crystal_map/_neighbors.py
```diff
@@ -0,0 +1,226 @@
...
+
+    return (
+        indices,
+        neighbor_indices,
+        foot_values,
+        num_neighbors_valid,
+        nodes_valid,
+        nodes_no_valid_neighbors,
+        nodes_invalid,
+    )
```
argerlt: Here is my suggestion for an alternate function, which will require editing the remaining functions slightly, but I think will be far more useful down the road

```suggestion
def _find_neighbors(mask: np.ndarray | CrystalMap, 
                    kernel: int | np.ndarray = 8,
                    return_indices: bool=False,
                    ):
    """Given a 2D boolean array, return the neighbors for each pixel.

    Parameters
    ---------- 
    mask
    kernel
    return_indices

    Returns
    -------
    neighbors
    indices
    """

    # Convert mask to a 2D aray of booleans
    if isinstance(mask, CrystalMap):
        mask = mask.is_indexed.reshape(mask.shape)
    mask = np.atleast_2d(mask).astype(bool)        
    # Convert kernel to a 2D array of floats
    if isinstance(kernel, int):
        von_neuman_dict = {
            4:np.array([[0,1,0],[1,0,1],[0,1,0]]),
            8:np.array([[1,1,1],[1,0,1],[1,1,1]]),
            }
        try:
            kernel = von_neuman_dict[kernel]
        except KeyError:
            raise ValueError(
                "'kernel` must be either 4, 8, or a 2D numpy array")
        kernel = np.atleast_2d(kernel).astype(float)
        if not np.all([mask.ndim ==2,kernel.ndim==2]):
            raise ValueError("_find_neighbors only supports 2D arrays")

    # calculate the padding, offsets, and indices's of the queried points. 
    # note: padding must be at least one in all directions.
    pad = [(max(x//2,1),max(x-(x//2)-1,1)) for x in kernel.shape]
    offsets = _raveled_offsets(padded_mask.shape, kernel)
    nodes = np.flatnonzero(mask)

    # calculate neighbors for the padded array.
    padded_mask = np.pad(mask, pad, mode='constant', constant_values=False) 
    padded_nodes = np.flatnonzero(padded_mask)
    neighbors = padded_nodes[:, np.newaxis] + offsets

    # replace the padded indices with the correct unpadded ones, and set
    # out-of-bounds or invalid neighbor indices to -1
    # NOTE: this section is roughly equivalent to the following, faster
    # Cython code:
    #    from skimage.util._map_array import map_array
    #    neighbors = map_array(neighbors, padded_nodes, nodes+1) -1
    #
    # However, _map_array is a private function in skimage 0.26, so the
    # following vectorized method is used instead. If this slowdown becomes
    # problematic in the future, we should consider writing our own cython
    # code.
    is_neighbor = np.isin(neighbors,padded_nodes)
    depad_dict = dict(zip(padded_nodes, nodes))
    depad_func = np.vectorize(lambda x: depad_dict.get(x,-1))
    neighbors[is_neighbor] = depad_func(neighbors[is_neighbor])
    neighbors[~is_neighbor] = -1
    
    if return_indices:
        return neighbors, indices
    
    return neigbors
```

orix/crystal_map/_neighbors.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def neighbor_misorientation(
```
argerlt: I'm not sure this needs to be a function, but also the quick one-line version of this is:

``` (ori1*~ori2).reduce().angle```

(reads as "find the difference between ori1 and ori2, reduce  them with regard to their symmetries, and return just the angle")

it also works for big arrays, ie:

```
oris = xmap.orientations
misos = oris[nodes,np.newaxis]*~oris[neighbors]
```

File: orix/crystal_map/_neighbors.py [hunk 6]
```diff
@@ -0,0 +1,226 @@
...
+    O_neighbors = Orientation(xmap.rotations[neighbor_indices], crystall_symmetry)
+    mis_ori = O_central_points.angle_with(O_neighbors)
+
+    if degree:
+        mis_ori = mis_ori * 180 / np.pi
+
+    return mis_ori
+
+
+def KAM_calc(
```
argerlt: This function should take `xmap`, and `kernel` as required inputs and `mask` and `phases` as optional (you can skip `phase` and I can do it if you want, but for multiphase EBSD's, there should be options for all, some, or none of the phases to be included). The docstrings should also be reformatted and the name changed (I'd recommend `_kernel_averaged_misorientation_map` or '_calc_kam`)


.... also, not an issue now, but down the road, should we include a `grain_map` input? Most software calculates KAM maps intragrainularly.

orix/crystal_map/_neighbors.py
```diff
@@ -0,0 +1,226 @@
...
+    kam_map = np.full(xmap.size, np.nan, dtype=np.float32)
+    kam_map[nodes_valid] = cumulative_miso / num_neighbors_valid
+    kam_map[nodes_no_valid_neighbors] = no_neighbors_value
+    kam_map[nodes_invalid] = non_index_value
+    kam_map_im = kam_map.reshape(xmap.shape)
+
+    return kam_map_im
+
+
+def NSN_calc(
```
argerlt: Same advice as with `KAM_calc`.

argerlt (dismissed):

argerlt:
@Tagebh, I added some comments and suggested changes as a review, feel free to ask for clarification or suggest alternatives.

Also, once these changes are made, you/we will need to write unit tests and examples.

orix/crystal_map/_neighbors.py
hakonanes: I think we should keep absolute imports internally and only use these short form imports in the documentation.

hakonanes:
As discussed off GitHub, @argerlt, I think the public API from this PR should be the kernel average misorientation calculation only. How we handle the other functions internally is then of secondary nature, thankfully.

We can look to make more functions public afterwards, if needed.

hakonanes:
And it's my understanding from @Tagebh that you can make changes to his branch directly.

hakonanes:
@argerlt, if you don't have time, I can wrap this up. Up to you.

argerlt:
@hakonanes , sorry this took me a bit to get back around to. I swapped the functions around amd resolved my own review comments. 

I'll finish up some changes and documentation, then ping you when this is ready for review.

hakonanes:
Sounds good.

PR #653, opened 2026-05-25
Change left/right symmetry in reduction of (mis)orientations to FZ

hakonanes:

This PR fixes inconsistent results from reduction of (mis)orientations to the fundamental zone by correcting the order of the left and right symmetries. This is in line with our definition of a misorientation as

M = O2 * O1^-1 = s2 * M * s1^-1 = s2 * M * s1

I suggested this change back in 2023 (https://github.com/pyxem/orix/pull/442). However, I didn't pursue it further since @maclariz obtained inconsistent results for an hexagonal crystal after the change. As stated then, I believe the error is in our alignment of hexagonal symmetry operations with our crystal axes (a, b, c). As part of this fix, we should resolve this inconsistency, too. @maclariz, can you provide the dataset and analysis you referred to so I can troubleshoot?

Other changes:
- Make `Miller.get_nearest()` raise `NotImplementedError` instead of returning `NotImplemented`. This correctly signals that this is not supported and doesn't silently pass (bad!).
- Make `Miller.mean(use_symmetry=True)` raise `NotImplementedError` instead of returning `NotImplemented`. Again, explicit raise instead of silently pass (again, bad!).
- Update clustering notebooks (across fundamental zone boundaries, orientations, and misorientations) to be in line with new interpretation.
- Added Numba kernel for outer misorientation angles between orientations, `Orientation.angle_with_outer(other)`. This avoids large intermediate arrays, rendering the lazy path unnecessary. The lazy path is kept, but I'm fairly certain it's not needed anymore.
- Added Numba kernel for dot products between every misorientation, `Misorientation.get_distance_matrix(lazy=False)`. This avoids large intermediate arrays, rendering the lazy path unnecessary. The lazy path is kept, and the non-lazy is currently opt-in (`lazy=True` by default), but I'm fairly certain it's not needed anymore.

#### Progress of the PR
- [x] Update tests to align with fix
- [x] Update examples to align with fix
- [x] Changelog entry

```python
import numpy as np
import orix.quaternion as oqu
from orix.quaternion.symmetry import _groups

np.random.seed(42)

for group in _groups:
    Or1 = oqu.Orientation.random(symmetry=group)
    print(group.name, Or1.angle_with(Or1.reduce(), degrees=True))
```

Results before this fix:

```1 [0.]
-1 [0.]
211 [169.92343384]
121 [164.54086872]
112 [115.74856019]
m11 [0.]
1m1 [0.]
11m [0.]
2/m [169.35923973]
222 [43.17654809]
mm2 [0.]
mmm [0.]
4 [45.12764262]
-4 [175.36956292]
4/m [54.33477675]
422 [29.19569107]
4mm [90.]
-42m [100.32067289]
4/mmm [65.50051379]
3 [30.91841927]
-3 [69.99391495]
321 [56.06720968]
312 [86.95741444]
32 [60.29174314]
3m [117.99256843]
-3m [0.]
6 [24.31739648]
-6 [0.]
6/m [28.37775602]
622 [83.84686112]
6mm [120.]
-6m2 [55.99148019]
6/mmm [13.20013796]
23 [65.76928352]
m-3 [45.23047749]
432 [30.91580286]
-43m [41.15050328]
m-3m [18.29110171]
```

After this fix, all are 0, as expected.

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


hakonanes:
I discovered that I needed this fix back in '23 when projecting patterns from the detector to the sphere:

| 0.14.2  | This PR |
| ------------- | ------------- |
| test1  |  test2 |

maclariz:
No time to do this in detail just now @hakonanes !  I think there might be a minor mistake in my own original analysis and I'd need to check everything, not simply provide a final result which may itself have a problem in there.

hakonanes:
No problem! Thanks for the quick reply. But is there some data out there I can use for testing myself?

argerlt:
> I discovered that I needed this fix back in '23 when projecting patterns from the detector to the sphere:

As someone once again trying to once again incorporate HR-EBSD into my work, @hakonanes , you are very casually dropping an incredibly cool figure there.

argerlt:
EDIT: I originally had a way longer off-topic response, which I converted into a new issue instead. 

This fix looks right to me as well. The old way would give the expected coloring in #442 if users calculated miller indices as ` miller = vec*oqu.Rotation(ori)` or something similar which is different from the intended ORIX convention of  ` miller = ori*vec`. My assumption is both here and elsewhere, users have reversed the ordering to get the coloring they wanted, which is why they give nonsensical-looking data when the algorithm is corrected to respect the right-to-left convention elsewhere in ORIX.

hakonanes:
@argerlt, this is ready for review.

hakonanes:
Hm, the clustering across fundamental zone boundaries tutorial is a bit flaky, sometimes returning five instead of six initial clusters (reduced to two and three after accounting for the boundaries, respectively). I think I'll tighten the distributions slightly.

orix/quaternion/misorientation.py
```diff
@@ -568,6 +585,8 @@ def get_distance_matrix(
         chunk_size: int = 20,
         progressbar: bool = True,
         degrees: bool = False,
+        *,
```
argerlt, line 588: Did you really mean to put this wildcard here? I cannot tell what it is doing.

orix/tests/test_quaternion/test_misorientation.py
```diff
@@ -0,0 +1,196 @@
...
+            ]
+        )
+        assert np.allclose(distance1, expected, atol=1e-4)
+
+        distance2 = m1.get_distance_matrix(degrees=True)
+        assert np.allclose(np.rad2deg(distance1), distance2)
+
+        distance3 = m1.get_distance_matrix(lazy=False)
+        assert np.allclose(distance3, distance1, atol=1e-4)
+
```
argerlt, line 72: I think we should test the numba method here as well, just for completeness.

orix/tests/test_quaternion/test_misorientation.py
```diff
@@ -0,0 +1,196 @@
...
+        # computation of mismisorientation
+        # eq 6 in Johnstone et al. 2020
+        p1 = s1.outer(m).outer(s2)
+        p2 = s1.outer(~m).outer(s2)
+        # for identical symmetries this is equivalent to the old
+        # distance function:
+        # d = s2.outer(~m).outer(s1.outer(s1)).outer(m).outer(s2)
+        p12 = p1.outer(p2)
+        angle2 = p12.angle.min(axis=(0, 2, 3, 5))
+        assert np.allclose(angle1, angle2)
```
argerlt: I think we should also add your numba method here as well.

Thinking ahead to a worse-case future, if passive/active, right/left, row/column get mixed up again by a future dev, we would want a unit test that assured the dask and numba methods agree (and thus also reduce the chance that someone makes a misguided switch away from our intended conventions.)

argerlt (changes requested):
This looks good to me, other than asking for a few tests.

hakonanes:
Thanks, I'll look at your comments this afternoon.

orix/quaternion/misorientation.py
hakonanes: This means parameters after here must be given as keyword arguments. You can't do `get_distance_matrix(20, True, False, True)`, e.g., you have to do `get_distance_matrix(20, True, False, lazy=True)`.

I see the two biggest benefits as being:
- "Special parameters" must be set explicitly
- We can re-order the special parameters without breaking the public API

Since the `lazy` parameter is new, we can force it to be a keyword argument without breaking the API.

hakonanes: This [Real Python tutorial](https://realpython.com/python-asterisk-and-slash-special-parameters/#in-short-the-python-asterisk-and-slash-control-how-to-pass-values-to-functions) explains it well.

orix/tests/test_quaternion/test_misorientation.py
hakonanes: Do you mean the `.py_func()`? Because we _are_ testing the Numba function by setting `lazy=False`.

orix/tests/test_quaternion/test_misorientation.py
hakonanes: Good idea, I've added angles from the Numba path too.

orix/tests/test_quaternion/test_misorientation.py
argerlt: Ah, your right, I see the logic now.

argerlt (approved):

PR #663, opened 2026-06-08
Remove `outdated` package  causing dependency errors in github workflows

argerlt:

Replaces `outdated.check_outdated` with a simplified logic that doesn't rely on the depreciated `pkg_resources` module.

@hakonanes, I believe this will fix the relevant github action, but in the interest of time, I went ahead with publishing 0.14.3. This might be worth fixing before 0.15.0 though.

This update also silences several github action warnings regarding this [upcoming change to github actions](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)

### DIFF
--- a/.github/workflows/perhaps_make_a_tagged_release_draft.yml
+++ b/.github/workflows/perhaps_make_a_tagged_release_draft.yml
@@ -43,12 +43,10 @@ jobs:
       with:
         python-version: '3.13'
 
-    # TODO: Remove install of setuptools once
-    # https://github.com/pyxem/orix/issues/629 is resolved
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install packaging outdated 'setuptools<82'
+        python -m pip install json re requests packaging 
 
     - name: Check whether a tagged release draft should be made
       run: |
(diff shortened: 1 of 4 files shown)


argerlt:
pre-commit.ci autofix

.github/workflows/perhaps_make_a_tagged_release_draft.yml
```diff
@@ -34,21 +34,19 @@ jobs:
...
+    - uses: actions/checkout@v6
 
     - uses: r-lib/actions/setup-pandoc@v2
 
     - name: Set up Python ${{ runner.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v6
       with:
-        python-version: '3.x'
+        python-version: '3.13'
```
hakonanes: Why was this changed? Do we need 3.13 specifically? If not, I suggest to leave "3.x". For this workflow, the version isn't important, at least it hasn't before.

hakonanes (approved):
Brilliant, good solution. One question to resolve, otherwise fine to merge.

.github/workflows/perhaps_make_a_tagged_release_draft.yml
argerlt: It shouldn't have been changed, this is a relic of an earlier failed troubleshooting attempt.

PR #669, opened 2026-06-18
Calculate mean (mis)orientation using symmetry

### LINKED ISSUE #670: Misorientation.reduce() tosses proper/improper information
When working on #669, I noticed the following operation wipes out all proper/improper information:
```
import numpy as np
from scipy.stats import norm
import orix.quaternion as oqu
import orix.quaternion.symmetry as osm

m_as_is = oqu.Misorientation.random(1000,symmetry=(osm.C2v, osm.Oh))
m_as_is.improper = np.arange(0,1000)%2>0
m_reduced = m_as_is.reduce()

print(np.mean(m_as_is.improper))
print(np.mean(m_reduced.improper))
```

Do we really want to do that? We definitely want to include inversions and rotoinversions when defining the bounds of the fundamental zone, but I think wiping the inversion from the reduced result might be problematic.

If the answer is a definite "Yes this is should be fixed", I can fix it as part of #669, otherwise we can push #669 and discuss this more if needed.

### LINKED ISSUE #671: OrientationRegion does not calculate the proper fundamental zone for most improper misorientations
`OrientationRegion.from_symmetry` does not correctly calculate the fundamental zone for several of the possible misorientation symmetries. Looking at the 32*32 = 1024 possible combinations of point groups, I believe `OrientationRegion.from_symmetry` returns:

- The correct value for 704 cases.
-  A `NotImplemented Error` for the 100 cases where both start and end symmetries contain a roto-inversion but no inversion.
-  A fundamental zone twice as large as the correct value for all 220 cases where either the start or end symmetry contain a rotoinversion, but neither an inversion.

Looking deeper into this, i believe this is occurring because roto-inversions are being incorrectly ignored when they are the highest order symmetry.  The example code below shows this effect:

```
import numpy as np

import orix.quaternion as oqu

import orix.quaternion.symmetry as osm

# These are arranged so Laue groups are first .

groups = [
    # laue first
    osm.Ci,
    osm.C2h,
    osm.D2h,
    osm.C4h,
    osm.D4h,
    osm.S6,
    osm.D3d,
    osm.C6h,
    osm.D6h,
    osm.Th,
    osm.Oh,
   # Non-Laue with inversion centers
    osm.C1,
    osm.C2,
    osm.D2,
    osm.C4,
    osm.D4,
    osm.C3,
    osm.D3,
    osm.C6,
    osm.D6,
    osm.T,
    osm.O,
   # Non-Laue, no inversion but rotoinversion
   osm.Cs,
   osm.C2v,
   osm.S4,
   osm.C4v,
   osm.D2d,
   osm.C3v,
   osm.C3h,
   osm.C6v,
   osm.D3h,
   osm.Td,
    ]
# %%
n = len(groups)
representations_in_FZ = np.zeros([n,n],dtype=float)
r = oqu.Rotation.random(1)
for i, s1 in enumerate(groups):
    print(str(i)+ " of "+str(n))
    for j, s2 in enumerate(groups):
        try:
            fz = oqu.OrientationRegion.from_symmetry(s1,s2)
        except:
            continue
        q = s2.outer(r).outer(s1)
        mask = q<fz
        q_in_fz = oqu.Rotation(q[mask])
        q_in_fz.data[q_in_fz.a<0]=-q_in_fz.data[q_in_fz.a<0]
        total = np.unique(np.around(q_in_fz.data,5),axis=0).shape[0]
        representations_in_FZ [i,j] = total

# %%

fig,ax = plt.subplots()
ax.imshow(representations_in_FZ)
ax.grid('on',color='k')
ax.set_yticks(np.arange(n))
ax.set_xticks(np.arange(n))
ax.xaxis.set_ticklabels([x.name for x in groups],rotation=90)
ax.yaxis.set_ticklabels([x.name for x in groups])
```
Green means one and only one representation is inside the fundamental zone, yellow means two representations are, and purple means the fundamental zone could not be calculated.

Image

I believe this is occurring due to some [incorrectly implemented logic](https://github.com/pyxem/orix/blob/cc4f7938b6ea92c5568b3accbfd8cfb27bde7125/orix/quaternion/orientation_region.py#L86)  in  `OrientationRegion.from_symmetry`. Replacing this function with the following code seems to remedy the issue.

```
def get_proper_groups(Gl: Symmetry, Gr: Symmetry) -> tuple[Symmetry, Symmetry]:
    """Return the appropriate groups for the asymmetric domain
    calculation.

    Parameters
    ----------
    Gl
        First point group.
    Gr
        Second point group.

    Returns
    -------
    Gl
        First proper subgroup(s) or proper inversion subgroup(s), as
        appropriate.
    Gr
        Second proper subgroup(s) or proper inversion subgroup(s), as
        appropriate.

    """
    if Gl.is_proper:
        if Gr.is_proper:
            return Gl, Gr
        elif Gr.contains_inversion:
            return Gl, Gr.proper_subgroup
        else:
            return Gl, Gr.laue_proper_subgroup
    if Gr.is_proper:
        if Gl.contains_inversion:
            return Gl.proper_subgroup, Gr
        else:
            return Gl.laue_proper_subgroup, Gr
    if Gl.contains_inversion:
        if Gr.contains_inversion:
            return Gl.proper_subgroup, Gr.proper_subgroup
        else:
            return Gl.proper_subgroup, Gr.laue_proper_subgroup
```

I'm not sure it matters, but neither the old method nor my new version above exactly match the method described in [Krakow 2017 et al](https://royalsocietypublishing.org/rspa/article/473/2206/20170274/57500/On-three-dimensional-misorientation-spacesThree), which is more or less a voronoi decomposition in quaternion space. It does seem to give the same answer though for the 32 crystallographic point groups.

If no one sees an error in this implmentation, I can add it to #669, as it is relevant to the definition of a Misorientation's mean.

### LINKED ISSUE #673: `Orientation.dot_outer()`  scrambles axis ordering and can give the wrong answer
While writing tests for uncovered lines, I noticed the logic in  `Orientation.dot_outer` scrambles axis ordering for 2D or greater arrays or orientations:

```
import orix.quaternion as oqu
>>> g1 = oqu.Orientation.random(shape=(2,3,4))
>>> g2 =  oqu.Orientation.random(shape=(5,6))
>>> g12 = g1.dot_outer(g2)
>>> print(g12.shape)
(3,4,5,6,2)
```

Looking deeper, this logic for finding the smallest angle also  isn't guaranteed to work for roto-inversion symmetries:

```
        symmetry = _get_unique_symmetry_elements(self.symmetry, other.symmetry)
        M = other * ~self
        all_dot_products = Rotation(M).dot_outer(symmetry)
        highest_dot_products = np.max(all_dot_products, axis=-1)
        return highest_dot_products
```

I fixed both issues in #669, but I'm opening a second issue here just to discuss in case someone disagrees with my method.

argerlt:
~Currently a draft~, but this would be an up-to-date solution to #434 and #437
EDIT: complete, ready for review. Description below also updated for clarity.

As originally discussed, `mean` is calculated for transforms with symmetry using a variation of the 'flip-flop' method:

    1) reduce everything with respect to symmetry.
    2) find a rough mean (allow weighted averages if desired)
    3) convert (mis)orientations to the point closest to the rough mean
    4) recalculate the mean, again with weighted averages if desired.

In the process of this simple fix, it turned out ORIX had some inconsistent handling of fundamental zones and reduction.  This update makes it match exactly with the [original paper](https://royalsocietypublishing.org/rspa/article/473/2206/20170274/57500/On-three-dimensional-misorientation-spacesThree) previously mentioned as the souce for OrientationRegion.from_symmetry.

Other things fixed along the way:
1) added some unit tests for uncovered lines
2) fixed the error with `Orientation.dot_outer` discussed in #673
3) Added a lot of clarifying language
4) Detangled the previous logic in orientation_region.get_point_groups and added the logical extension discussed in the above paper.
5) fixed the loss of proper/improper discussed in #670

#### Progress of the PR

#### For reviewers

      section in `CHANGELOG.rst`.
      `.zenodo.json`.


hakonanes:
This is good, it's the same approach as taken in MTEX (I believe some call it the "flip-flop" algorithm).

argerlt:
pre-commit.ci autofix

argerlt:
Almost done, but the new failing test is a real error and I need to run some tests to see if #671 is the correct solution.

argerlt:
~~@hakonanes, if you get a chance, check out this updated docstring for `Misorientation.reduce()`.~~ 

Nevermind, heavily edited.

argerlt:
pre-commit.ci autofix

orix/quaternion/misorientation.py
```diff
@@ -672,6 +669,117 @@ def inv(self) -> Misorientation:
...
+        Setting `include_improper=True` will instead investigate all
+        symmetry options and treat all options as proper rotations
+        when calculating the mean.
+        """
+        if ignore_symmetry is True:
+            # convert to a rotation to emphasize loss of symmetry information
+            return Rotation(self.data).mean(weights=weights)
+
+        if verbose:
+            print("reducing to fundamental zone...")
```
hakonanes: I suggest to use logging for this, what do you think? We can then log these messages as debug logs.

orix/quaternion/orientation.py
```diff
@@ -638,11 +638,9 @@ def dot(self, other: Orientation) -> np.ndarray:
         >>> O1.dot(O2)
         array([0.92387953, 0.92387953])
         """
-        symmetry = _get_unique_symmetry_elements(self.symmetry, other.symmetry)
-        M = other * ~self
-        all_dot_products = Rotation(M).dot_outer(symmetry)
-        highest_dot_products = np.max(all_dot_products, axis=-1)
-        return highest_dot_products
+        sym = [self.symmetry, other.symmetry]
+        M = Misorientation(other * ~self, symmetry=sym).reduce()
```
hakonanes: Unless the old algorithm is incorrect, I suggest to revert this change, because the new algorithm is about 145x slower for the dot product between 10 000 random orientations (6/mmm and m-3m):

```python
import orix.quaternion as oqu
ori1 = oqu.Orientation.random(10_000, symmetry=oqu.symmetry.D6h)
ori2 = oqu.Orientation.random(10_000, symmetry=oqu.symmetry.Oh)

# Old
%timeit ori1.dot(ori2)
# 4.64 ms ± 205 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# New
% timeit ori1.dot(ori2)
# 676 ms ± 21.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

hakonanes: You claim in the commit where you made this change that the change lead to a speed-up? Can you give an example?

orix/quaternion/orientation.py
```diff
@@ -671,15 +669,9 @@ def dot_outer(self, other: Orientation) -> np.ndarray:
...
         """
-        symmetry = _get_unique_symmetry_elements(self.symmetry, other.symmetry)
-        M = other.outer(~self)
-        all_dot_products = Rotation(M).dot_outer(symmetry)
-        highest_dot_products = np.max(all_dot_products, axis=-1)
-        # need to return axes order so that self is first
-        order = tuple(range(self.ndim, self.ndim + other.ndim)) + tuple(
-            range(self.ndim)
-        )
-        return highest_dot_products.transpose(*order)
```
hakonanes, line 682: Again, I see a slow-down of about 88x with the new approach using reduction (similar setup to https://github.com/pyxem/orix/pull/669/changes#r3838690523). I suggest to just fix the ordering.

orix/quaternion/orientation_region.py
```diff
@@ -56,80 +84,109 @@ def _get_large_cell_normals(s1, s2):
     return normals
 
 
-def get_proper_groups(Gl: Symmetry, Gr: Symmetry) -> tuple[Symmetry, Symmetry]:
+@deprecated(since="0.15", removal="0.16")
```
hakonanes: Why is this function deprecated? I can't remember if we discussed doing so... Can you remind me whether we did?

I assume you just want to make it private and not actually remove it? If that's the case, I suggest to drop renaming the arguments now, as that requires deprecation of old parameter names, which would in this case just add unnecessary work, in my opinion.

orix/quaternion/orientation_region.py
```diff
@@ -56,80 +84,109 @@ def _get_large_cell_normals(s1, s2):
...
-        special consideration is needed which is not yet implemented in
-        orix.
+    start
+        Initial proper, inversion, or laue subgroup as appropriate.
+    end
+        Final proper, inversion, or laue subgroup as appropriate.
+
+    Notes
+    -----
+    ORIX follows the asymmetric domain/fundamental zone definitions
```
hakonanes: I suggest we stick to calling it just fundamental zones (and orientation regions) and not mix more terms like asymmetric domain. If you want to notify people that when they read about the latter somewhere, it really is a fundamental zone they're reading about, we can write "fundamental zone (sometimes called asymmetric domain)".

hakonanes (changes requested):

examples/misorientations/reducing_and_averaging_misorientations.py
hakonanes, line 1: Not sure why, but this example currently errors in the docs build (https://app.readthedocs.org/projects/orix/builds/34193732/).

orix/quaternion/orientation.py
argerlt: I agree, bad addition on my part.  It also shouldn't be part of this PR.

This is leftover from an unrelated dot product issue specific to misorientations with rotoinversions that I was trying to diagnose. It's only faster compared to the buggy in-between code I wrote, and is slower in all cases than the original.

orix/quaternion/orientation.py
argerlt: Agreed.

orix/quaternion/orientation_region.py
argerlt: EDITED version:
We never discussed it, but I think it should be ~~removed~~. renamed 

The function is used only once in ORIX (`OrientationRegion.from_symmetry`), and it also has a misleading name (it sometimes returns proper groups, sometimes Laue groups ). Importantly as well, it's internal logic differs from the Krakow paper and produces a "Not Implemented" error for the 200 rotoinversion misorientations.

However, it is the if/then/else logic necessary to get the asymmetric groups described in Morawiec's book. With that in mind, I think it should be renamed to "get_asymmetric_groups" or something similar.

Let me know if you like my change. If not, I'd ask that we still delete it, as it is an outdated method, but I can see cases where it might be handy to have around.

orix/quaternion/misorientation.py
argerlt: I'm indifferent (never use it) but it's easy to implement, so lets give it a go!

It's also very easy to remove later if it causes problems.

examples/misorientations/reducing_and_averaging_misorientations.py
argerlt: I also cannot tell. the documentation builds on my computer fine.

I'll try pushing versions on my fork until i find the problem code. Everything else should be ready for a re-review.

argerlt:
pre-commit.ci autofix

orix/quaternion/orientation_region.py
hakonanes: I agree that we should deprecate it in favor of `get_asymmetric_groups()`. But, I'm making two changes compared to how it is in this branch now:
- Reverting any change to it, except the decoration. We shouldn't change the API without a deprecation.
- We have to deprecate from 0.16.0 and remove in 0.17.0, not from 0.15 and remove in 0.16, as we've already released 0.15.0.

examples/misorientations/reducing_and_averaging_misorientations.py
```diff
@@ -0,0 +1,230 @@
...
+This example introduces the concept of reducing an orientation or
+misorientation with respect to symmetry, as well as the related
+concept of averaging a misorientation.
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy.stats import norm
+
+from orix.plot.rotation_plot import _setup_rotation_plot
```
hakonanes: We cannot document private functionality. I'll try to find a workaround.

hakonanes: Document as in use it in public docs.

orix/quaternion/misorientation.py
hakonanes: OK, I'll change the current implementation to the one used in kikuchipy, HyperSpy, and elsewhere (https://github.com/pyxem/kikuchipy/blob/develop/src/kikuchipy/logging.py): a public method to set the log level of orix, and then we log internally at a given level. It's important that the user gets to determine the log level at runtime.

hakonanes (approved):
I'm happy with this PR, as long as the doc build now passes (edit: it does). Great additions and fixes, including examples and expanded documentation.

@argerlt, can you look over my changes to make sure you're happy?

examples/misorientations/reducing_and_averaging_misorientations.py
hakonanes: Fixed. The problem was apparently using `eig` instead of `eigh`, because the former can incorrectly return small complex numbers (https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigh.html).

argerlt:
pre-commit.ci autofix

argerlt:
@hakonanes , Thank you for your edits, this is much better. I just made one final alteration:

My original purpose for the `print` statements was that the following code:
```
import orix.quaternion as oqu
x  =oqu.Misorientation.random(40,symmetry=[oqu.symmetry.C4,oqu.symmetry.D6])
out = x.mean(verbose=True)
```
Would give the output below, which was unhelpful unless you were already knowledgeable about how orix worked internally.
```
 98%|█████████▊| 47/48 [00:00<00:00, 2115.45it/s]
100%|██████████| 48/48 [00:00<00:00, 31107.32it/s]
```

I wanted to instead print this:
```
Reducing to fundamental zone
 98%|█████████▊| 47/48 [00:00<00:00, 2115.45it/s]
Checking for closer equivalent representations
100%|██████████| 48/48 [00:00<00:00, 31107.32it/s]
```

Implementing `logging` is an improvement regardless, but as-is, it didn't fix my original concern. 

The three solutions I tried:
1) Change the logger level inside the function based on the value of verbose, which might have ill-advised side effects
2) Add the log messages to `tqdm` via `tqdm(desc=log_txt)`, which in testing seems to cause tqdm to behave unpredictably.
3) Add the log message to `sys.stderr`, which is what `tqdm` uses for printing to the terminal as well.

I chose Option 3, it seems to have the best looking messages and is least likely to break things down the line, but since it's a functionality change, it's probably worth one last check by you before I merge.

hakonanes:
OK, but I suggest to either show the progressbars with the description printed by tqdm, or logging the info messages; not both.

Setup:

```python
import orix.quaternion as oqu
import orix.utils as ous

ous.set_log_level("DEBUG")

pg_m3m = oqu.symmetry.Oh
pg_6mmm = oqu.symmetry.D6h
mori = oqu.Misorientation.random(10, symmetry=(pg_m3m, pg_6mmm))
```

Then, verbose

```python
mori_mean = mori.mean(verbose=True)
```

outputs

```python
Reducing to fundamental zone:  46%|████▋     | 533/1152 [00:00<00:00, 1292.20it/s]
Checking for closer equivalent representations: 100%|██████████| 288/288 [00:00<00:00, 31117.74it/s]
```

While default with logging

```python
mori_mean = mori.mean()
```

outputs

```python
INFO:orix.quaternion.misorientation:Reducing to fundamental zone
INFO:orix.quaternion.misorientation:Checking for closer equivalent representations
```

I'll push this.

hakonanes:
I'm very happy with this now! Will merge once checks pass.

Issue #660, opened 2026-06-05
New patch release 0.14.3 and RC candidate 0.15.0 with left/right symmetry change


hakonanes:
I suggest to release the current crystal map shape fix on `develop` as a 0.14.3 patch release. Branch off of main, cherry-pick, release.

We can then make a release candidate 0.15.0 with the breaking left/right symmetry change from #653 from `develop` and wait for two weeks while addressing downstream broken functionality/assumptions. In this time, I don't think we should make any other releases.

This way, the crystal map shape fix thus doesn't have to wait those two weeks before being released. Other changes will have to wait to be released.

How does this sound, @argerlt?

hakonanes:
We can include https://github.com/pyxem/orix/pull/639 in the patch.

argerlt:
Sounds good to me.

argerlt:
@hakonanes , as a note, I am going to keep working on other PR's because they are on my mind, but even if I finish something like #647, I think we should still let it sit for a few weeks and stick to your plan here.

hakonanes:
Sounds good. I'll prepare the patch release tomorrow once #639 is in.

hakonanes:
0.14.3 patch release branch: https://github.com/pyxem/orix/pull/661

hakonanes:
0.14.3:
- [x] PyPI: https://pypi.org/project/orix/0.14.3/
- [x] conda-forge: https://anaconda.org/channels/conda-forge/packages/orix/overview
- [x] Zenodo: https://zenodo.org/records/20601180
- [x] RTD: https://orix.readthedocs.io/en/stable/index.html

hakonanes:
0.15.0rc1:
- [x] PyPI: https://pypi.org/project/orix/0.15.0rc1/
- [n/a] conda-forge: not releasing RC on conda-forge
- [x] Zenodo: https://zenodo.org/records/20632016
- [n/a] RTD: Docs aren't updated

argerlt:
Coming back to this after a bit, it looks like no one has reported any issues with the misorientation fix. 

Should we push 0.15.0 and mark this as done?

argerlt:
version 0.15rc has been accepted as the next official minor release 0.15.0. Related checklists below.

For Main:
* * [x] [documentation build](https://readthedocs.org/projects/orix/builds) is successful
* * [x] [Zenodo](https://doi.org/10.5281/zenodo.3459662) displays the new release.
[...]

PR #674, opened 2026-07-22
Fix sign error in quaternion-to-Euler conversion at Phi = pi

binggao1230:

`Rotation.to_euler()` returns wrong Bunge Euler angles for 180° rotations about an axis in the xy-plane that is not aligned with x or y (off-axis equatorial two-folds). `from_euler(to_euler(r))` then gives a *different* rotation (the axis is reflected).

The cause is a sign error in the `Φ = π` singular branch of `qu2eu_single` (`orix/quaternion/_conversions.py`), which handles `Φ = π ⇔ 180°` rotations with `w = z = 0`:

```python
else:  # Phi = pi
    a = -2 * qu[1] * qu[2]   # wrong sign
    b = qu[1] * qu[1] - qu[2] * qu[2]
```

Inverting `eu2qu` at `β = π` gives `qu[1] = -cos(δ)`, `qu[2] = -sin(δ)` with `δ = (α - γ)/2`, so `α = arctan2(2·qu[1]·qu[2], qu[1]² - qu[2]²)`, i.e. `a = +2·qu[1]·qu[2]`. Both quaternion components carry the minus sign here, so their product is positive — unlike the neighbouring `Φ = 0` branch, where only `qu[3]` is negated and `a = -2·qu[0]·qu[3]` is correct. This matches Rowenhorst et al. (2015) Eq. A.14. The bug is masked when `x` or `y` is zero (`2·qu[1]·qu[2] = 0`), which is the only equatorial case the existing tests covered.

I audited the other singular branches of the conversion suite (`ax2qu`/`qu2ax` at `ω = 0, π`, `om2qu` for 180° rotations, `ho2ax`, homochoric round-trips, and the `Φ = 0` branch of `qu2eu`) via round-trip consistency and found them clean — this is a single localized sign.

#### Minimal example of the bug fix

```python
>>> import numpy as np
>>> r = Rotation.from_axes_angles([1, 1, 0], np.pi)   # 180° about [1, 1, 0]
>>> r2 = Rotation.from_euler(r.to_euler())
>>> np.allclose(r.to_matrix(), r2.to_matrix())
True    # was False before this fix
```

#### Notes

- Added a round-trip regression test for equatorial two-fold rotations in `test_conversions.py` (jit, pure-Python, and public-API paths).
- `test_get_sample_reduced_fundamental` asserted `unique(phi1) == [0, π/2]` for C4/C6. That held only because the wrong sign collapsed the affected two-fold's `phi1` onto `π/2`; on `develop` those sampled rotations do not actually round-trip (matrix error 2.0). With the fix `phi1` gains its correct degenerate value `3π/2` and the whole sample round-trips, so I updated that expectation.
- Full `pytest orix/tests` suite passes locally.

#### Progress of the PR
- [x] Clean code style by running black via pre-commit
- [x] New features, API changes, and deprecations are mentioned in the unreleased section in `CHANGELOG.rst`

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -72,6 +72,9 @@
       "name": "Alexander Clausen",
       "orcid": "0000-0002-9555-7455",
       "affiliation": "Jülich Research Centre, Ernst Ruska Centre"
+    },
+    {
+      "name": "Vincent Gao"
     }
   ]
 }
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,17 @@ on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`__, and this project
 its best to adhere to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`__.
 
 
+Unreleased
+==========
+
+Fixed
+-----
+- ``Rotation.to_euler()`` now returns correct Bunge Euler angles for 180 degree
+  rotations about an axis in the xy-plane not aligned with x or y. A sign error in
+  the singular branch of the quaternion-to-Euler conversion previously reflected the
+  reconstructed axis, so the angles did not round-trip back to the input rotation.
+
+
 2026-06-10 - version 0.15.0
 ===========================
 
--- a/orix/__init__.py
+++ b/orix/__init__.py
@@ -39,4 +39,5 @@ __credits__ = [
     "Dorian Depriester",
     "Eric Prestat",
     "Alexander Clausen",
+    "Vincent Gao",
 ]
--- a/orix/quaternion/_conversions.py
+++ b/orix/quaternion/_conversions.py
@@ -1016,7 +1016,7 @@ def qu2eu_single(qu: np.ndarray) -> np.ndarray:
             a = -2 * qu[0] * qu[3]
             b = qu[0] * qu[0] - qu[3] * qu[3]
         else:
-            a = -2 * qu[1] * qu[2]
+            a = 2 * qu[1] * qu[2]
             b = qu[1] * qu[1] - qu[2] * qu[2]
             eu[1] = np.pi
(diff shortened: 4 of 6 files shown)


argerlt:
Hello, thanks for contributing to ORIX!.

Good catch, and thank you for referencing the equations as well, it made this much easier to verify.

For any other maintainers looking at this issue, another simple proof:
```
>>> x = Rotation.from_euler(Rotation.from_axes_angles([1, 1, 0], np.pi-1e-12).to_euler())
>>> y = Rotation.from_euler(Rotation.from_axes_angles([1, 1, 0], np.pi-1e-3).to_euler())
>>> print(x.angle_with(y))

array([3.14159265])
```
These should be less than 1/10th of a degree apart, not almost 180.

I'll add some minor edits this afternoon, then this will be ready for merging. 

Additionally, you should add your name to the [init file](https://github.com/pyxem/orix/blob/develop/orix/__init__.py) and your name + affiliation to the [zenodo.json file](https://github.com/pyxem/orix/blob/develop/.zenodo.json). I'll add my best guess as a suggestion later if you don't get to it first. You can also omit it if you choose, but if you contribute to ORIX, you should be credited on the DOI.

binggao1230:
Thanks — and nice extra proof, that one's cleaner than mine.

Added myself to `__init__.py` and `.zenodo.json` (`2f20347e`). I left the affiliation and ORCID off the Zenodo entry since I don't have one to give; a couple of the existing entries are name-only so it should validate fine.

No rush on the edits, ping me if you'd rather I fold them in.

argerlt (approved):
Approved. I was going to make some minor language changes originally (technically `Quaternion.to_euler` was fixed, not `Rotation.to_euler`), but I like your wording more, it's more relevant to the average user. 

Waiting for #660 to be finished, then we will merge this to `develop`

